Issue #172, opened 2014-10-22
Improve error handling and general CLI output


d0ugal:
This discussion started in #166. See that for some relevant comments.

Basically, the issue is that we handle user output of errors and generally information all over the place. A quick grep (well, ack) of the MkDocs source shows that we print in these files:

    ack --py print -l
    mkdocs/serve.py
    mkdocs/build.py
    mkdocs/gh_deploy.py
    mkdocs/main.py
    mkdocs/new.py
    setup.py

Some of these are legitimate and fine, but others are a bit more questionable. I don't think we should print anything from the build module for example, the current output would be better handled with logging.

Error handling is one of the other areas we should improve, and I'll just quote @tomchristie here:

    I'd suggest we'll probably want to have a standard exception
    type for config errors and other exiting statuses. We can
    then just wrap that at the very top level with the exit and
    print only happening in one place.

There are probably some other places I can't think of right now too...

d0ugal:
Closing in favor of #259

PR #177, opened 2014-10-26
Fix #164: serve ignores missing files

### LINKED ISSUE #164: `mkdocs serve` doesn't recover from build errors
For example, add a new page to the `pages` setting before you create the file, you then need to manually restart the serve command.

ngzhian:
We log an error about the missing file, and skip building of that page.
The logging is simplified here, and should be updated along with #172


d0ugal:
Merged in #317, Thanks!

PR #221, opened 2014-11-14
Implement installation of external theme via `them_url`

ngzhian:
First shot at enabling installation of theme via a url, specified by the config param `theme_url`.
`theme_url` should point to a zip file online. The zip file will be downloaded and unzipped to the `themes` folder.
Note that as of current implementation, `theme_url` installs the theme, but does not use it, i.e. the `theme` parameter needs to be specified in order for the newly installed theme to be used.
A point of consideration:
    - Should we check if the theme already exists? This is to prevent
      downloading every single time we run `config.py`


d0ugal:
Happy to consider this with an updated PR.

PR #223, opened 2014-11-14
Add Dockerfile for docs

tobegit3hub:
This Dockerfile works for all mkdocs projects. It would be great to add in docs so that everyone can reuse that.

Usage:
- `docker build -t docs .` 
- `docker run -d -p 8000:8000 docs`

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -1,0 +1,17 @@
+
+FROM ubuntu:12.04
+MAINTAINER tomchristie <https://twitter.com/_tomchristie>
+
+RUN apt-get update
+
+RUN apt-get install -y python-pip
+
+RUN pip install mkdocs
+
+ADD . /docs
+
+WORKDIR /docs
+
+EXPOSE 8000
+
+CMD ["mkdocs", "serve"]


waylan:
Hmm, I missed this issue before. I'm with @tomchristie on this. This is not something to include in the core project. Docker is hardly the only solution (it has competitors) and why should Docker be "the one"? Either MkDocs supports all such systems or none of them. If you really need a docker file, then create one and host it elsewhere. That's just my opinion, which doesn't really count for much here.

huglester:
Hello everyone.

Well installing mkdocs on some of systems could cause dependency issues. I think it would be nice to have this 'Dockerfile' ? I don't think it ads much of hassle? 

THank you

d0ugal:
It might not be that much hassle, but it also isn't that much hassle to use without us including. So long as it works well, you should use it.

If you want to try getting a few people behind it, I'd suggest making a mkdocs-docker (or similar) repo to collaborate on.

waylan:
The only thing that Docker file script does is install pip (and then use pip to install MkDocs), which is Python's standard package management tool. And that script will only run on Debian/Ubuntu based systems (it uses `apt-get`), which happens to exclude a tremendous amount of other systems.

It would be easier to provide instructions for how to install pip, than try to provide a cross-platform Docker file. And that would also work for non-Docker users.

Of course, "Python 2.7.9 and later (on the python2 series), and Python 3.4 and later include pip by default" ([source][1]).  Annoyingly, Debian/Ubuntu break up the standard package and do not install pip by default. Seems to me like that is a problem with Debian/Ubuntu, not a problem that MkDocs should solve.

Personally, while I use Debian/Ubuntu based systems, I actually prefer to not use the "apt" version of pip as it is always behind the latest release. IMO, it would actually be better to install pip directly from PyPI (in other words to no use the provided Docker script). Therefore, even if I used Docker (I don't), this script would be useless to me as it would muck up my system by installing the wrong version of pip. For that matter, I would most likely always be installing into a virtual environment, which this script won't even work with. The point is that this script hardly addresses the issue it claims to address.

In any event, the following two commands will get pip installed on any system (expect perhaps Windows--which doesn't matter as recent Python versions for Windows have various tools in addition to pip included by default to make this easy):

    wget https://bootstrap.pypa.io/get-pip.py
    python get-pip.py

Just be sure that second command is run with administrative privileges (perhaps use `sudo`).

[1]: https://pip.pypa.io/en/stable/installing/#pip-included-with-python

waylan:
> Annoyingly, Debian/Ubuntu break up the standard package and do not install pip by default. Seems to me like that is a problem with Debian/Ubuntu, not a problem that MkDocs should solve.

Sorry, perhaps that was a little unfair. However, if you (or the powers that be) require the use of apt for installation/dependency management, then it seems to me the solution would be to create a deb file for MkDocs and upload it to a PPA or something (i.e; let apt handle the dependencies). That would be a lot more useful to a lot more people than a Docker file. Then perhaps a PR adding a note to the docs with a link to the PPA  would get further along that this one.

PR #237, opened 2014-11-21
Initial suggestion for extension implmentation.

erichulser:


Argoday:
The idea behind this plugin design looks nice (publish events optionally consumed by listeners) - though to be useful a lot more events would need to be defined, and likely at a more granular level than macro-events like BuildPage.

I'm interested in seeing this updated with more polish - with documentation and an example or two to instruct users on how to actually use it

tomchristie:
My personal take is that anything like this is going to be an awkward distraction from instead focusing on a polished product without a plugin API.

The amount of support tickets etc that would be raised by implementing a feature like this would likely have a negative impact on the project as a whole.

This is @d0ugal's project and totally behind any decision he chooses to make, but my personal preference would be that we adopt an active stance *against* any kind of plugin API, with the view that it'd be harmful to the project as a whole.

Argoday:
lol - I guess I left that part out where I agree that this should be post-1.0

Was more just commenting on the construction so that when the project get's that far there is some feedback on how to build it

d0ugal:
Perhaps my fault for pointing @Argoday here from IRC :) Were were discussing how it could be a nice thing to have as @Argoday has a specific requirement that we agreed didn't belong in core, but yeah, we also agreed that there are more important gaps to fill.

The only driving force for **me** and having something like this, is that I have some ideas that I want to experiment with in MkDocs and doing that without a way to hook in is really hard work. I'd either need to fork or try and pre/post process files. Having said that, I've got plenty keeping me busy that I don't really have time for those wishful ideas :)

A good example is something like search, I'd have had a release out months ago if it were external and I didn't feel the same commitment as I do with putting it in MkDocs core. That could potentially have helped get it "finished" faster as one of the hurdles has been getting testers and feedback.

I do still generally agree however, I'm not in a rush to release and support this, but I would still love to use it!

tomchristie:
> lol - I guess I left that part out where I agree that this should be post-1.0

Noted :)

robodude666:
@tomchristie I would have to disagree. Something like this should be implemented and iterated over several releases prior to finalized 1.0. Pre-1.0 releases are expected to be volatile and changing, which will ensure that extensions are written in as part of the core and easy to use... instead of something hacked on post-1.0 that may require breaking changes in order to "correct".

waylan:
@robodude666 normally I would agree with your position. However, the only publicly facing API for mkdocs are the CLI and the config file (both of which have had some significant breaking changes in recent releases). While an extension API might add additional config settings, the recent redesign of the config has made that easy, should the need arise. Same goes for the CLI.

While adding hooks might break the current internal API, I don't see that affecting the public API.

d0ugal:
I'm trying to figure out what hooks we need and where, I'd like to see real use cases rather than us guessing on an API that might be useful. To that end I have started to experiment with that extensions might look like. So far I have [one work in progress](https://github.com/d0ugal/mkdocs-versioned) that I am building and using private APIs. Obviously I am doing that at my own risk and I'll happily break myself.

d0ugal:
I am also coming round to the idea of landing an extension API before a 1.0 just because if I start to invest much time I don't want to be doing it all with private APIs and I think we could benefit from a 1.0 opportunity to break things :)

d0ugal:
I'm going to close this as it is now woefully out of date. After the current restructuring we should pick it up and refer to this for ideas/inspiration.

PR #244, opened 2014-11-27
Don't use the default theme if the user only specifies a theme_dir

### LINKED ISSUE #241: Default theme files are added when a custom theme is used and no "parent" theme is specified
I'm using `theme_dir` to specify a custom theme directory with the latest versions of Bootstrap and Font Awesome, and I'm not specifying `theme` because I don't want to inherit files from a parent theme.  However, since `theme` defaults to 'mkdocs', my site directory ends up with two different versions of these libraries in addition to other files I'm not using like prettify.

I tried explicitly setting `theme` to null, but that didn't work.  As a hack, setting `theme` to a custom name works for `mkdocs build`, but not `mkdocs serve`.

This wasn't an issue with v0.9, but I've noticed it since upgrading to v0.11.1.

d0ugal:
fixes #241

### DIFF
--- a/mkdocs/config.py
+++ b/mkdocs/config.py
@@ -85,6 +85,9 @@ def load_config(filename='mkdocs.yml', options=None):
 
 def validate_config(user_config):
     config = DEFAULT_CONFIG.copy()
+
+    theme_in_config = 'theme' in user_config
+
     config.update(user_config)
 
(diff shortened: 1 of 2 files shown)


d0ugal:
Gonna land this, I ran into it when playing with a custom theme

PR #290, opened 2014-12-17
Added build datetime to context

roadhead:
Useful to know when the docs were last built. Datetime format can be controlled in the template like so:

``{{ build_date.strftime('%Y-%m-%d') }}``


d0ugal:
Okay, let's hold off on this for now. It might make sense to do this when we document and formalise the context that we pass to themes.

tomchristie:
Incidentally "current year" usage has slightly made me consider this one again, but even so still not a priority so I'd still suggest focusing on *core* stuff until we're at 1.0. Anything that doesn't fall into that category is a distraction.

PR #294, opened 2014-12-19
Fix TOC processing cutting off Markdown content

jdhoek:
The `footnotes` Python-Markdown extension appends a HTML div containing
the footnotes to the end of the generated HTML. This commit prevents the
`toc` class from accidentally moving that div to the TOC HTML output
instead of the main HTML output.


d0ugal:
Closing this for now, happy to consider an updated version if you get time.

PR #297, opened 2014-12-22
Initial stab at awesome-slugify integration

### LINKED ISSUE #212: Non ascii headers do not slugify nicely.
If we have a \<h1>web\</h1> in the markdown file, it will generate a hyperlink like "#web". This is what we expect but for other languages, it just returns "#_1" or "#_2".

d0ugal:
Closes #212 

The import magic is a bit messy at the moment, but works reasonably well. One gotcha here is that we don't test against awesome-slugify being installed yet.

### DIFF
--- a/mkdocs/build.py
+++ b/mkdocs/build.py
@@ -10,6 +10,21 @@ import json
 import markdown
 import os
 
+try:
+    from slugify import Slugify
+
+    def _slugify(value, separator):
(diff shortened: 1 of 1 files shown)


d0ugal:
Closing for now - the solution seems solid but let's come back to this in time for the 1.0 release as per the comments in #212

PR #299, opened 2014-12-22
Page titles no longer contain the site name

samhatfield:
If a page has a specific title, `page_title` will contain this and only
this. If the page has no title, or is the homepage, `page_title` will be
`None`. The `<title>` tag will always display at least the site name.

All of the themes' base.html files have been updated accordingly.


d0ugal:
Decided the risk is minimal - will document this in the release notes for 0.12. Thanks @samhatfield!

samhatfield:
My pleasure!

PR #317, opened 2015-01-05
Fix #164: serve ignores missing files

### LINKED ISSUE #164: `mkdocs serve` doesn't recover from build errors

d0ugal:
We log an error about the missing file, and skip building of that page

### DIFF
--- a/mkdocs/build.py
+++ b/mkdocs/build.py
@@ -9,6 +9,9 @@ import jinja2
 import json
 import markdown
 import os
+import logging
+
+log = logging.getLogger('mkdocs')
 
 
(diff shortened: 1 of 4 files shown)


tomchristie:
Suggest 'Closes #...' style in the description, rather than including in the title.

* It'll then be hyperlinked.
* It'll close the issue automatically once resolved.

I'd always prefer not to cross reference in titles - doesn't feel right.

d0ugal:
@tomchristie I agree, but this is just me rebasing a commit by @ngzhian in #177 decided not to edit it otherwise.

tomchristie:
Gotcha.

PR #320, opened 2015-01-06
Removed map file reference from RTD CSS

ePirat:
This removes the map file reference (`/*# sourceMappingURL=theme.css.map */`) from the RTD CSS, as it causes 404's because mkdocs doesn't provide this file.

Another solution would be to include the .map file.


d0ugal:
This change should happen upstream first, they appear to be missing the map file too. We are trying really hard to avoid editing this downstream.

https://github.com/rtfd/readthedocs.org/blob/master/readthedocs/templates/sphinx/sphinx_rtd_theme/static/css/theme.css

ePirat:
I think they will generate their theme with sass so this will produce the map file

Von meinem iPhone gesendet

d0ugal:
Maybe, I can't see it online however - do you get the same 404 when viewing a project on RTD? For example: http://pip.readthedocs.org/en/stable/

ePirat:
Yes you are right. Can you tell me where I find the original repo for the RTD theme, so I can file the bug there?

Issue #324, opened 2015-01-10
jquery loading (from CDN or local filestore) for local builds with readthedocs theme doesn't work


ci5er:
To fix this, there will either need to be local copies of both the CDN hosted javascript files in the theme directory (which there currently are not), or http: needs to be explicated as the protocol handler.

d0ugal:
Duplicate of #143.

PR #325, opened 2015-01-10
Update base.html

ci5er:
d0ugal marked #324 closed because of something to do with with a previous discussion about local store. That said: this templates src URLs are still "broken" (a bug) for non-HTTP-served content. 

It is just a reference error that only shows up when "http:" is not the default protocol to be handled -- i.e. when not being served. An alternative to this 'fix': store jquery and modernizr within the template directory structure itself and update this path reference to use {{base_url}}.

### DIFF
--- a/mkdocs/themes/readthedocs/base.html
+++ b/mkdocs/themes/readthedocs/base.html
@@ -18,8 +18,8 @@
   <link href="{{ path }}" rel="stylesheet">
   {%- endfor %}
 
-  <script src="//cdnjs.cloudflare.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
-  <script src="//cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js"></script>
+  <script src="http://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
+  <script src="http://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js"></script>
   <script type="text/javascript" src="{{ base_url }}/js/highlight.pack.js"></script>
   <script src="{{ base_url }}/js/theme.js"></script>
   {%- for path in extra_javascript %}


d0ugal:
I didn't mean to suggest the bug is invalid. I closed it as the issue was described in another issue which was also open - however, yours probably highlighted it a bit more clearly.

I think we should go for having a local copy rather than upadating the CDN link, since that wont always work offline.

ci5er:
That is what I have done locally, and given what you have done w/ the
mkdocs theme, I agree. I was baffled by the change to these lines in the
file done in November and couldn't tell which way *you* wanted to take it.

Thanks.

d0ugal:
Yeah, the issue is probably my fault. I missed that bit in #168 I think, or I didn't fully appreciate what it was changing.

Do you want to update this to bundle a copy with mkdocs? or shall I close this and do that.

ci5er:
I'm not sure that I know how ... but I've give it a shot. Best way to
learn, right? If you have a few hours -- I've got to take care of something
first.

If I understand you correctly, you want these resources made locally
available in the directories of both the mkdocs and the readthedocs (this
one) themes?

ci5er:
OK ... I think I have wrangled the delta/diffs (mostly) off of what I have
done to your x.11 release. Plus some things that were based on my
interpretation of what you were asking earlier (CST) today.

What do you want to do? These are (in total) some number of (mostly small)
unrelated changes. There are the local storage things, a couple of template
javascript (refactoring results?) things, and a couple of my "enhancements"
-- such as for sequence diagrams.

Is there a good way to get these to you discretely? That is -- in separate
bundles? I only ask because I would like to support you in the least
burdensome way to you. It's all the same to me to just throw it into a big
set of deltas -- but I'm assuming that this would be harder for you to
figure out.

Your guidance/help/mentoring is appreciated. And, if in fact, pushing any
of my deltas upstream is not needed, that is fine by me as well. Just let
me know.

If you decide to guide me a little, there is one type of question that will
come up. I am in love with the functionality of your mkdocs theme. It took
me a little while -- but that thing is super snazzy. I also have to do
silly/stupid stuff like (for example, as given above) ... attach notes to
sequence diagrams based on the parts of a complicated-ish system being
broken last night. The generated HTML from MD document is almost a roadmap
for the people in the morning.

That was ramble-ey. Sorry. The question is this: if someone finds a way to
integrate the (say) sequence-diagram-UML-markdown --> SVG into your system,
I believe that others would like this. What is the best way and the best
place to get that into your hands and or your process?

Best Regards,
W. Archibald
=

d0ugal:
Hmm, so, all I think we need to do is include the JavaScript files in the themes (much like they include other JavaScript files). So we shouldn't need to change anything other than the html (to make it a local link) and add the JS files.

PR #330, opened 2015-01-14
fixes #329 and #294

### LINKED ISSUE #329: footnote text missing
This is related to #294: When using the footnotes extension, the actual footnote texts are missing. 
Example markdown source:
```
This is normal text with a footnote[^2]
Another line of normal text

[^2]: This is the footnote text
```
This is rendered as follows:
```
This is normal text with a footnote²
Another line of normal text
```
The superscript number to indicate a footnote is properly rendered, but the actual text and the link target `This is the footnote text` is nowhere to be found.

jniggemann:


jniggemann:
I don't understand why this fails, but I'm sure I found the root cause of the issue. 
Perhaps this facilitates a proper fix from someone more knowledgeable than me.

jniggemann:
Ahh, the first failures are because of additional newlines (test_anchor_only_link, Test absolute image URL's are correct for different base_urls, Test relative image URL's are the same for different base_urls)...
The 4 tests regarding the TOC may well fail since I changed a bit in that area, leaving **Ensure that basic Markdown -> HTML and TOC works** and **Ensure that the fenced code extension is supported** to be looked into...

d0ugal:
Yup, looks like two new newlines at the start and the one at the end has been removed. Just updating the expected test output will be fine.

d0ugal:
Looking again, there are some more serious failures. For example, in this test we have lost the full table of contents, which is the same problem that @jdhoek had in #294.

    FAIL: test_flat_toc (mkdocs.test.TableOfContentsTests)
    File "/home/travis/build/tomchristie/mkdocs/mkdocs/test.py", line 212, in test_flat_toc
    self.assertEqual(str(toc).strip(), expected)
    AssertionError: '' != 'Heading 1 - #heading-1\nHeading 2 - #heading-2\nHeading 3 - #heading-3'

jniggemann:
In `test.py` (line 183), the order of variables needs to be reversed for `post_process` just like in `build.py`.

jniggemann:
Fixed that, two more to go:
Ensure that basic Markdown -> HTML and TOC works. fails because somehow a `<p>page_title: custom title</p>\n` is inserted, must be some meta-related issue.
And fenced codeblocks fail, too:
`<p><code>print 'foo'</code></p>" != "<pre><code>print 'foo'\n</code></pre>"`...

Working on it.

jniggemann:
Where is the source file `index.md` that's used in `test_copying_media`?

d0ugal:
The index.md file doesn't actually exist. It is essentially a mock or fake project.

jniggemann:
4 failures and 1 error remain...

### Failures
- test_anchor_only_link
- Test absolute image URL's are correct for different base_urls
- Test relative image URL's are the same for different base_urls
- test_dont_convert_code_block_urls

The first 3 are just about newlines and can be adapted later.

### Error
- test_copying_media

So that leaves us with one error and one failure, both of which I don't fully understand :-(

d0ugal:
I've taken a look, but I can't see the issue obviously. I guess I'd need to take a look at the footnotes extension and see what is going wrong.

jniggemann:
On a sidenote: It seems we don't need to care about inserting a [TOC] tag. From the toc extensions' documentation (https://pythonhosted.org/Markdown/extensions/toc.html):
>If a marker is not found in the document, then the Table of Contents is available as an attribute of the Markdown class.

jniggemann:
Footnotes are no longer an issue after applying #331

PR #331, opened 2015-01-16
removes TOC tag creation

jniggemann:
The [documentation for the toc extension](https://pythonhosted.org/Markdown/extensions/toc.html) states:
>If a marker is not found in the document, then the Table of Contents is available as an attribute of the Markdown class.

I refactored toc.py and build.py accordingly. Local tox testing showed newline differences in some tests due to my refactoring, I also adapted these tests.
A remaining issue is that test_copying_media errors out and I can't find the reason why. [Trace is here](https://gist.github.com/anonymous/db5dfa1b82fcc010c170).

### DIFF
--- a/mkdocs/build.py
+++ b/mkdocs/build.py
@@ -24,9 +24,6 @@ def convert_markdown(markdown_source, site_navigation=None, extensions=(), stric
     to the default set.
     """
 
-    # Prepend a table of contents marker for the TOC extension
-    markdown_source = toc.pre_process(markdown_source)
-
     # Generate the HTML from the markdown source
     builtin_extensions = ['meta', 'toc', 'tables', 'fenced_code']
     mkdocs_extensions = [RelativePathExtension(site_navigation, strict), ]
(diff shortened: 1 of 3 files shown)


jdhoek:
I tried this PR, and it appears to fix the issue I tried to fix in #294. Nice! :+1:

mkdocs/build.py
```diff
@@ -36,9 +33,7 @@ def convert_markdown(markdown_source, site_navigation=None, extensions=(), stric
     )
     html_content = md.convert(markdown_source)
     meta = md.Meta
-
-    # Strip out the generated table of contents
-    (html_content, toc_html) = toc.post_process(html_content)
+    toc_html = md.toc
```
d0ugal: huh, so they can give us the toc HTML? Neat! This is much better.

d0ugal:
Just reviewed this - looks great. I want to give it a quick manual test and then I'll merge. Really appreciate the research and debugging you done to push this forward. :)

d0ugal:
Seems great! Thanks again.

Issue #181, opened 2014-10-29
Document how to write custom themes.


d0ugal:
We need a section in the MkDocs docs for "Theme authors" - what files they need, what context variables etc. are availabe for them.

d0ugal:
I've started an experiment here: https://github.com/d0ugal/mkdocs-basic-theme

I think it is probably easier to demonstrate this stuff than document it in prose. Thoughts?

d0ugal:
(It isn't fully functional yet, but not far off either)

waylan:
Wow, that will be extremely helpful. I'd still like to see some prose in the docs, but the basic example is great for more advanced users.

d0ugal:
We aso need to cover how to package themes if #530 lands.

d0ugal:
With the addition of 43c64ab7669c6d4a492763b08c4b0918a0154248 and https://github.com/mkdocs/mkdocs-basic-theme, I think this is covered well enough for now.

PR #332, opened 2015-01-17
First stab at custom theme documentation

samhatfield:
I moved the custom theme discussion from the configuration page to the styling-your-docs page and expanded on it a little. Perhaps we should expand on the navbar and toc generation too. See #181.


d0ugal:
This is fantastic, thanks very much. I've been meaning to do this for a long time. A start like this will be much easier to build up. Just giving it a read through now.

d0ugal:
Thanks!

samhatfield:
No worries! It's not much, but it's a start.

PR #338, opened 2015-02-01
Add Google Analytics block to all themes

### LINKED ISSUE #333: Google Analytics works on the MkDocs theme only
The Google Analytics code is actually implemented on the MkDocs theme only (file base.html). The option google_analytics is ignored on any other theme.

In addition, looks like the second parameter of the google_analytics setting is no longer necessary. The latest code generated by Google Analytics uses 'auto' instead of the domain.

dustinfarris:
Copy+paste google_analytics block to all base.html files.

Fixes #333


d0ugal:
LGTM, thanks.

PR #343, opened 2015-02-06
Solving issues with local url and params.

robertogerson:
Previously, when pointing to a local url with params, for example:

  a.php?a=x&b=y

I was getting something like:

  a.php;a=x&b=y?a=x&b=y

Now I am able to get it correctly.


d0ugal:
Good spot, not sure how I managed to do that. Thanks! I just want to confirm this and maybe add a test but I think it makes sense.

d0ugal:
Manually tested, I don't think a test is really worth it in this case. Thanks!

PR #344, opened 2015-02-07
Have slash at end of all page links

rkday:
As discussed in https://github.com/rtfd/readthedocs.org/issues/1101, this makes directory links consistently have a trailing slash.

It looks like the fix for Python 2.6 in https://github.com/tomchristie/mkdocs/commit/f437ab058532851b369139e8ad261fbaee684448 could strip off `suffix` in addition to the additional `/` it was meant to remove - I've just switched the order so that `suffix` is appended after `rstrip('/')` is called.

I've also added a test (which failed without my fix).

### DIFF
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -98,10 +98,10 @@ class URLContext(object):
                 # Workaround for static assets
                 return '.'
             return url.lstrip('/')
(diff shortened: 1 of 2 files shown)


d0ugal:
Good catch!

PR #345, opened 2015-02-13
It's unnecessary condition for Bitbucket

abelloso:
Less code same result


tomchristie:
As it is allows us to better deal with unknown host strings.

PR #347, opened 2015-02-16
Fix hostname for BitBucket

### LINKED ISSUE #339: Hostname for bitbucket is incorrect
The host name used in the automatic `repo_name` detection is wrong. It should be using `bitbucket.org`, but instead it is `bitbucket.com`. This is found in config.py line 140.

pradyun:
Fixes #339


d0ugal:
Thanks!

pradyun:
You're welcome!
Also, I would like to help, more substantially, to this project. :innocent:

BTW, great job with this. It's a pretty good project. :)

PR #359, opened 2015-03-01
add documentation to sdist

alunduil:
When creating packages or installing from source it's good to have the
option of using the documentation provided with that code base.

I'm not sure if I've included the correct scope of items (really wasn't
sure about docs/CNAME) so if I need to edit this commit; please, let me
know.


tomchristie:
Appreciate the input, but the documentation shouldn't be part of the distributed package, so closing this one off. :)

Issue #143, opened 2014-09-19
Include jQuery locally in themes.


tsenart:
It seems that most themes are loading jQuery from `https://code.jquery.com/jquery-1.10.2.min.js`.
Sometimes this halts the loading of the whole page. Wouldn't it make sense to serve this resource ourselves instead?

PR #360, opened 2015-03-06
fixed loading jquery.min.js and modernizr.min.js with readthedocs theme

m1lhaus:
fixed loading jquery.min.js and modernizr.min.js with readthedocs theme when runing built html doc from file (i.e. file:///..../index.html)


d0ugal:
Changing to this causes mixed content warnings on an HTTPS page. Instead we should just include a copy of jQuery in the themes. As per #143.

Issue #351, opened 2015-02-21
Support Markdown Extension Configs


waylan:
Mkdocs uses Python-Markdown and offers support for users to specify extensions supported by Python-Markdown. However, currently mkdocs provides no way (sort-of; see below) to pass config settings into the extensions specified. For example, I would often like to use the [TOC][1] Extension's `permalink` feature to add the paragraph symbol (¶ or `&para;`) to the end of each header (`h1-6`) as a self-link. A very handy feature in documentation (in fact, it's used by Python-Markdown's docs, which I've linked to). Another use may be to customize how [syntax highlighting][2] works. Or non-english users may want to override the default substitutions used by [SmartyPants][3]. etc, etc...

Sure, mkdocs could hard-code some specific settings for things that make sense. Granted, not all extensions configs make sense in the context of mkdocs. But do you really want to be fielding bug reports for years to come asking for some additional config setting? And don't forget about [third party extensions][9]. Some of them (like a few of the math extensions) make a lot of sense for some users of mkdocs. But you couldn't possibly expect to provide good support for them without offering a holistic approach.

Interestingly, as mkdocs' config file is YAML and Python-Markdown's [`extension_configs`][4] keyword accepts a dictionary of dictionaries, it is pretty easy to replicate in the config file. Of note is the recent addition to Python-Markdown's CLI of support for a [YAML config file][5] for this very thing. One difference I would suggest is that, unlike the CLI feature, there should be no need to also provide a list of extensions separate from the extension configs.  I would expect the implementation to be something like this:

```python
if isinstance(config['markdown_extensions'], dict):
    extensions = config['markdown_extensions'].keys()
    extensions_configs = config['markdown_extensions']
else:
    # backward compat with old list type
    extensions = config['markdown_extensions']
    extensions_configs = {}

# then later in the code...
md = markdown.Markdown(
    extensions=extensions,
    extensions_configs = extensions_configs
)
html_content = md.convert(markdown_source)
```

Note that support for the existing list of extensions is still supported. The new behavior only happens when the config setting is of the dict type. And the new format in the config file would look something like this:

```yaml
markdown_extensions:
    markdown.extensions.toc:
        permalink: True
    markdown.extensions.codehilite:
        linenums: True
    markdown.extensions.smarty:
        smart_angled_quotes: True,
        substitutions:
            left-single-quote: '&sbquo;', 
            right-single-quote: '&lsquo;',
            left-double-quote: '&bdquo;',
[...]

PR #366, opened 2015-03-14
Add example for configuring markdown extensions

JoakimSoderberg:


waylan:
I would strongly recommend that this be rejected. While the suggestion does currently work without any changes to the code, Python-Markdown's support is [deprecated][1] in the latest release and it will not work at all in the next release.

While it is true that in the version of Python-Markdown supported by mkdocs, this feature has not yet been deprecated. However, based on my understanding of the mkdocs' road map, mkdocs will be using a version in which this feature is deprecated by the time mkdocs 1.0 is released. I doubt you want to encourage people to use a deprecated (and very limited) feature which will stop working by the next release.

Note that, even in older versions, the feature was very limited as no spaces or quotes were allowed and only string types could be used (booleans were tricky -- usually a 0 or 1 and then special code in the extension to call int() on the string).

It is for the above reasons that I have proposed what I consider to be a better way to support extension configs in #351. Of course, if users want to use the deprecated feature they can for now, But I wouldn't document it.

[1]: https://pythonhosted.org/Markdown/release-2.6.html#extension-configuration-as-part-of-extension-name-deprecated

d0ugal:
Agreed!

JoakimSoderberg:
Ah now I see why it wasn't documented :) As long as there is a solution!

d0ugal:
A long term solution anyway, at the moment what you found is all we have that really works :)

PR #388, opened 2015-04-01
Build on OSX

d0ugal:


d0ugal:
This isn't enabled for everyone, so disabling for now.

PR #395, opened 2015-04-02
Change the order of the Navigation processing to improve child_titles

### LINKED ISSUE #392: Title is used as a section if file is in subdirectory
Assuming I have a file at `research/stats.md` and a config line:

```
pages:
- ["research/stats.md", "Stats about Our Collection"]
```

I would assume that it would generate a top-level nav item titled  "Stats about Our Collection".

In reality, it generates a section **Stats about Our Collection** with a sub-item titled **stats**.

I'm 90% sure this has to do with the logic in [nav.py](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/nav.py#L212-L218) around `child_titles`.

d0ugal:
The current code was incorrectly processing titles for pages when an
explicit title was provided but the path to the file was nested in a
directory.

Fixes #392

### DIFF
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -209,13 +209,16 @@ def _generate_site_navigation(pages_config, url_context, use_directory_urls=True
             )
             raise exceptions.ConfigurationError(msg)
 
+        # If both the title and child_title are None, then we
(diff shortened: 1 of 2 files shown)


d0ugal:
Tests currently only failing on windows.

PR #396, opened 2015-04-02
Fix some path handling under windows

d0ugal:


d0ugal:
Well, that didn't go well. I went from 2 failing tests to 8.

PR #407, opened 2015-04-03
Fixing the scroll to anchor tags

AndyA13:
Consistent behaviour on page load and anchor click now.

Using the :before css to insert a margin above the item being linked.


d0ugal:
Thanks!

Issue #179, opened 2014-10-28
Implement reusable themes with `theme_url`


micahhausler:
I was thinking about how to reuse a custom MkDocs theme on multiple projects without copying files or using a git submodule, and it occurred to me that the [`jinja2.PackageLoader()`](http://jinja.pocoo.org/docs/dev/api/#jinja2.PackageLoader) could be used instead of the filesystem loader in [`build.py`](https://github.com/tomchristie/mkdocs/blob/b764af72abd1f30c707ac728ea192e6ccfe32d24/mkdocs/build.py#L167)

This could be as simple as:

https://pypi.python.org/pypi/my-mkdocs-theme
```
setup.py
my_mkdocs_theme/
    __init__.py
    version.py
    templates/
        css/
        fonts/
        img/
        js/
        base.html
        content.html
        nav.html
        search.html
        toc.html
```
And in each of my projects that use `my-mkdocs-theme`, 
`mkdocs.yml`
```yaml
template_pkg: my_mkdocs_theme
```
`requirements.txt`
```
my-mkdocs-theme==0.0.1
```

Thoughts?

d0ugal:
Closing in favor of #476

PR #410, opened 2015-04-03
Install theme via theme_url [implement #179]

ngzhian:
This is an updated PR from #221.


d0ugal:
Nice! By the way, don't worry about the appveyor fail - the tests are currently failing on Windows. See #396 where I'm trying to fix that. The travis tests will need fixed tho' and then I can do a review :)

ngzhian:
@d0ugal weird, the 2 test that failed seem to be some 404 error. That's a good chance to discuss with the way the test is written currently is suitable at all.

d0ugal:
@ngzhian The problem is that you put testing data (yeti.zip) into the themes directory. We should only have themes directories in there. I'd move testing data into the tests directory.

d0ugal:
To add further details - the integration test builds all themes, to find all the themes it looks in the theme directory. Until you added a zip file it only contained directories and no files - we could add a check to make sure it is a directory but I'd rather not have test data in there anyway.

ngzhian:
@d0ugal sweet, moved the zip file to `tests/` and tests passed.

d0ugal:
I played around with the code a bit and had a few suggestions. I sent you a pull request here: https://github.com/ngzhian/mkdocs/pull/1 - do you think they all make sense?

d0ugal:
I've rebased this to fix the conflict in #424

Issue #178, opened 2014-10-28
Live reload + Docker/Vagrant


robinandeer:
I'm trying to build my documentation in a Docker container with a shared folder between the VM and my local computer. However, mkdocs (or watchdog I guess) doesn't seem to register when I edit files locally. Therefore the live rebuild feature doesn't work.

Has anyone experienced something similar? I know that setting up e.g. Flask (using --reload) in the same way works just as expected.

My Dockerfile:

``` Dockerfile
FROM dockerfile/python
ADD . /code
WORKDIR /code
RUN pip install mkdocs

EXPOSE 8000
CMD ["mkdocs", "serve"]
```

My fig.yml file (if you've use it, basically just setting command line options for Docker) looks like this:

``` yaml
docs:
  build: docs
  volumes:
    - docs:/code
  ports:
    - "8000:8000"
```

PR #420, opened 2015-04-04
Automatically refresh the browser on edits

### LINKED ISSUE #163: Automatically refresh documentation when using `mkdocs serve`
This would make the eediting experience really neat. @ericholscher had an idea for a GUI app that could do this, but this approach might be enough.

One possible approach - inject some JavaScript that either polls or uses web sockets to be notified of an update.

d0ugal:
Use python-livereload to automatically refresh your browser when changes are saved.

Closes #163

### DIFF
--- a/mkdocs/config.py
+++ b/mkdocs/config.py
@@ -77,6 +77,8 @@ def load_config(filename='mkdocs.yml', options=None):
     options = options or {}
     if 'config' in options:
         filename = options['config']
+    else:
+        options['config'] = filename
     if not os.path.exists(filename):
         raise ConfigurationError("Config file '%s' does not exist." % filename)
     with open(filename, 'r') as fp:
--- a/mkdocs/serve.py
+++ b/mkdocs/serve.py
@@ -1,72 +1,9 @@
-# coding: utf-8
-from __future__ import print_function
-
-from watchdog import events
-from watchdog.observers.polling import PollingObserver
-from mkdocs.build import build
-from mkdocs.compat import httpserver, socketserver, urlunquote
-from mkdocs.config import load_config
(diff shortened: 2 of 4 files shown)


d0ugal:
Dropping watchdog and our hacky webserver for livereload seems like an epic win on it's own, but then we get automatic page refreshing thrown in fro free.

d0ugal:
The auto-reloading isn't as fast and slick as I would like, but generally this is working well. The serve output is totally different. Looking for a way to make it seem more familiar as we lost some information from before.

ngzhian:
wow this is great! the `serve.py` has almost no code right now :+1:

d0ugal:
@ngzhian Yeah, it's really neat :) Have you tried it? The output is quite different than before, but I think it is probably okay.

d0ugal:
@robinandeer if you still use a similar setup, would you be able to help by testing out the latest git master? You helped in the past with an issue regarding auto-reload and docker/vagrant (#178) . I'm concerned this change may have brought that issue back, but I'm not sure how to best test it.

robinandeer:
I've dropped my Docker setup. It quickly became a little too complex running on a Mac. However, I'd be happy to try out a Vagrant environment! Will get back soon.

d0ugal:
@robinandeer Thanks! I'll try and give it a test too - I was being lazy :)

d0ugal:
Unfortunately I had to revert this, I was able to test on Windows early this morning but it seems that python-livereload doesn't work well on Windows. I'll investigate later and then hopefully figure out a way to make this work.

d0ugal:
New attempt will be at #433

Issue #144, opened 2014-09-30
Global template variables?


heathermiller:
This is more of a question (and potentially a feature request if this isn't supported) than an issue – is there some way to declare global variables?

For example, in my config.yml, I'd like to define some global variable, `version-num` that, of course, represents the current version of the library that I'm documenting. I'd naturally like to use this all over in my documentation and have only one place to update the version number when releasing a new version – in config.yml.

Though I've tried to do this and beyond the pre-defined variables in the mkdocs documentation, nothing that I declare seems to be picked up.

Is there some way to do this that I don't see?

PS, Jekyll allows for this, so I'm basically assuming the same semantics as Jekyll, i.e., that user-defined variables in config.yml are globally accessible

waylan:
Reading through this PR, it occurs to me that the original feature request assumes that the user can define any custom settings (including settings that mkdocs knows nothing about) and that those should then be available in the template. However, the work in #496 makes that impossible as only "known" settings will validate. Creating an "unknown" setting would result in an error. In that case, a custom setting which contains a dict of user defined keys and values makes sense. Something like `custom`, `extra`, or `user_defined` seems logical (I like `extra`). Then, that setting can be passed to the template regardless of whether all of the config is.

I suppose the real question is: Is this issue about the custom user-defined settings or all settings being available in the template. If it is the former, then that should be an easy addition to #496 and the rest can be punted for later consideration. However, if it is the later, a final decision needs to be made first.

For the record,  I don't understand the aversion to having all of the config available to the templates. For Markdown files, I agree that not everything should be available (if #304 ever gets implemented), but whats the big deal for templates?

d0ugal:
@waylan 

> For the record,  I don't understand the aversion to having all of the config available to the templates. For Markdown files, I agree that not everything should be available (if #304 ever gets implemented), but whats the big deal for templates?

I'm not sure anyone is against it really, just trying to make the most sensible choice. I think it would make sense to pass the full config.

I'd like to look at this after I finish the work in #496. I think my plan will be to make the full config available as `{{ config.setting_name }}` to templates. Then use `extra` to have a dictionary of whatever the user wants. Then it could be accessed with `{{ config.extra.my_custom_thing }}`.

[...]

PR #424, opened 2015-04-06
Install remote themes with theme_url

d0ugal:
This is based on the work in #221 and #410 which I've just rebased to fix a conflict.

I think I'm happy with this, but I want to take some time to think about it as it is a big feature change to add. I'm going to look into how other projects are solving similar problems and see if there is anything we can learn before landing this.


d0ugal:
Moving this to the 0.13.0 milestone because I'm still not 100% sure on it and I want to try and get 0.12.0 out of the door in the next few days.

Argoday:
Looks fairly straightforward - why not just use the ETag for caching? (aka: make a HEAD request and if it comes back clean with consistent cache information leave the zip directory alone) ... it does feel a bit heavyweight to do a full download + unzip for every build

Though this can be a second pass upgrade as an improvement

d0ugal:
Good idea. It would be worth doing that for caching, that is probably the most technically correct way for fetching a HTTP resource.

We also need to make sure that we don't error if the user is offline (and use the most recent version).

waylan:
I have two concern with this. First, I dislike that the theme is being inserted into the existing package. The next time I upgrade mkdocs. any themes downloaded this way will be wiped out. In my experience that is just bad practice.

Second, I do not want the theme to be downloaded every time I build. Yes, I see some discussion about caching the download, but that discussion seems to assume that the download would still happen on any build in which a difference is detected between the cache and the remote file. What if updates to the remote file break my build? I want to be able to explicitly update the theme, or in the alternative, chose to continue to use an old version that I know works.

Seems to me that #431 is a much better option with each theme being a separate package on the PyPI (Cheeseshop). You get version specificity and download/installation for free.

d0ugal:
The theme isn't inserted into the package. It should be added to a .mkdocs directory next to the config (inspired by tox). If it isn't, that is a bug.

I agree, versioning is a concern. I had the same worry originally but at the time @tomchristie was against using python packaging for this. Honestly, I don't care for either approach (this or pypi) that much so I'd be welcome to new ideas.

If nothing else, the download needs to be smarter for when you use 'mkdocs serve'

#431 is regarding hosting built documentation on pythonhosted.org (it should be better worded).

tomchristie:
> I agree, versioning is a concern. I had the same worry originally but at the time @tomchristie was against using python packaging for this.

True, but I'm probably (?) lukewarm to *anything* other than plain ol' theme directories, so \<shrug\> :smile:

waylan:
Oh right. I misread the source code. No themes are being copied to the package dir. Sorry about that.

It seems to me this feature would be better suited to a CLI command as it should only ever need to be run once. Not sure why a setting needs to be used either. Once the theme is installed, it should be added to version control (assuming the rest of the docs are) and should never need to be fetched again. That is, of course, unless the user wants to update to a newer version of the theme. But the CLI command could handle that. Perhaps a theme could contain a data file (which starts with a dot so mkdocs ignores it) with the version and url that the CLI command could use. That way the user only ever needs to enter the full url on the command line the first time. From that point forward, it is all handled by the CLI command. Just a thought.

jimporter:
> Once the theme is installed, it should be added to version control (assuming the rest of the docs are) and should never need to be fetched again.

I'm not sure you'd want to add a theme's files to your project's repository. You don't necessarily want to add the source of mkdocs - or any other dependencies - directly to your repo.

Also, having themes represented by a URL is not necessarily a great idea, since it means that building the docs requires (for a `clean` build) the ability to connect to said URL. That's not always reliable for various reasons, and even when it works, the file you're fetching may have changed (as mentioned above). While documentation generally doesn't need `clean` builds as much as other parts of a project (e.g. automated tests), I think it could still be pretty frustrating.

I don't know what the solution to this is, but maybe there doesn't *need* to be one: as long as there's some (semi-) canonical place to put themes on your filesystem (maybe specified by an environment variable?), individual projects could probably come up with solutions that work for them. This could mean using a git submodule, or having a `make doc` command that fetches from a URL like this PR.

d0ugal:
So, I'm still feeling a bit unsure about this. The problem with the current solution is that it isn't any better than: `wget ...`, `unzip ...` and `mkdocs build`. People may as well use standard tools to do that.

I'm now wavering back towards using Python packaging for this. It gives us a solid solution for most of our problems. The only downside I see is the complexity for theme authors, somebody that is well versed in frontend development but not Python packaging would be able to create a theme otherwise. However, I expect if we have a good example for people to copy, it wouldn't be that bad.

If we go down that path, it would allow us to progress and we can perhaps worry about an easier to use approach later?

tomchristie:
I'll lay out my thoughts on this properly...

* Python packaging would be appropriate for an installable thing that has dependancies.
* Python packaging would be appropriate for code modules.

I'm not 100% convinced I see any great requirement for any kind of *automatically install theme remotely* behavior. I can see that it'd be nice to change a single line in the theme and have `mkdocs serve` instantly display a new theme. However as it stands we've one simple unambiguous way of installing a theme.

Would it make sense to defer work on this and start by:

1. Pulling the bootswatch themes out of the main package and having them as downloadable packages in the mkdocs org.
2. Finishing the [styling your documentation](http://www.mkdocs.org/user-guide/styling-your-docs/) section.

We could then consider versioned packaging at a later date?

d0ugal:
> 1. Pulling the bootswatch themes out of the main package and having them as downloadable packages in the mkdocs org.

@tomchristie What do you mean by "downloadable packages"? Just zip files that users download and extract into the project and then set `theme_dir`? or something else?

tomchristie:
> Just zip files that users download and extract into the project and then set theme_dir?

Yeah, that was my meaning.

waylan:
> So, I'm still feeling a bit unsure about this. The problem with the current solution is that it isn't any 
> better than: wget ..., unzip ... and mkdocs build. People may as well use standard tools to do that.

With one exception. That assumes that people have zip/upzip tools installed on their system. Most Windows users do not (they don't have wget either) and while those commands are available on Linux, they are considered non-standard (tar.gz or tar.bz is usually expected) and must be installed by the user. The one exception is that if they have Python installed (which mkdocs requires of course), then they have Python's zip lib. Therefore, I think it is extremely beneficial for mkdocs to have its own theme install command if you are going that root. `mkdocs theme URL` (where "URL" is the URL to a zipped theme) should fetch the URL, and unzip it to the appropriate location.

d0ugal:
> [..] unzip it to the appropriate location.

@waylan What were you thinking would be the appropriate location? Just `.mkdocs/theme` like in this PR or something else?

waylan:
> @waylan What were you thinking would be the appropriate location? Just `.mkdocs/theme` like in this PR or something else?

I'm unsure. I suppose `.mkdocs/theme` is a good place to start. Although I could also see an environment variable which defines a location (like `~/.mkdocs_themes/theme`). Of course, that means only one install for the entire system and doesn't allow different versions etc. An optional `--global` flag could address that I suppose. 

Whatever solution is chosen, it needs to work with tox, etc. For example a coding project  may want to automate the building of their project's docs as a test. They would need to tell tox to run the following commands (the "checkspelling.sh" command it simply for demonstration purposes as a reminder that the test may do more than simply "build"):

    pip install -r doc-requirements.txt
    mkdocs theme http://example.com/path/to/some_theme.zip
    mkdocs build
    checkspelling.sh path/to/site

That needs to run in isolation, even in multiple environments on the same system. I think that if you use a dir within the project then the `mkdocs theme` command needs to do the following (in this order): fetch the zip file, verify the zip file (does it look like a theme? Is it a valid zip file?), delete the existing theme dir (if any), and finally copy the newly downloaded theme to the theme dir. Of course, any failure should exit with an error code immediately. And it is important that the new download be verified before wiping out the old theme so as not to hose someone using a theme that no longer exists etc (maybe a 404 error page was downloaded). And yes, you need to delete the entire contents of the existing theme dir as a new version of the theme my not contain a file which existed in an old version and therefore wouldn't get overwritten.

Personally, I understand why other projects have adopted PyPI as a distribution system. If feels like you will just be reinventing the wheel here. But is does feel weird to have a python package which is entirely non-Python.

d0ugal:
> Personally, I understand why other projects have adopted PyPI as a distribution system. If feels like you will just be reinventing the wheel here. But is does feel weird to have a python package which is entirely non-Python.

This sums up my thoughts entirely. We wouldn't be the [first people to do this](https://pypi.python.org/pypi?%3Aaction=search&term=xstatic&submit=search), but it is still a bit strange and a horrible experience for average front end folks.

So, I'm somewhat torn about this as you can probably all tell. I'm going to try and state my motivations and see if it helps me more clearly see this.

- I actually don't care much for installing themes. I'm going to use the readthedocs theme for everything probably (other than MkDocs itself). I would like to try a version of the Sphinx [alabaster](https://pypi.python.org/pypi/alabaster) theme, *maybe*. Incidently, that is another example of non-python on PyPI.
- I really want to remove the default themes from MkDocs other than mkdocs and readthedocs to help with maintenance - these themes could do with itterating much faster than MkDocs itself. They also are mostly Bootswatch themes and are stuck on an old version of Bootswatch. Given I don't use any of these, they currently only get a best effort support.

With those motivations, I also would like to avoid just stripping out the current themes and telling users to go and fetch them. A bunch of people do use these themes, so that's a fairly nasty process. If we do that, I'd image a deprecation warning in `0.13` and then removing in `0.14` (or `1.0` if that comes sooner.)

So, I think, with all of the above. I have no interest in supporting something that requires us to implement our own packaging, however simple it is, I can see it getting complicated fast. So I think I'm going to abandon this PR, sorry @ngzhian for the change in direction. As for Python packaging? I'm not sure, it could work, I'm not entirely sure what it would look like yet.

I'm going to take the discussion to #476 (and updated related issues).

jimporter:
> I really want to remove the default themes from MkDocs other than mkdocs and readthedocs to help with maintenance - these themes could do with itterating much faster than MkDocs itself.

Would it make sense to allow parameterizing themes? The bootswatch themes appears to just be color swaps of the MkDocs theme, so if you could specify color swatches for the MkDocs theme, you'd probably be able to de-duplicate a lot of the current themes.

d0ugal:
@jimporter Yup, I think so. At the moment I do a search and replace to [update them all](https://github.com/mkdocs/mkdocs/commit/e19ae13b26a0a4b528a64cca5c4f35147786008b) which is somewhat tedious. Having said that, I don't want the infrastructure in MkDocs to do this build process, so I was waiting until I could move them out to a repo like mkdocs-bootswatch or similar.

jimporter:
@d0ugal Just to be clear, I'm imagining something like #144. Then you can have global template variables in your mkdocs.yml that can set the colors for a particular theme.

d0ugal:
@jimporter I see, thanks - I guess that could work if somebody wants to make a theme that can be customised like that. For something like bootswatch themes I think it would require too many settings to be useful. However, it's worth considering as I've not looked to see how much difference there is between the bootswatch themes.

Issue #352, opened 2015-02-21
Some symbols from the markdown aren't encoded correctly


AncientSwordRage:
I am trying to use the £ symbol in my markdown, and it's being displayed as 'Â£' which to me implies my utf-8 markdown is being served as non-utf-8 enconding.

Is this a known issue? I'm serving it via powershell on windows. Or is this a problem with how windows works?

d0ugal:
I suspect this is an issue related to Windows encoding. I know we have some users that use unicode characters (i.e. full chinese, I think, pages). Is your project somewhere I can access to confirm this?

IgorZh:
I have same problem with russian characters in github wiki.
[sample page](https://github.com/IgorZh/AngularFirstApp/wiki)
system: Windows 7, Python 3.4.3

waylan:
Those of you experiencing this problem, can you verify whether #428 fixes the issue for you?

IgorZh:
Can you give me some instruction, how to check, because I am no fimilar with python.
I download last source code. Run `python setup.py install` . Then in my project folder run `mkdocs build`. This not help.

d0ugal:
@IgorZh Did you checkout @waylan's `encoding` branch? You can also install it with :

    pip install -Ihttps://github.com/waylan/mkdocs/archive/encoding.zip

d0ugal:
This should be fixed by #428 - happy to reopen if somebody can confirm this is still an issue.

tkirill:
This is still an issue but this is a bug in livereload.  I reproduced it with Python 3.4.3 on Windows 8.1:

* mkdocs form master
* livereload 2.3.2
* run `mkdocs serve`
* Euro sign "€" displays as "в‚¬"

There is no issue if I run `mkdocs serve --no-livereload`.  Also, there is no issue if I use livereload from master.  Seems like switching to Tornado static file handler (lepture/python-livereload@1c67c02) fixed the issue.

d0ugal:
Yup, so as you found out, this will be fixed with the 2.4 release. https://github.com/lepture/python-livereload/issues/105

sugatoray:
I came here searching for how to include icons (fontawesome, octicon, etc.) along side the sections/titles in the nav-bar. It turns out that with `mkdocs-material` you can do this, but you need to use a small dirty trick. I documented the process [here](https://github.com/squidfunk/mkdocs-material/issues/2318#issuecomment-934173785).

[...]

Issue #239, opened 2014-11-25
Encoding problem in Windows


tobegit3hub:
When I run `mkdocs` in Windows, it fails and here's the log.

```
/D/Projects
$ mkdocs new my-p
Creating project directory: my-p
Writing config file: my-p\mkdocs.yml
Writing initial docs: my-p\docs\index.md
Traceback (most recent call last):
  File "c:\Python34\Scripts\mkdocs-script.py", line 9, in <module>
    load_entry_point('mkdocs==0.11.1', 'console_scripts', 'mkdocs')()
  File "C:\Python34\lib\site-packages\mkdocs\main.py", line 60, in run_main
    main(cmd, args=sys.argv[2:], options=dict(opts))
  File "C:\Python34\lib\site-packages\mkdocs\main.py", line 45, in main
    new(args, options)
  File "C:\Python34\lib\site-packages\mkdocs\new.py", line 51, in new
    open(index_path, 'w').write(index_text)
UnicodeEncodeError: 'gbk' codec can't encode character '\xa0' in position 348: i
llegal multibyte sequence
```

It seems to the problem about encoding. I think it would be better to support because python itself is cross-platform.

vbd:
Desktop Windows 8 + Python 2.7.9: similar error message
Notebook Windows 8.1 + Python 3.4.3: no error message, works as expected.

waylan:
This is almost certainly a problem with using the built-in `open` function without defining the encoding. In my experience, for cross-platform consistency, you must define the encoding when you open the file, not later. The `codecs` lib certainly addresses this, however, the `io` lib was added in Python 2.6 and it provides the Python 3 implementation of `open` in [Python 2][1]. In fact, the Python 3 [docs][3] state that `io.open` is "an alias for the built-in open() function."

So, I expect the fix for this is to import `open` from `io` and then do:

```python 
input_content = open(input_path, 'r', encoding='utf-8').read()
```

I'll test this on my windows box tonight.

[1]: https://docs.python.org/2/library/io.html#io.open
[2]: https://docs.python.org/3.4/library/io.html#io.open

waylan:
Hmm, I can't seem to reproduce this on my Windows 7 box. That said, I still think the above change should be made.

waylan:
Those of you experiencing this problem, can you verify whether #428 fixes the issue for you?

d0ugal:
This should be fixed by #428 - happy to reopen if somebody can confirm this is still an issue.

PR #428, opened 2015-04-07
Fix cross-platform encoding issues.

waylan:
By using io.open and explicitly defining the encoding on the call to
open every time, we should get consistant behavior across platforms
as open will not rely on each individual system's local settings
to determine what encoding to use. Note that io.open is an alias
to the built-in open function in Python 3. In Python 2.6+ it is a
Python 3 compatable implementation which works in Python 2. In the
future when all support for Python 2 is dropped, simply delete the
import statements with no additional changes to the code.

I expect this will fix both #239 and #352.

### DIFF
--- a/mkdocs/build.py
+++ b/mkdocs/build.py
@@ -6,13 +6,14 @@ from datetime import datetime
 from jinja2.exceptions import TemplateNotFound
 import mkdocs
 from mkdocs import nav, toc, utils
-from mkdocs.compat import urljoin, PY2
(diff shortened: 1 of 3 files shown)


d0ugal:
Neat! This makes lots of sense to me. I'll wait a little bit to see if anyone comments on it helping, or not, but otherwise I'm happy to merge it anyway.

d0ugal:
I'm testing in a Windows VM and this seems to help, or at least, not cause further issues.

PR #433, opened 2015-04-08
Automatically refresh the browser on edits

### LINKED ISSUE #163: Automatically refresh documentation when using `mkdocs serve`

d0ugal:
Use python-livereload to automatically refresh your browser.

Closes #163

### DIFF
--- a/mkdocs/config.py
+++ b/mkdocs/config.py
@@ -78,6 +78,8 @@ def load_config(filename='mkdocs.yml', options=None):
     options = options or {}
     if 'config' in options:
         filename = options['config']
+    else:
+        options['config'] = filename
     if not os.path.exists(filename):
         raise ConfigurationError("Config file '%s' does not exist." % filename)
     with open(filename, 'r', encoding='utf-8') as fp:
--- a/mkdocs/serve.py
+++ b/mkdocs/serve.py
@@ -1,72 +1,9 @@
-# coding: utf-8
-from __future__ import print_function
-
-from watchdog import events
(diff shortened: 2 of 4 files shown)


robinandeer:
I've tested this branch in Vagrant (Ubuntu Precise) and everything works as expected: edits made locally are synced to the virtual machine (where mkdocs is running) which triggers an automatic browser reload.

The only issue I noticed with the simplest setup is that you can only port forward the same port e.g. ``4000:4000`` and not ``4000:4001``. I guess that would require a proxy setting somewhere but I don't think it's a big issue.

Very nice improvement :+1:

d0ugal:
@robinandeer Thanks!

Unfortunately, when I tested this in Windows I had a strange issue that I need to take upstream and see if they can help before I can land this.

rickpeters:
Just today I tried using mkdocs on docker and found out that the refresh didn't work (so no live reloading). After searching all through the web I found this issue. **I can confirm that it is very essential that you use exactly the same port within the container for docker serve as you use on the host side. So when I use port 8000 within and outside the container it works great!** Maybe this can be mentioned in documentation somewhere? I am using boot2docker (1.5) on osx (yosemite) and a docker image based on official python image.

d0ugal:
@rickpeters Thanks - sounds like it would be a good addition to the documentation.

PR #448, opened 2015-04-09
bootstrap_base.css headerfix

peter1000:
Adds all headers plus  a classname `headerfix` to `bootstrap/css/base.css`

see @d0ugal gist: https://gist.github.com/d0ugal/f6c4790bf8c586004159

see discussion on MkDocs google groupe [Top Banner: Permalink question]
(https://groups.google.com/forum/#!topic/mkdocs/wTvEYbjRD88)

### DIFF
--- a/mkdocs/themes/bootstrap/css/base.css
+++ b/mkdocs/themes/bootstrap/css/base.css
@@ -3,7 +3,7 @@ body {
 }
 
 @media (min-width: 992px) {
-    h1[id]:before, h2[id]:before {
+    .headerfix, h1[id]:before, h2[id]:before, h3[id]:before, h4[id]:before, h5[id]:before, h6[id]:before {
         content: "";
         display: block;
         margin-top: -75px;


mkdocs/themes/bootstrap/css/base.css
```diff
@@ -3,7 +3,7 @@ body {
 }
 
 @media (min-width: 992px) {
-    h1[id]:before, h2[id]:before {
+    .headerfix, h1[id]:before, h2[id]:before, h3[id]:before, h4[id]:before, h5[id]:before, h6[id]:before {
```
d0ugal: I think we should support all the h1-h6 tags in MkDocs. The .headerfix is more specific and should be added by users with an extra css file.

If we were to include it, we would need to start documenting the CSS classes users can insert into markdown and I don't want to do that.

d0ugal:
Because we need to do this on all themes, I've updated my change and applied it to them all in #449.

Issue #431, opened 2015-04-08
Support cheeseshop docs hosting


d0ugal:

tomchristie:
Seconded. Was considering the same.
Alternative:

    `mkdocs deploy`

With the deployment list in the `mkdocs.yaml`, and with `github` as the default.
Further future (not worth a ticket yet) we could have this extensible via plugins.

waylan:
@wilhelmer the existing command is a light wrapper around a vendored (and slightly modified) copy of the [GitHub Pages Import][1] script. That script is specifically designed for GitHub Pages and outputs the messages which specifically name "GitHub." As we intend to only maintain our own modified copy until davisp/ghp-import#69 is accepted and released, making additional modifications to the vendored script doesn't really make sense.

That said, you make some valid points. My proposal for this issue is to create a new command `deploy`, which contains a list of subcommands, each of which is supplied by a plugin. That way, various third-party plugins could be created to handle all sorts of different deployment scenarios. A "deploy to git" plugin certainly seems like a great candidate, among others. 

[1]: https://github.com/davisp/ghp-import

wilhelmer:
Great. My only wish is that the "deploy to git" plugin is available right from the start so we won't lose the current functionality (which may only be an unintended side effect but is still super useful).

Issue #206, opened 2014-11-09
Plugin API.


erichulser:
Hey Tom,

I've worked on my own documentation generator for a while and used that but am in the process of upgrading it, so I wanted to see what was out there besides Sphinx (which is way to bloated).  I came across this project and I really like the simplicity of it.  One of the requirements I have for documentation is auto-module/class docs as a means to augment the prose style.

I know you've already stated that you have no interest in that, but I was wondering if you'd be interested in allowing me to work with you on this project.  I have all the code for generating docs that output a format like the ones used in Qt (http://qt-project.org/doc/qt-4.8/qdialog.html) from the python docs.  I would propose simply adding a plugin architecture to mkdocs (similar to django/pyramid) that would allow a developer to run additional addons so as to not pollute the core of the project.  Something like:

```
site_name: my_site
extensions:
       mkdocs_autodoc
pages:
       - ['index.md', 'Introduction']
       - ['user-guide/creating.md', 'User Guide', 'Creating a new Marshmellow project']
       - ['autodoc:my_project', 'API', 'My Project']
```

Where there would be a render hook that would dynamically generate the content from the `my_project` package as markdown files that could be then used and referenced from the prose style.  I would be happy to work with you on the structure and provide the work I've been doing for the auto-class documentation as a plugin.

Please let me know if this is something that you'd be interested in discussing, otherwise I may end up forking this project and implementing that as an option.

waylan:
@d0ugal stated above:

> Personally I'd be much more interested in autodoc functionality that enabled me to write prose. I'm not sure how this would work, but I'd expect markers or something in the Markdown that then pull in the relevant functions. So, rather that dumplicating the information in my docs, i write a paragraph and add that will then find and include the method dignature for me. I'm still thinking this idea through tho'

Just an FYI, I created a proof-of-concept Markdown Extension that does just that: [mddocs]. No mkdocs Plugin API required. Just add `mddocs` to the list of markdown_extensions. It is currently experimental and I wouldn't use it for anything serious until is stabilizes -- if it ever does (I'm not sure it is even a good idea yet).

[...]

PR #450, opened 2015-04-10
Refactor the current CLI implementation to use Click

### LINKED ISSUE #259: Improve Command Line Interface
I was reminded by #258 about this. When I first came to mkdocs, I tried this:

    mkdocs --help

That was fine, then I tried:

    mkdocs new --help

That doesn't do what I expected. It creates a documentation project in the directory `./--help/`

That is perhaps a bug in it's own right, but really, my point is that I think we should consider using a command line library (I have no strong preference) to improve the CLI. In terms of smarter arguement handling and better help messages. We would want to keep things compatible with what we have now.

I don't know *when* we would want to do this.

d0ugal:
This reimplements the CLI interface to use Click, which in turn gives us
a much easier to use interface as it provides help messages for all the
commands and individual commands.

Fixes #259


waylan:
Click is great. Glad to see you using it. But no descriptions of the commands? I don't even know what the `json` command does or why I would need it. The others are self-evident to me, but might not be to others. How about:

* Run a dev server
* Build documentation
* Create new project
* Deploy to Github Pages

I'm sure there is room for improvement, but click makes it easy so why not make use of its features and help your users.

d0ugal:
Sure. This is also work in progress - I'll nerd to start marking these more clearly. Not used to having an active reviewer :)

d0ugal:
In the current CLI we allow people to pass arbitrary arguments which are merged with the config. Click doesn't allow this out of the box, and I don't think we want it really. So we need to support a smaller subset that can be overwritten with the CLI. These are the ones I use most.

- `mkdocs build|serve --config-file`
- `mkdocs build|serve --theme=X`
- `mkdocs build|serve --dev_addr=IP:PORT`

It's hard to know what other people are doing at the moment. I can't think of anything else obvious that is missing.

d0ugal:
We also need to make sure we return with the right error codes etc. If nothing else, `--strict` should return a non-zero on any warnings.

waylan:
I'm assuming you will also be added the `--config-file`, `--strict` and `--theme` options to `build` and the `--config-file` and `--strict` options to `json`.

Just a suggestion, but it might be interesting to have a more generic `deploy` command which has an option to deploy to different services. A couple which might make sense out-of-the-box are GitHub (of course) and PyPI's [pythonhosted.org][3] (upload a zip file of the build dir). 

It would be really interesting if third parties could create their own deploy commands as plugins. While I certainly understand if that feels like more than you want to implement/support, if each deploy script is a separate subcommand of the `deploy` subcommand (eg: `mkdocs deploy github --option`), Click actually makes this very easy. Here's an [example][1] (with [docs][2]) from an unrelated project that I've implemented using entry-points. It was surprisingly easy (only two lines of code). Basically, the third party dev creates their own package which contains a Click command (which runs their deploy script) and registers it to the appropriate entry-point in their setup.py file (which can be distributed on PyPI, etc). Users of that third party package simply install it (using pip, etc.) and the new subcommand becomes available system/virtualenv wide and is even listed as a subcommand in the help menu created by Click.

[1]: https://github.com/waylan/msword-cli/blob/master/msword_cli.py#L411
[2]: https://github.com/waylan/msword-cli#plugins
[3]: http://pythonhosted.org

d0ugal:
@waylan Yup, that's a good idea. I was podering a generic deploy command too when I reported #431 

I do like the idea of allowing people to extend this, but I'd like to try and land this basic converstion first. It would be worth planning something like this with #206 if/when we do that.

d0ugal:
@waylan At the moment I've got a `--verbose` flag which almost works exactly as I want.

    mkdocs --verbose build

Do you know if there is a way in click to have it work so the order doesn't need to be like that? I'd like this to work too:

    mkdocs build --verbose

The only way I can see to do that is by adding the flag to all subcommands and delaying the logging config.

d0ugal:
Need to make sure I had help text for all the command line options - forgot about that.

waylan:
@d0ugal yes, you need to define the flag for each and every command you want it to be available on (documented [here][1]). Less than ideal I know. This is the one thing I dislike about the package. The relevant issue is [here][2] and includes some possible workarounds.

[1]: http://click.pocoo.org/4/commands/#passing-parameters
[2]: https://github.com/mitsuhiko/click/issues/108

d0ugal:
I'm going to leave it as it is for now. It's a little inconvenient, but not all that bad. I'll see if it gets really annoying and then it can be added to subcommands later.

fghaas:
Nice! Took the liberty to point out what I think are two trivial bugs (typos, really).

d0ugal:
@fghaas much appreciated, my typing can be a bit sloppy when I try to move too fast. Fixed in #487

Issue #306, opened 2014-12-31
Platform-dependent paths in extra CSS


orlade:
On Windows, if I create some file `/css/foo.css`, it is included in the final served HTML document as `<link href="../css\override.css" rel="stylesheet">`. Firefox returns a 404 on the file and doesn't load the styles. Chrome and IE11 rewrite the URL as `../css/foo.css` when they send the request, so the styles are loaded.

CSS (and JS) files included in the rendered page should only use forward slashes, regardless of the platform they are built on.

Lunatrius:
Using the master branch:
```html
  <link rel="stylesheet" href="../css/theme.css" type="text/css" />
  <link rel="stylesheet" href="../css/theme_extra.css" type="text/css" />
  <link rel="stylesheet" href="../css/highlight.css">
  <link href="../css\test.css" rel="stylesheet">
```
```log
"GET /css/theme.css HTTP/1.1" 200 -
"GET /css/theme_extra.css HTTP/1.1" 200 -
"GET /css/highlight.css HTTP/1.1" 200 -
code 404, message File not found
"GET /css%5Ctest.css HTTP/1.1" 404 -
```

d0ugal:
Thanks for confirming @Lunatrius. I'll try and have a look at this, I don't have a Windows machine I can easily test on however which makes things a bit trickier.

d0ugal:
We currently have tests failing on Windows - I suspect the issue is related. https://github.com/mkdocs/mkdocs/pull/396 (see the appveyor CI failures)

Lunatrius:
I usually use this *hack* when saving/writing relative paths (there is most likely a better alternative to this):
```py
__REPLACE__ = os.path.sep != '/'

def relpath(path, relroot=None):
    if relroot:
        path = os.path.relpath(path, relroot)

    if __REPLACE__:
        path = path.replace(os.path.sep, '/')

    return path
```

d0ugal:
Interesting, that could work well.

d0ugal:
@orlade, @Lunatrius Would either (or both!) of you be able to try out #451? It should fix your issues, I've only been able to do limited Windows testing. It's a fairly large refactor, so hoping to get as many people look at it as possible.

Lunatrius:
That does fix it.

d0ugal:
Thanks @Lunatrius

d0ugal:
Fixed in #451

PR #451, opened 2015-04-10
Improve Path handling across MkDocs

### LINKED ISSUE #439: Cross platform path handling problems.
There is some wonky path handling in MkDocs. We need to fix it to make it consistent and work reliably.

d0ugal:
This change aims to standardise path handling in MkDocs. Paths should
always be in the form expected by the current platform and then they
only need to be converted to URL's when used.

Fixes #439

### DIFF
--- a/mkdocs/utils.py
+++ b/mkdocs/utils.py
@@ -160,13 +160,15 @@ def is_html_file(path):
     ]
 
 
-def create_media_urls(nav, url_list):
+def create_media_urls(nav, path_list):
     """
-    Return a list of URLs that have been processed correctly for inclusion in a page.
+    Return a list of URLs that have been processed correctly for inclusion in
+    a page.
     """
     final_urls = []
-    for url in url_list:
+    for path in path_list:
         # Allow links to fully qualified URL's
+        url = path_to_url(path)
         parsed = urlparse(url)
         if parsed.netloc:
             final_urls.append(url)
--- a/mkdocs/utils.py
+++ b/mkdocs/utils.py
@@ -166,15 +166,18 @@ def create_media_urls(nav, path_list):
     a page.
     """
     final_urls = []
+
     for path in path_list:
         # Allow links to fully qualified URL's
-        url = path_to_url(path)
-        parsed = urlparse(url)
+        parsed = urlparse(path)
         if parsed.netloc:
-            final_urls.append(url)
-        else:
-            relative_url = '%s/%s' % (nav.url_context.make_relative('/'), url)
-            final_urls.append(relative_url)
+            final_urls.append(path)
+            continue
+        # We must be looking at a local path.
+        url = path_to_url(path)
+        relative_url = '%s/%s' % (nav.url_context.make_relative('/'), url)
+        final_urls.append(relative_url)
+
     return final_urls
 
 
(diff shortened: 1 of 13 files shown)


d0ugal:
This is very much a work in progress.

mkdocs/utils.py
```diff
@@ -216,12 +214,10 @@ def create_relative_media_url(nav, url):
...
+def path_to_url(path):
+    """Convert a system path to a URL."""
 
-    if ntpath.sep in path:
-        path = path.replace(ntpath.sep, posixpath.sep)
+    if os.path.sep == '/':
+        return path
 
-    return path
+    return pathname2url(path)
```
waylan: Are you sure pathname2url does what you want? In my experience it url encodes the path but does not flip backward slashes to forward slashes. It may make sense to call pathname2url on all paths converted to urls (system files may contain spaces or other characters generally not allowed in a URL), but only after the slashes are normalized.

waylan: I think perhaps this would be best:

```python
def path_to_url(path):
    """Convert a system path to a URL."""
    if os.sep != '/':
        path = path.replace(os.sep, '/')
    return pathname2url(path)
```

d0ugal: From a quick test of seemed to do what I want, but I'll verify. Otherwise, that looks good!

d0ugal: I done a very basic test in Windows and it worked and I googled a little and this seems to suggest pathname2url does do what I wanted. http://pymotw.com/2/urllib/#paths-vs-urls

d0ugal:
I've rebased on #433 and it fixed all the problems I was having there. The rebase also saved me looking into some weird issues with the old serve code after this change.

I've spent a bunch of time testing on Windows and things are looking pretty good. Still needs a bit more testing tho'

d0ugal:
I think I could be tempted to add this to the up and coming release, if it is right, it seems like it could really help with upcoming support. Anyway, will continue testing and see if I can get some other folks to try it out.

ci5er:
I've been testing these various cross-platform path-related changes on
Ubuntu Linux 14.x and the most up-to-date Cygwin. I don't know if that's
what you mean by "Windows testing", but if it is, it seems to be working
well.

There have been a few 'gotchas' with recent changes I have made to mkdocs
and python-markdown that I should have done a couple of PRs on long-ago,
but those all seem to be me causing my own problems. Again.

d0ugal:
General cross platform testing is what we need. To try and catch anything missed by the test suite and my manual testing. I'm least familiar with any and all windows environments, so that is lacking the most.

Thanks for the report!

Happy to look at the other issues you have had @ci5er - probably best in a new issue tho'

ci5er:
Happy to do what I can. To confirm: when you say "Windows testing", do you
mean Cygwin or MS-DOS, or Powershell or ... what? I can do the first
easily, and the others, not so much.

Cheers!

d0ugal:
I don't know? All of the above? I don't know how different they are. I'd hope MkDocs is consistent across all of them. So, maybe just test the ones you care about.

d0ugal:
Okay, positive feedback here and #306. I'll plan to merge this tomorrow unless I hear back from anyone else looking to do a review.

PR #457, opened 2015-04-15
Add search to MkDocs

d0ugal:
This work is built on the work done in #222, which in turn was built on #60.

After some frustration with Tipue search, I've changed the implementation to use Lunr.js. This is now functionally complete and all known issues are listed under the tasks.

## Tasks

- [x] Implement search index generation
- [x] Add Search to the MkDocs and ReadTheDocs themes
- [x] Style the search results to match those on ReadTheDocs (partly blocked by https://github.com/rtfd/readthedocs.org/issues/1222)
- [x] The code currently assumes that the theme will have jQuery included, this may be okay but we can probably find a better approach. jQuery requirement removed.
- [x] Prove implementation with DRF docs. https://github.com/tomchristie/django-rest-framework/pull/2849
- [x] Found a bug with [JavaScript being added to the index](https://dl.dropboxusercontent.com/u/14255217/Screenshots/Screenshot%202015-04-21-08.25.44.Google-chrome.png
).
- [x] On sites that use a modal to display the search, the index is being loaded on every page. We should only load it when the modal is opened. The problem here is that search is initialised when the search.js file is included by the theme. This isn't very flexible. We probably need to change this and add a requirement on the theme to call the initialisation - it's a pain as it is an extra step, but probably worth it.
- [x] Expand on the custom theme documentation and check for anywhere else that needs a mention for search.

## Testing

**Note:** only the `mkdocs` and `readthedocs` themes are supported for now.

Feedback is very welcome, if you want to test out this implementation, you can install it like this:

    pip install https://github.com/d0ugal/mkdocs/archive/search-new.zip

It is recommended you do this in a virtualenv and all usual disclaimers about using unreleased and unsupported code apply.


fedelibre:
For those interested in testing, this is the easiest way to install:

    sudo pip install git+https://github.com/d0ugal/mkdocs@search-new

@d0ugal The links in the search result don't have the base path. I guess that you are aware of it..

d0ugal:
@fedelibre Yup, this is very much a work in progress. I wouldn't recommend trying it yet. However, FWIW, installing with `pip install https://github.com/d0ugal/mkdocs/archive/master.zip` may be slight better for some people as it doesn't require you have git installed.

fedelibre:
Ok, I was curious.. Please let us know when it's ready for testing.
BTW, I just found a problem when using RTD theme:

    jinja2.exceptions.UndefinedError: 'current_page' is undefined

d0ugal:
@fedelibre The two issues you mentioned should be resolved now.

d0ugal:
I've added some styles to the ReadTheDocs results. They are fairly simplistic, but it now looks decent I think. Currently ReadTheDocs search is broken for Sphinx projects, so I can't compare with them to see what I'm missing.

fedelibre:
It works great! Both on local apache server and on file:/// protocol.
I like the istant search. I like also that I can easily browse using the keyboard only (tab key to switch from one resul to another and Alt + left arrow to go back).

I hope that it will be merged soon, once the few remaining issues are fixed.
Thanks @d0ugal !

d0ugal:
Rebased on master to fix a conflict. I'm not aware of any other issues. Finally I just want to try and get more feedback in general and specifically from the DRF folks here: https://github.com/tomchristie/django-rest-framework/pull/2849

d0ugal:
I'm going to leave this for a few days to give anyone interested a chance to give feedback and test.

If I don't hear anything I'll land it!

d0ugal:
Boom!

tomchristie:
Nice work :smile:

retrack:
Hello, great! 

May I suggest a refresh of Pypi ?

d0ugal:
@retrack I'll be working on that soon, there are a few other bits I want to finish before I make another release.

d0ugal:
... I also hope that now it is in master more people will come and shout at me and say what I managed to break ;)

olivernn:
@d0ugal glad to see lunr being put to use, if there were any issues integrating lunr or if you run into any problems, please let me know and I'll do my best to help out.

d0ugal:
@olivernn Thanks! It has been really easy, great job! If you feel like it, you could take a gander at the code and tell me what I've done wrong. I expect your JavaScript will be far superior to mine. 

https://github.com/mkdocs/mkdocs/blob/master/mkdocs/assets/search/mkdocs/js/search.js

PR #469, opened 2015-04-21
Create Wheels

d0ugal:


waylan:
I don't see where you specify that a "universal" wheel should be created. Without that, a wheel would only be created for the version of Python used to build the wheel. You would need to build a separate wheel for each supported version by running the command for each version. Given the same code runs unaltered on all supported versions, just tell it to build a single "universal" wheel. See docs here: https://packaging.python.org/en/latest/distributing.html#universal-wheels

d0ugal:
@waylan Good catch, I forgot to add the file in git.

PR #472, opened 2015-04-23
Fixes #316 - better navigation titles

### LINKED ISSUE #316: If no title is provided in the mkdocs.yml use the title in the markdown file
As per the title, this came up in #308 but it's something I've wanted for a while.

ugocupcic:


mkdocs/utils.py
```diff
@@ -226,3 +228,29 @@ def path_to_url(path):
         return path
 
     return pathname2url(path)
+
+
+def convert_markdown(markdown_source, extensions=[], extension_configs={}):
```
d0ugal: use extensions=None and extensions_configs=None.

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

mkdocs/nav.py
```diff
@@ -14,18 +14,35 @@
...
 
-    title = os.path.splitext(filename)[0]
-    title = title.replace('-', ' ').replace('_', ' ')
-    # Captialize if the filename was all lowercase, otherwise leave it as-is.
-    if title.lower() == title:
-        title = title.capitalize()
+    table_of_contents = None
+    try:
+        with open(filename, 'r') as f:
+            lines = f.readlines()
```
d0ugal: Use `f.read()` it means you don't need to join it again like you do below.

mkdocs/nav.py
```diff
@@ -14,18 +14,35 @@
...
+            html_content, table_of_contents, meta = utils.convert_markdown('\n'.join(lines), ['toc'])
```
d0ugal: If you add the following line here:

    return table_of_contents.items[0].title

Then you can simplify this a bit and remove the `table_of_contents = None` and the if statement below.

mkdocs/utils.py
```diff
@@ -226,3 +228,34 @@ def path_to_url(path):
...
+def convert_markdown(markdown_source, extensions=None, extension_configs=None):
+    """
+    Convert the Markdown source file to HTML content, and additionally
+    return the parsed table of contents, and a dictionary of any metadata
+    that was specified in the Markdown file.
+
+    `extensions` is an optional sequence of Python Markdown extensions to add
+    to the default set.
+    """
+    if extensions is None:
```
waylan: `extensions = extensions or []` is how this is typically how this is done.

mkdocs/nav.py
```diff
@@ -14,18 +14,33 @@
...
     """
     if utils.is_homepage(filename):
         return 'Home'
 
+    table_of_contents = None
+    try:
+        with open(filename, 'r') as f:
+            lines = f.read()
+            html_content, table_of_contents, meta = utils.convert_markdown(lines, ['toc'])
+            if len(table_of_contents.items) > 0:
```
waylan: Shouldn't the meta-data be checked first? Perhaps the document author defined the title there:

    title: Some Document Title

    # A header that is not the document title.

Then, if the title is not defined in the meta-data, fall back to the first header in the document.

d0ugal: That is never something I imagined doing. I've never seen that pattern in Markdown - why would you want a title that isn't the first header?

d0ugal: Actually, no, we shouldn't do that. If you really want to manually provide a title you can do it in the pages config. We don't need it here too.

waylan: Personally, that is the only way I would want to define the title. In fact I expected it to work this way until I discovered it didn't.

waylan: Why would I want to define meta-data bout a file separate from the file? I assumes that the pages list provided for defining the title because it was less expensive than reading the file twice. But if you are going to read the file twice (or at least once while building the nav), then I absolutely expect all meta-data to be defined in the file's meta-data. In my experience other static site generators offer this as an option. The meta-data is checked first, then a fallback to other methods is used. For some reason I don't understand, it seems that most users depend on the fallbacks.

ugocupcic: I've no real opinion on this but since we're already parsing this metadata, I can get that title from metadata. Since it's a supported way of setting the title in markdown it seems only fair that it'd work in this parser?

d0ugal: Sure, let's check the metadata first.

mkdocs/build.py
```diff
@@ -2,19 +2,17 @@
...
 import jinja2
 import json
-import markdown
-
-from mkdocs import nav, search, toc, utils
-from mkdocs.relative_path_ext import RelativePathExtension
-import mkdocs
+import os
+import logging
+from io import open
```
d0ugal: Why have you moved these around?

mkdocs/utils.py
```diff
@@ -272,7 +281,160 @@ def path_to_url(path):
...
+    return (html_content, table_of_contents, meta)
+
+
 def get_theme_names():
     """Return a list containing all the names of all the builtin themes."""
 
     return os.listdir(os.path.join(os.path.dirname(__file__), 'themes'))
+
+"""
+# Relative Path Markdown Extension
```
d0ugal: I really don't want this to be dumped in here. Why move it?

ugocupcic: There were some circular dependencies otherwise :S

mkdocs/tests/build_tests.py
```diff
@@ -11,12 +11,13 @@
 from mkdocs import build, nav, config
 from mkdocs.exceptions import MarkdownNotFound
 from mkdocs.tests.base import dedent
+from mkdocs.build import convert_markdown
```
d0ugal: What was wrong with using it the way it was used before as a property of build?

mkdocs/utils.py
d0ugal: hmm, I think we need to find a better solution for that.

mkdocs/tests/build_tests.py
ugocupcic: circular dependencies as well I'm afraid

d0ugal: Hmm, that sounds weird to me. If we are getting this many circular imports we are doing something wrong.

d0ugal:
If the circular import hacking is due to looking the addition of looking at metadata, then we should either not do that or find another way to do that.

mkdocs/utils.py
ugocupcic: I'll check for that one it might not be needed anymore. give me 1min

ugocupcic:
@d0ugal sorry I'm having a bad day :D Fixed most of the problems you commented on. No more circular dependency problem. The changeset is much smaller now.

d0ugal:
Phew, I was getting a bit worried there :) that looks far batter. I'll wait for the build to finish, give it a second quick check over and then I think it is good to merge. Thanks!

d0ugal:
Looks like a few PEP8 errors, then it should be good.

ugocupcic:
@d0ugal should be ready now

d0ugal:
@ugocupcic Great, I'm going to merge this with `git merge --squash` to tidy up the history, otherwise you show up as a git blame for all of `relative_path_ext.py` etc. The downside being is that you wont show the author, if you want to rebase and squash the commits I'm happy to wait.

If I don't hear anything I'll go ahead and do it in the morning.

ugocupcic:
@d0ugal I'll push a new branch in a few minutes to cleanup that history.

ugocupcic:
created #497.

PR #485, opened 2015-04-28
Preserve order of Python Markdown extensions

facelessuser:
Python Markdown inclusion order can be important when including various
custom extensions.  Conflicts can arise as to where extensions get
inserted in the workflow.  By preserving order of how extensions are
defined in the settings file, this can allow a user to control
insertion order and work around these issues.  As Py26 supported is
noted to be removed in the future, I saw no reason to add additional
dependencies for OrderedDict.  Try catch statements can be removed when
Py26 support is removed.

### DIFF
--- a/mkdocs/tests/utils_tests.py
+++ b/mkdocs/tests/utils_tests.py
@@ -67,3 +67,33 @@ class UtilsTests(unittest.TestCase):
         for path, expected_result in expected_results.items():
             urls = utils.create_media_urls(site_navigation, [path])
             self.assertEqual(urls[0], expected_result)
+
+    def test_yaml_load(self):
+        try:
+            from collections import OrderedDict
+        except ImportError:
+            # Don't test if can't import OrderdDict
+            # Exception can be removed when Py26 support is removed
+            return
+
+        yaml_text = '''
+test:
+    key1: 1
+    key2: 2
+    key3: 3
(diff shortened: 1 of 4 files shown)


facelessuser:
Looks like I'm out of sync.  If the pull looks good, I can rebase.  I'll wait for feedback.

mkdocs/tests/utils_tests.py
```diff
@@ -67,3 +67,33 @@ def test_create_media_urls(self):
...
+
+    def test_yaml_load(self):
+        try:
+            from collections import OrderedDict
+        except ImportError:
+            # Don't test if can't import OrderdDict
+            # Exception can be removed when Py26 support is removed
+            return
+
+        yaml_text = '''
```
waylan: To be consistent with the rest of mkdocs' tests, you might want to use [dedent][1] here. You can see many examples in [build_tests.py][2].

Otherwise, this looks great to me. Of course, approval is up to @d0ugal.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/base.py#L7
[2]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/build_tests.py#L42

facelessuser:
I'm going to go ahead and rebase while I still feel like it.

PR #486, opened 2015-04-28
Preserve order of Python Markdown extensions

facelessuser:
Python Markdown inclusion order can be important when including various custom extensions.  Conflicts can arise as to where extensions get inserted in the workflow.  By preserving order of how extensions are
defined in the settings file, this can allow a user to control insertion order and work around these issues.  As Py26 supported is noted to be removed in the future, I saw no reason to add additional dependencies for OrderedDict.  Try catch statements can be removed when Py26 support is removed.

### DIFF
--- a/mkdocs/tests/utils_tests.py
+++ b/mkdocs/tests/utils_tests.py
@@ -4,6 +4,7 @@
 import unittest
 
 from mkdocs import nav, utils
+from mkdocs.tests.base import dedent
 
 
 class UtilsTests(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


mkdocs/tests/utils_tests.py
```diff
@@ -67,3 +68,34 @@ def test_create_media_urls(self):
         for path, expected_result in expected_results.items():
             urls = utils.create_media_urls(site_navigation, [path])
             self.assertEqual(urls[0], expected_result)
+
```
d0ugal: Rather than try/catching in the unittest I'd prefer:

    @unittest.skipIf(sys.version_info[:2] == (2, 6))

d0ugal:
One mine comment, otherwise happy to merge this. Thanks!

mkdocs/tests/utils_tests.py
facelessuser: I don't think the Py26 unittest lib has `skipIf`.  Seems to be new in Py31 https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures.

d0ugal: Ah, of course, sorry. It is in py27 but not py26, so nevermind!

facelessuser: Yeah, I was just about to correct about the Py31 to Py27.  But yeah, it would have been much cleaner.

d0ugal:
Thanks again!

PR #496, opened 2015-05-03
Configuration Refactor

d0ugal:
This change breaks up the config parsing into a set of validators. This
makes it somewhat easier to follow and add extra checks. As a result we
now also catch a few config errors we wouldn't have before.

The code probably still needs some tidying up, but it works well.


waylan:
This looks like a great start. Obviously, the failing tests will need to be cleaned up, but I think this has a lot of potential. As things progress, more stuff can be moved into the validators. For example, the code which splits the `markdown_extensions` from the extension configs (currently in build.py) could go in a validator. Although, that raises an interesting question: To do so, an additional config setting would need to be created which holds the extension configs. Is there a way to make that setting private so that it cannot be set by the user in their config file (setting it would raise an error), but it would be available internally? I suppose a custom validator could be created for it which raises such an error, but how do you make sure that validator runs before the validator on `markdown_extensions` which would then populate that private setting? Really just thinking out loud here.

d0ugal:
> For example, the code which splits the markdown_extensions from the extension configs (currently in build.py) could go in a validator. 

That's an interesting idea, I think it makes a lot of sense to do this. We probably need to rename validators tho', perhaps they could just be called ConfigOptions as they do validation and post processing too.

> To do so, an additional config setting would need to be created which holds the extension configs. Is there a way to make that setting private so that it cannot be set by the user in their config file (setting it would raise an error), but it would be available internally?

I could do with a better way of handling this too.

https://github.com/mkdocs/mkdocs/pull/496/files#diff-75ea449ab275270869c65fcd82bdbcf8R15

At the moment if the user provided this it would be overwritten, given it is a reference to the config file the user is providing they can't ever give me anything useful here :)

It would be easy to handle something like this in the current code by making a validator that errors if it does exist and then handle the required work in a `post_process` function which is called after all the validation is done.

> I suppose a custom validator could be created for it which raises such an error, but how do you make sure that validator runs before the validator on markdown_extensions which would then populate that private setting? Really just thinking out loud here.

I needed a two pass, so the `post_process` is a hacky way to add this. It would look something like...

```python
class MarkdownExtension(BaseValidator):

    def check(self, value):

        if value is not None:

            raise ValidationError("..")

    def post_process(self, config, key_name):

        config['markdown_extensions_extra'] = {}
```

However, I wonder if the concepts should be separated, perhaps into two dictionaries one that holds the config, one that holds internal state and then a session object could bind them together and be passed around.

d0ugal:
I think I'm happy enough with this to land it now. @waylan or anyone else want to review before I do?

I expect this will need follow up work, but it should all be much more minor.

mkdocs/config/config_options.py
```diff
@@ -0,0 +1,333 @@
...
+            elif self.required:
+                raise ValidationError("Required configuration not provided.")
+
+        return self._validate(value)
+
+    def _validate(self, value):
+        """
+        Perform validation for a value.
+
+        The _validate method should be implemented by subclasses.
```
waylan: You have this backwards. The private method should be the one called internally and the public method should be the one overridden by subclasses. Switch `validate` and `_validate`.

d0ugal: The reason for this, is that I didn't want to have to do the following on each subclass.

```python
def validate(self, value):
    value = super(CLASS, self)._validate(value)
    if value is None:
        return
```

So, validate only calls the private subclassed method if the value isn't None.

mkdocs/config/config_options.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        return value
+
+    def post_process(self, config, key_name):
+        """
+        After all options have passed validation, perform a post process to
```
waylan: Hey, cool! I just realized all of the `validate` methods run before any of the `post-process` methods run. This makes things easier when order matters. I like it.

d0ugal: Yeah, that's why it gets the full config too. So it can do something based on another value. It's slightly limited, if you require the result of another `post_process` then you are outta luck :) but good enough for the cases I had.

mkdocs/config/base.py
```diff
@@ -0,0 +1,86 @@
...
+        return failed, warnings
+
+    def validate(self):
+
+        failed, warnings = self.validate_update(self)
+
+        for key in set(self._schema.keys()) - set(self.keys()):
+            config_option = self._schema[key]
+            if config_option.is_required():
+                failed.append((key, "Required configuration not provided."))
```
waylan: Why is there a "required check" here? You also have a "required check" in the private `validate` method of `BaseConfigOption` (see line 43 of config_options.py)

mkdocs/config/config_options.py
d0ugal: I do feel like I am missing an obvious way to do this in a better way, but I can't quite think of it.

mkdocs/build.py
```diff
@@ -149,7 +150,7 @@ def build_sitemap(config, env, site_navigation):
 
 def build_template(template_name, env, config, site_navigation=None):
 
-    log.debug("Building {0} page".format(template_name))
+    log.debug("Building %s page", template_name)
```
waylan: Was this intentional? You are using `string.format` throughout the `config_options.py` file. Not sure why it can't be used here?

mkdocs/config/base.py
d0ugal: Good spot, I think that was left over from a previous approach. Will update.

mkdocs/config/config_options.py
facelessuser: What I think he meant was that the subclasses should override a method called `validate`.  Flip the names of what is exposed and what is not.  Subclasses should override private methods.  You shouldn't have to use `super` as you are still using things the same.

mkdocs/build.py
d0ugal: Yeah, technically speaking this is the correct way to use the logging API. It should be passed in unformatted and interpolation left to the logging function.

mkdocs/config/config_options.py
d0ugal: Oh, of course. I get it. Sorry, it's been a long day.

mkdocs/config/config_options.py
```diff
@@ -0,0 +1,333 @@
...
+class URL(BaseConfigOption):
+    """
+    URL Config Option
+
+    Validate a URL by requiring a scheme is present.
+    """
+
+    def _validate(self, value):
+
+        parsed_url = six.moves.urllib.parse.urlparse(value)
```
waylan: Should this be in a try/except block? If `urlparse` fails, what error do you want to be given?

mkdocs/config/config_options.py
waylan: Thanks @facelessuser, you have understood the intention of my suggestion correctly.

mkdocs/config/config_options.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+
+    def __init__(self, exists=False, **kwargs):
+        super(Dir, self).__init__(**kwargs)
+        self.exists = exists
+
+    def _validate(self, value):
+
+        if self.exists and not os.path.isdir(value):
+            raise ValidationError("The path {0} doesn't exist".format(value))
```
waylan: If `exists` is `False` then no validation happens? At least confirm it is a string, not a list, dict, date, ... or any other type supported by YAML. 

Also, why not set the value to an absolute path here? Then the `post_process` method on all the subclasses won't need to call that continually.

mkdocs/config/config_options.py
d0ugal: Yup, I'll add that.

mkdocs/config/config_options.py
d0ugal: Makes sense, I'll do both. Some of this didn't happen as I was trying to draw a line on this piece of work :) There is no validation for this at the moment, I expect to improve them all over time somewhat, but I should add a better baseline here.

mkdocs/config/config_options.py
d0ugal: Hmm, I must still be missing something. Is this quick example what you mean? Because that would require the subclasses to always handle the Nones which is a pain.

```python
class BaseValidator(object):

    def validate(self, value):
        # default to just the basic checks for required and defaults.
        return self._validate(value)

    def _validate(self, value):
        if value is None:
            if self.default is not None:
                value = self.default
            elif not self.required:
                return
            elif self.required:
                raise ValidationError("Required configuration not provided.")

        return value

class SubclassValidator(BaseValidator):

    def validate(self, value):

        value = self._validate(value)
        if value is None:
            return

        # do the sub class checks.
```

facelessuser: Think of `_validate` more as a protected function instead of a private function.  In python on think of `__validate` more as a private function as this is when python will mangle the name.  I know this may be more of a C++ concept, but let's roll with it :).

Anyways, a subclass really shouldn't override a protected function, but it can call a protected function, and anything outside of a subclass shouldn't call it.  Now a private function is used by the base class only and is not inherited (or in this case called by the subclass).

Does that make sense?

d0ugal: > Anyways, a subclass really shouldn't override a protected function

Sure, that makes complete sense. I'm just trying to figure out how to make the change in a nice way, adding a check for None in every subclass is somewhat ugly.

I could just have `validate` on the base class and `do_validation` or something on the subclasses, that removed the overrideing of the protected methods by just using another naming convention.

waylan: Right, all you need to do is rename `validate` to `_validate` and `_validate` to `validate`. Then, the one place in the code were `validate` is currently called, change that to call `_validate`. The entire change only effects three lines - oh accept for having all your subclasses override `validate` rather than `_validate`.

facelessuser: Oh, I didn't realize you literally swapped the functions.  What @waylan said,  it's a rename swap that was desired.

waylan: In other words, edit line 52 of `config/base.py`:

    - update[key] = config_option.validate(value)
    + update[key] = config_option._validate(value)

Line 29 of `config/config_options.py`:

    - def validate(self, value):
    + def _validate(self, value):

and Line 48  of `config/config_options.py`:

    - def _validate(self, value):
    + def validate(self, value):

Then in each subclass do:

    - def _validate(self, value):
    + def validate(self, value):

d0ugal: > Then, the one place in the code were validate is currently called, change that to call _validate

Aha, I see. I didn't want to call a private/protected/whatever method. I find that worse than overwriting one with a subclass.

I guess I'll go with the idea I comments above, which is something of a compromise between the two. https://github.com/mkdocs/mkdocs/pull/496#discussion_r29963272

waylan: The way I see it, if it is all part of the same code base (not just the same class) then it is okay to call the protected method internally. However, the public facing API should never be private/protected. In other words, third party code should never call or override in a subclass any method which starts with an underscore (Python's special double underscore methods being the exception). However, with my suggested change, the protected method is only ever used by mkdocs internally (although, admittedly, not from within the same class). If it ever changes in the future, third party code (or subclasses) would never need to change. Who cares if the internal mkdocs code needs to change (not including subclasses as that gets repeated many times over). 

But if you want to use `do_validation` instead and avoid the underscore altogether, then that works too. Even then, I would name the method that gets overridden `validate` and use `do_validation` for the internal method (what I was suggesting you call `_validate`).

Oh and @facelessuser is correct, a single underscore generally indicates a protected method. - although that is not actually enforced by Python. It is just good practice. Double underscore methods are private which is enforced. Sorry I used the wrong word earlier.

d0ugal: I probably wont call it `do_valiadation`, that name sucks. I need to think of another, but I think I'll take that approach.

I really avoid calling protected methods outside of the same class, or subclass, but this is certainly a grey area in Python.

facelessuser: The original recommendation was based on the original code.  It wasn't understood that there was an aversion to call `_validate`.  I personally  find overriding `_validate` more evil than calling it, but personal preference and all that ;).  I think it just depends on how people approach the language, my background in coding (C/C++) steers me towards the latter. But I get some people in python view `_` as don't touch, unless your using a named tuples, where they were force to expose the class methods that way :).

d0ugal: If you are curious, I scope it like this and it's the pattern used in most projects I'm involved with.

- Underscore class methods within the class and subclasses only
- Underscore functions (defined outside a class) used within the file only
- Underscore files (like `_file.py`) within the project only.

I rarely used double underscore anywhere, other than `__init__` an so on.

waylan: Yeah, I don't care much for `do_*` names either. But in this case, `validate` defines how the validator works, but `do_validation` runs the validation, which seems reasonable. You could use `run_validation` or `run_validators`. Just `run` works too.

facelessuser: Thanks for explaining.  I realize object oriented approaches can be completely different in other languages.  As there is no strict rule on how to use underscore methods, I guess it really depends on the community of authors you work with.  It's nice to understand different approaches.  I will definitely keep this approach in mind when looking at other people's Python projects moving forward.

d0ugal: @waylan Yup. makes sense. I might just go with one of those :)

@facelessuser You are welcome, there are few hard rules here so my way is just one approach. pylint will complain if you access private properties on an object tho, that's one of the reasons for my bias.

mkdocs/config/config_options.py
```diff
@@ -0,0 +1,339 @@
...
+            if self.default is not None:
+                value = self.default
+            elif not self.required:
+                return
+            elif self.required:
+                raise ValidationError("Required configuration not provided.")
+
+        return self.run_validatuon(value)
+
+    def run_validatuon(self, value):
```
waylan: Typo: `run_validation` not `run_validatuon` -- needs to be fixed in all the subclasses too.

I guess I expected this method to be called `validate` and the method currently called `validate` to be `run_validation`, but whatever.

d0ugal:
@waylan @facelessuser Thanks both for your input. I'm going to merge this and continue iterating on it in master over time. I think it's a big enough improvement and a big enough PR that I don't want to continue in a branch much longer and make bigger.

However, I'd still be very interested in more feedback - so if you have anything, either comment here or open a new issue. Thanks again!

PR #497, opened 2015-05-05
Fixes #316 - titles

### LINKED ISSUE #316: If no title is provided in the mkdocs.yml use the title in the markdown file

ugocupcic:
History is squashed.


ugocupcic:
I had a few commits on my master so they still appear. Hopefully it's fine with you. Let me know otherwise.

d0ugal:
Thanks for following this up and pushing it forward!

d0ugal:
Hmm, turned out I should have looked more closely. This doesn't work at all. The `file_to_title` function is only passed the filename, but the code in this PR expects the full path so it can open the file.

ugocupcic:
@d0ugal any idea how to get the full path to be able to parse the file?

d0ugal:
@ugocupcic It is surprisingly complicated :( I had a quick look, about 6 methods need to be updated to pass it down the chain and that starts to get really nasty.

I'll give it some thought, but I think we might need to find a new approach. Sorry, it turned out to be somewhat harder than I first estimated.

ugocupcic:
Might be easier to be able to parse that `mkdocs.yml` even if we have the documentation at the root of the repository instead of in this separate docs folder?

PR #498, opened 2015-05-05
Add a CLI option to output in a different directory

AlexPerrot:
Hi,

I use mkdocs with CMake and wanted to be able to give an output directory on the command line. This adds flexibility to the build process.
I added an `--outputdir` option to the `build` command. If not given, it behaves as it is now and compiles into the current directory.

Comments are welcome :)

### DIFF
--- a/mkdocs/cli.py
+++ b/mkdocs/cli.py
@@ -34,6 +34,8 @@ strict_help = ("Enable strict mode. This will cause MkDocs to abort the build "
                "on any warnings.")
 theme_help = "The theme to use when building your documentation."
 theme_choices = utils.get_theme_names()
+site_dir_help = ("The directory where the output HTML and other files are created."
+                 " Overrides the value from the config file.")
 
 
 @click.group()
@@ -66,12 +68,14 @@ def serve_command(dev_addr, config_file, strict, theme):
 @click.option('--config-file', type=click.File('rb'), help=config_file_help)
 @click.option('--strict', is_flag=True, help=strict_help)
 @click.option('--theme', type=click.Choice(theme_choices), help=theme_help)
-def build_command(clean, config_file, strict, theme):
+@click.option('--site-dir', type=click.Path(), help=site_dir_help)
+def build_command(clean, config_file, strict, theme, site_dir):
     """Build the MkDocs documentation"""
     build.build(load_config(
         config_file=config_file,
         strict=strict,
-        theme=theme
+        theme=theme,
+        site_dir=site_dir
     ), clean_site_dir=clean)
 
 


waylan:
@AlexPerrot, I notice that the config setting is named `site_dir` yet you named the command line option `builddir`. That seems confusing to me. I would prefer the names to be the same. Granted, "build" it a clearer name than "site," but site has been used for existing projects and consistency is more important IMO. Besides, the help string can clear up what the "site_dir" is.

Second, I notice that you had to jump through a couple hoops because `load_config` doesn't accept a `site_dir` keyword. I wonder if perhaps `load_config` should accept `**kwargs` where the keys in `kwargs` could be any keys in the config. Note that this is not a criticism of the provided patch, but a general observation about the internal API and what will make it easier to make similar additions in the future. @d0ugal, any thoughts on this?

mkdocs/cli.py
```diff
@@ -66,13 +67,18 @@ def serve_command(dev_addr, config_file, strict, theme):
...
-    build.build(load_config(
+    config = load_config(
         config_file=config_file,
         strict=strict,
         theme=theme
-    ), clean_site_dir=clean)
+    )
+    if(outputdir is not None):
+        site_dir = config['site_dir']
+        config['site_dir'] = os.path.join(outputdir, site_dir)
```
d0ugal: An output path option should replace the `site_dir`, not be joined with it. I think that behavior is confusing.

d0ugal:
I think this change should just be:

```python
@cli.command(name="build")
@click.option('--clean', is_flag=True, help=clean_help)
@click.option('--config-file', type=click.File('rb'), help=config_file_help)
@click.option('--strict', is_flag=True, help=strict_help)
@click.option('--theme', type=click.Choice(theme_choices), help=theme_help)
@click.option('--site-dir', help="Help text here to make it as obvious as possible.")
def build_command(clean, config_file, strict, site_dir):
    """Build the MkDocs documentation"""
    build.build(load_config(
        config_file=config_file,
        strict=strict,
        theme=theme,
        site_dir=site_dir
    ), clean_site_dir=clean)
```

mkdocs/cli.py
AlexPerrot: I was interpreting the `site_dir` as the name of the site directory and `outputdir` as a prefix to build to. But looking at the configuration, I see that `site_dir` can be a full or relative path, so problems could indeed arise in those cases.

AlexPerrot:
Thank you for your comments.
My first intention was to redirect the build to another directory on the command line, and not *replace* the user configuration.
On the other hand, I don't think the name of the `site_dir`is that important that it should be kept at all cost. After all, if the user is specifying a new dir on the command, why keep the configuration ? I was maybe overthinking that a little.
Do you want to replace the commits with something similar to what @d0ugal proposed ?

This seems to be a case of overriding the configuration with CLI options. Could other options be also concerned ?

d0ugal:
> On the other hand, I don't think the name of the site_dir is that important that it should be kept at all cost

While I agree the name isn't ideal, we are not going to change it now and we don't want two names for the same thing.

> This seems to be a case of overriding the configuration with CLI options. Could other options be also concerned ?

I don't think so, `site_dir` is the only option that changes the output directory.

AlexPerrot:
> > On the other hand, I don't think the name of the site_dir is that important that it should be kept at all cost

> While I agree the name isn't ideal, we are not going to change it now and we don't want two names for the same thing.

Sorry, that's absolutely not what I meant :). I only meant that the value entered by the user in the configuration file for `site_dir` and especially the last directory name, could very well be replaced by what's given on the command line.
Sorry for the confusion.

Anyway, i made another commit with what you proposed.

PR #502, opened 2015-05-06
Start getting the 0.13 release notes together

d0ugal:


docs/about/release-notes.md
```diff
@@ -8,13 +8,92 @@ To upgrade MkDocs to the latest version, use pip:
...
+[extra_templates]: /user-guide/configuration/#extra_templates
+[global variables]: /user-guide/styling-your-docs/#global-context
+
+### Other Changes and Additions
+
+* Add support for [Markdown extension configuration options]. (#435)
+* MkDocs now ships Python [wheels]. (#486)
+* Only include the build date and MkDocs version on the homepage. (#490)
+* Generate sitemaps for documentation builds. (#436)
+* Automatically take titles from the first H1 on a Makrdown document if they
```
peter1000: typo: Makrdown

d0ugal: Thanks!

maemual:
Bugfix? Such as #532

d0ugal:
@maemual Yup, quite a few need to be added still.

waylan:
If the release notes had been included in the source already, I would have updated them when I made my pull requests. In fact, many projects make it a practice to up the version number to the next number (with a "dev" status) and create the release notes right then. As each feature gets added/fixed after that, the release notes get updated along with everything else for the feature/fix (code, tests, docs). Of course, right before a release, a through edit of the release notes may help spruce things up, but it beats doing it from scratch. Just a suggestion.

d0ugal:
I have bumped the version number to `0.13.0.dev` already. I have been meaning to follow a stricter process with this, as the number of contributors increase it makes more sense too. I'll try and do a better job for `0.14.0`.

PR #513, opened 2015-05-09
Match the signature from super

d0ugal:


d0ugal:
Was curious to see if Landscape would be happy if I matched the super signature. But that requires using dict as a variable name, so it complains about that instead :smile: Never happy!

Issue #201, opened 2014-11-07
Pull alternate themes into separate package.


tomchristie:
We probably shouldn't try to maintain any more than two themes in the core package.
I'd recommend that we conside doing the following:

* [ ] Pull the bootswatch themes into a separate package. (We might even consider rebasing the git history after that as it's unnecessarily large)
* [x] Remove mention of the non-existent "ghostwriter" theme.

d0ugal:
Done! https://github.com/mkdocs/mkdocs/commit/ff05df30dd2de8840fb185363afa41a5517a69ef

PR #515, opened 2015-05-10
This is a really confusing test failure, I can't figure out what is going wrong.

d0ugal:
I was making this change, to start warning users about #201. We don't have a concrete plan yet but the earlier we start warning them, the earlier we can start making the move.

I didn't get very far with it, because It causes this error, which almost makes sense except that it doesn't at all...

```
FAIL: test_extra_context (mkdocs.tests.build_tests.BuildTests)
  File "/home/dougalmatthews/Code/python/mkdocs/mkdocs/tests/build_tests.py", line 359, in test_extra_context
    self.assertEqual(config.validate(), ([], []))
AssertionError: Tuples differ: ([], [('theme', "The theme 'cosmo' will be remo[97 chars]s")]) != ([], [])

First differing element 1:
[('theme', "The theme 'cosmo' will be removed in an upcoming MkDocs release. See http://www.mkdocs.org/about/release-notes/ for more details")]
[]

+ ([], [])
- ([],
-  [('theme',
-    "The theme 'cosmo' will be removed in an upcoming MkDocs release. See "
-    'http://www.mkdocs.org/about/release-notes/ for more details')])
```

There are two big issues here.

- This test doesn't provide a theme, so it should use the default, `mkdocs`. The error is about cosmo.
- cosmo is only used in [this test](https://github.com/mkdocs/mkdocs/blob/cb3a24cff44f22bfd9295bb3decedb855c6ba3a5/mkdocs/tests/config/config_tests.py#L120), which should be totally unrelated. Updating that test causes the one above to fail with the new theme.

So, there must be some leaky global state. I can't figure out where it is coming from. Ideas?

### DIFF
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -29,8 +29,6 @@ class Config(six.moves.UserDict):
         self.set_defaults()
         self.user_configs = []
 
-        self.config_file_path = None
-
     def set_defaults(self):
         """
         Set the base config by going through each validator and getting the
@@ -72,7 +70,7 @@ class Config(six.moves.UserDict):
 
         return failed, warnings
 
-    def update(self, patch):
+    def load_dict(self, patch):
 
         if not isinstance(patch, dict):
             raise exceptions.ConfigurationError(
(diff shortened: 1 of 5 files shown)


d0ugal:
Managed to fix it, I don't quite understand what was wrong but tidying up the test sorted things out.

lbenet:
While I understand the rationale behind #201, it is worth to document how to include the bootswatch themes, which are already a separate package. This may even speed up the deprecation process for those themes.

d0ugal:
Yup, that is a good point. It isn't documented because it doesn't work yet. It should be possible with the next release and then it will be documented. A release after that will remove the themes you see a warning about by default.

PR #516, opened 2015-05-11
Adding custom commit message option to gh_deploy

mbacho:
Adds an option of configuring the commit message to be set when using gh_deploy

### DIFF
--- a/mkdocs/cli.py
+++ b/mkdocs/cli.py
@@ -122,14 +122,14 @@ def json_command(clean, config_file, strict, site_dir):
 @cli.command(name="gh-deploy")
 @click.option('--clean', is_flag=True, help=clean_help)
 @click.option('--config-file', type=click.File('rb'), help=config_file_help)
-@click.option('--commit-message', is_flag=False, help=commit_message_help)
-def gh_deploy_command(config_file, clean, commit_message):
+@click.option('--message', '-m', help=commit_message_help)
+def gh_deploy_command(config_file, clean, message):
     """Deply your documentation to GitHub Pages"""
     config = load_config(
         config_file=config_file
     )
     build.build(config, clean_site_dir=clean)
-    gh_deploy.gh_deploy(config, commit_message=commit_message)
+    gh_deploy.gh_deploy(config, message=message)
 
 
 @cli.command(name="new")
--- a/mkdocs/cli.py
+++ b/mkdocs/cli.py
@@ -40,6 +40,7 @@ site_dir_help = "The directory to output the result of the documentation build."
 reload_help = "Enable and disable the live reloading in the development server."
 commit_message_help = ("A commit message to use when commiting to the "
                        "Github Pages remote branch")
+remote_branch_help = "The remote branch to commit to for Github Pages"
 
 
 @click.group()
@@ -123,13 +124,14 @@ def json_command(clean, config_file, strict, site_dir):
 @click.option('--clean', is_flag=True, help=clean_help)
 @click.option('--config-file', type=click.File('rb'), help=config_file_help)
 @click.option('--message', '-m', help=commit_message_help)
-def gh_deploy_command(config_file, clean, message):
+@click.option('--branch', '-b', help=remote_branch_help)
+def gh_deploy_command(config_file, clean, message, branch):
     """Deply your documentation to GitHub Pages"""
     config = load_config(
         config_file=config_file
     )
     build.build(config, clean_site_dir=clean)
-    gh_deploy.gh_deploy(config, message=message)
+    gh_deploy.gh_deploy(config, message=message, branch=branch)
 
 
 @cli.command(name="new")
--- a/mkdocs/cli.py
+++ b/mkdocs/cli.py
@@ -38,6 +38,8 @@ theme_help = "The theme to use when building your documentation."
 theme_choices = utils.get_theme_names()
 site_dir_help = "The directory to output the result of the documentation build."
 reload_help = "Enable and disable the live reloading in the development server."
+commit_message_help = ("A commit message to use when commiting to the "
+                       "Github Pages remote branch")
 
 
 @click.group()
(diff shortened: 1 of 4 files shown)


waylan:
Maybe there is something I'm missing here, but I don't follow this change. First, shouldn't the option passed to `ghp-import` be `-m` for the message, not `-b` (which is used for the branch)? Second, why is the message dependent on the config? Third, shouldn't the actual message be passed in on the command line, presumably by adding a new option to the [`cli.gh-deploy`][1] command?

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/cli.py#L120

d0ugal:
Yup, I agree with @waylan's comments. It should be added in the cli.py file and passed through.

mbacho:
@waylan @d0ugal I agree with the first issue. It was my bad. 
I also agree on the second and third issues. I updated the patch to get custom commit messages from `cli.gh_deploy` rather than the config.

d0ugal:
Cool, looks good to me, will give it a second look over later and let anyone else that wants to review have a look.

mkdocs/gh_deploy.py
```diff
@@ -5,19 +5,25 @@
...
         log.info('Cannot deploy - this directory does not appear to be a git '
                  'repository')
         return
 
-    log.info("Copying '%s' to `gh-pages` branch and pushing to GitHub.",
-             config['site_dir'])
+    command = ['ghp-import', '-p', config['site_dir']]
+
+    remote_branch = config.get('remote_branch', 'gh-pages')
+    command.extend(['-b', config['remote_branch']])
```
waylan: Either this needs to be in an if statement (`if config['remote_branch'] is not None:`) or it needs to pass in the `remote_branch` var obtained in the previous line. That said, I don't see a `remote_branch` config setting defined in the [defaults][1] (which means it will never exist as unknown keys will raise an error).  I suspect this may have been a bug in the existing code as well (at least after 9d3594c landed).

And I wonder if this is the right place to define the default. Perhaps the default should be left to the ghp-import tool. Thus the `if not None` suggestion -- a branch name should only be passed if one is defined.

I notice that you do use the `remote_branch` var later in the log message. Perhaps rather than assuming the default is "gh-pages", when a branch is not explicitly defined the message should simply say: "Copying <site_dir> to the default branch..." rather than actually naming the branch. In fact, if the `remote_branch` var is only used for the log message, seems like it should be calculated there. That way future reviewers of the code won't wonder why it is not used when building the command.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/defaults.py

d0ugal: Oh! The remote branch functionality will be currently broken as a side effect of #450. The old CLI parser had a one to one relationship between config settings and CLI arguments. This highlights the lack of test coverage in a few places like this.

So - we should add a remote branch argument like this PR adds message. @mbacho would you be happy to do that? Otherwise I can add it afterwards - I did break it after all :smile:

mkdocs/cli.py
```diff
@@ -120,13 +122,14 @@ def json_command(clean, config_file, strict, site_dir):
 @cli.command(name="gh-deploy")
 @click.option('--clean', is_flag=True, help=clean_help)
 @click.option('--config-file', type=click.File('rb'), help=config_file_help)
-def gh_deploy_command(config_file, clean):
+@click.option('--commit-message', is_flag=False, help=commit_message_help)
```
waylan: I realize this is bikeshedding, but why not follow Git's pattern (like most Git add-ons do) and use `--message`. I would think all Git users will expect either that or `-m`. Of course, Click can support both, which is my personal preference (yes I realize mkdocs does not in most cases - to my annoyance). The point here is that `--message` would be the least surprising and the easiest to remember for users. The help text identifies it specifically as a *commit* message.

d0ugal: Pull requests welcome to add short names :)

--message does sound like the best choice. Let's use that. You can add the short version of you like, otherwise it can be added with all the others later.

mkdocs/gh_deploy.py
mbacho: @waylan nice catch! I've updated the [defaults](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/defaults.py) in config. About the log, I would prefer the log informs the user what branch is being used, just for transparency. Thus the use of the branch name in the log.

@d0ugal since the `remote_branch` is less likely to change, unlike the commit message, I think config setting would suffice. However, I could make another PR to add it to CLI arguments, and maybe add tests for the `gh_deploy` module.

mkdocs/cli.py
mbacho: I agree `--message` or `-m` is better than `--commit-message`. I've effected the change.

d0ugal:
Okay, I mostly agree that remote branch can be a config option. However, we also need it to be a command line argument and if it is going to be a config option we also need to document it.

d0ugal:
Ignore the errors on Travis for now, they are unrelated to this change. I'm looking into it.

mbacho:
@d0ugal I've added a CLI optional argument for remote branch and a fall back to the one set in config

mkdocs/cli.py
```diff
@@ -120,13 +123,15 @@ def json_command(clean, config_file, strict, site_dir):
 @cli.command(name="gh-deploy")
 @click.option('--clean', is_flag=True, help=clean_help)
 @click.option('--config-file', type=click.File('rb'), help=config_file_help)
-def gh_deploy_command(config_file, clean):
+@click.option('--message', '-m', help=commit_message_help)
+@click.option('--branch', '-b', help=remote_branch_help)
```
d0ugal: Call it `--remote-branch` to make it clear and that's the flag we have in the current version that got lost in a recent refactor.

mkdocs/cli.py
```diff
@@ -120,13 +123,15 @@ def json_command(clean, config_file, strict, site_dir):
...
+def gh_deploy_command(config_file, clean, message, branch):
     """Deply your documentation to GitHub Pages"""
     config = load_config(
         config_file=config_file
     )
     build.build(config, clean_site_dir=clean)
-    gh_deploy.gh_deploy(config)
+    gh_deploy.gh_deploy(config, message=message, branch=branch)
```
d0ugal: Rather than passing the parameter here, it would be passed to load_config (see the other commands, this overwrites the config with CLI parameters and keeps them consistent).

d0ugal:
Two tiny comments, and then I think this looks great - thanks very much!

mbacho:
@d0ugal changes made on the two issues.

d0ugal:
Looks great, thanks! Just waiting for CI to pass and then I'll merge.

PR #523, opened 2015-05-14
Use io.open

d0ugal:
This fix is a bit of a guess, no idea why it started failing on Travis. Passes locally for me.


d0ugal:
I didn't read the exception correctly - the error is in ghp-import.

PR #530, opened 2015-05-16
Move builtin themes to external packages other than MkDocs and ReadTheDocs

d0ugal:
This needs a few remaining things done:

- [x] Fix the to theme issues https://github.com/mkdocs/mkdocs-bootswatch/issues/1 and https://github.com/mkdocs/mkdocs-bootstrap/issues/1
- [x] Move MkDocs theme issues to the relevant repos
- [x] Document how to make theme packages
- [x] Add to the release notes


mkdocs/utils.py
```diff
@@ -299,10 +300,23 @@ def convert_markdown(markdown_source, extensions=None, extension_configs=None):
...
 
 
+def get_themes():
+    """Return a dict of theme names and their locations"""
+
+    themes = {}
+
+    for theme in iter_entry_points(group='mkdocs.themes', name=None):
+        theme = theme.load()
+        location = os.path.dirname(theme.__file__)
```
waylan: I like this. The theme author does not need to write any Python code. In sphinx, the theme author needs to write some code in the `__init__.py` file which returns the location. This is much better. Just include the empty file at the root of the theme, and it becomes available.

mkdocs/utils.py
```diff
@@ -299,10 +300,23 @@ def convert_markdown(markdown_source, extensions=None, extension_configs=None):
...
+        themes[theme.name] = location
```
waylan: I expect this is failing because you are replacing the entry_point (var named `theme`) with the loaded object. Before `theme = theme.load()` do `name = theme.name` or don't overwrite the `theme` var (load to a different var).

d0ugal: Yup, exactly - I just fixed this. There is a failure now I don't fully understand tho'

d0ugal:
I'm not sure what happens if two packages provide the same theme name, I can't find any documentation on this and you can't specify a duplicate in one package, so I guess I'll need to manually test and see.

waylan:
> I'm not sure what happens if two packages provide the same theme name, I can't find any documentation on this and you can't specify a duplicate in one package, so I guess I'll need to manually test and see.

Yeah, I'm not sure either. Although my understanding is that there is no restriction on duplicates. The duplicates are separated by the namespace/package that they ship in (the idea being that an entry_point named "foo" for package A would not be confused with an entry_point named "foo" for package B. Of course, when package B is a third party addon to package A, then package A can't make that distinction. As the only way to get entry_points across multiple packages is to iterate, the last entry_point named "foo" in your for loop will be the one that gets used. I am not sure what determined the order. If it is 'last installed' that is probably fine, but it my not be and I haven't worked that part out.

d0ugal:
Last installed would still be vague. Maybe we should change the theme config option to `{pageage}.{theme_name}`.

So in the case of the builtin themes, it would be `mkdocs.readthedocs` and `mkdocs.mkdocs`. I suspect many themes will end up being `custom_theme.custom_theme` which is annoying.

I'm tempted to say that the risk is really small enough that it is worth the convenience.

waylan:
I think the last paragraph in [this section][1] is relevant. Specifically, it is up to you to decide how to handle duplicates. That said, the docs for [iter_entry_points][2] states that duplicates are ordered by the package/distribution they are associated with on the PYTHONPATH.

So, if two themes exist with the same name, the second one to be found on the PYTHONPATH will always overwrite the first one in your loop. Which comes first or second depends on how each machine is configured and where each package in installed on the path. You will even find different configurations on different machines using the same version of the same OS. In other words, it is completely unpredictable.

You could check for duplicates and either raise a warning or an error. At least that way a user would have a clue as to why the wrong theme is being used. Just make sure each warning/error message lists which package the duplicates are associated with so they can uninstall the one they don't want.

For the themes included with mkdocs, you can special case them to not allow duplicates. That way, others will not even bother making duplicates of the builtin themes as it will never work. Doing `iter_entry_points(group='mkdocs.themes', name='mkdocs')` should give you only the themes that ship with mkdocs.

[1]: http://pythonhosted.org/setuptools/pkg_resources.html#entry-points
[2]: http://pythonhosted.org/setuptools/pkg_resources.html#convenience-api

waylan:
> Doing iter_entry_points(group='mkdocs.themes', name='mkdocs') should give you only the themes that ship with mkdocs.
Sorry, actually `get_entry_map(dist='mkdocs', group='mkdocs.themes')` will give you the themes that ship with mkdocs.

waylan:
Perhaps something like this would work (untested):

```python
themes = {}
builtins = get_entry_map(dist='mkdocs', group='mkdocs.themes')

for theme in iter_entry_points(group='mkdocs.themes'):
    if theme.name in builtins and theme.dist.key != 'mkdocs':
        # raise an error...
    if theme.name in themes:
        # raise a warning...
    themes[theme.name] = os.path.dirname(theme.load().__file__)
```

waylan:
Another possibility (again untested):

```python
themes = {}
for theme in iter_entry_points(group='mkdocs.themes'):
    name = '.'.join(theme.dist.key, theme.name)
    themes[name] = os.path.dirname(theme.load().__file__)
```

According to the [docs][1], `theme.dist.key` is the unique lowercase name used to register a distribution with PyPI. So if the entry point for the builitin theme was `readthedocs`, then the name entered in the config file by the user would be `mkdocs.readthedocs`. And if I created a new separate distribution (named `foo`) with a `readthedocs` entry point, I would need to use `foo.readthedocs` in my config to get it to work. 

There would be no duplicates. However, the entry point would not be the full name, only the last part and that could be slightly confusing to some. Clear documentation would be required.

[1]: http://pythonhosted.org/setuptools/pkg_resources.html#distribution-attributes

waylan:
And I just realized my last suggestion was the same as yours @d0ugal.

> So in the case of the builtin themes, it would be mkdocs.readthedocs and mkdocs.mkdocs. I suspect many themes will end up being custom_theme.custom_theme which is annoying.

That is a valid point. But what if you renamed the "mkdocs" theme to "default" or "basic" (or something better). That way, in the config the user would set `theme: mkdocs.default`. Even warn theme authors in the docs about the possible name duplication. With mkdocs setting the precedent, a package with one theme would more likely follow the same pattern. And naturally, a package with multiple themes would just give a name to each one.

d0ugal:
Thanks for throwing those ideas in, I am inclined to go for [erroring if a builtin theme is used and logging a warning if a name is used twice](https://github.com/mkdocs/mkdocs/pull/530#issuecomment-103177660). Simply because this provides us the easiest transition path, all the current names are will work and I expect the most common case will be theme packages that contain one theme. We can then encourage users use the package name for the theme name, ensuring some uniqueness (assuming they register and upload to PyPI).

waylan:
@d0ugal that sounds like a reasonable conclusion and my preference as well.

d0ugal:
It also means I can probably remove the themes for 0.13, and just add them as a requirement. It might delay the release a bit, but I've been itching to do this for a long time.

facelessuser:
Nice! I think this is a good idea.

d0ugal:
I think I'm happy with this, but I want to play with it a bit more. So targeting the next release and I'll try and get 0.13 out of the door by the end of the month.

waylan:
I like it.

d0ugal:
The failure is strange. This works for me locally, so I'm not sure where I went wrong. Something appears to be different in my setup.

waylan:
Locally, I'm getting the following error:

```
$ tox -e py27-integration
GLOB sdist-make: /home/waylan/Code/mkdocs/setup.py
py27-integration inst-nodeps: /home/waylan/Code/mkdocs/.tox/dist/mkdocs-0.14.0.dev.zip
py27-integration runtests: PYTHONHASHSEED='4293062537'
py27-integration runtests: commands[0] | /home/waylan/Code/mkdocs/.tox/py27-integration/bin/python -m mkdocs.tests.integration --output=/home/waylan/Code/mkdocs/.tox/py27-integration/tmp/builds
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/waylan/Code/mkdocs/mkdocs/tests/integration.py", line 27, in <module>
    MKDOCS_THEMES = utils.get_theme_names()
  File "mkdocs/utils/__init__.py", line 382, in get_theme_names
    return get_themes().keys()
  File "mkdocs/utils/__init__.py", line 374, in get_themes
    for name, theme in themes.items())
  File "mkdocs/utils/__init__.py", line 374, in <genexpr>
    for name, theme in themes.items())
  File "/home/waylan/Code/mkdocs/.tox/py27-integration/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2338, in load
    self.require(*args, **kwargs)
  File "/home/waylan/Code/mkdocs/.tox/py27-integration/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2355, in require
    items = working_set.resolve(reqs, env, installer)
  File "/home/waylan/Code/mkdocs/.tox/py27-integration/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 827, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'mkdocs-bootswatch>=0.1.0' distribution was not found and is required by the application
ERROR: InvocationError: '/home/waylan/Code/mkdocs/.tox/py27-integration/bin/python -m mkdocs.tests.integration --output=/home/waylan/Code/mkdocs/.tox/py27-integration/tmp/builds'
___________________________________ summary ____________________________________
ERROR:   py27-integration: commands failed
```

I'm not sure why we have the following requirements:

```
mkdocs-bootstrap>=0.1.1
mkdocs-bootswatch>=0.1.0
```

Aren't those used by the themes you removed? If so, shouldn't they be removed from the mkdocs requirements? Wouldn't they instead be listed as requirements of the theme package which utilizes them?

Once I remove those dependencies, then I am getting the same error as Travis. However, I am not getting all the warnings about the deprecated themes. Anyway, I will investigate further.

waylan:
Ok, the problem is two things. First the following small patch needs to be done to ensure that the theme locations are all absolute paths:

```diff
diff --git a/mkdocs/utils/__init__.py b/mkdocs/utils/__init__.py
index 4960762..37d2a31 100644
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -370,7 +370,7 @@ def get_themes():
 
         themes[theme.name] = theme
 
-    themes = dict((name, os.path.dirname(theme.load().__file__))
+    themes = dict((name, os.path.dirname(os.path.abspath(theme.load().__file__)
                   for name, theme in themes.items())
 
     return themes
```

And secondly, the setup.py script is not copying any of the `*.html` files in the theme on install. I checked the theme dir under site_packages, and none of them exist. Looking at the setup.py script, the `get_package_data` function only copies files in a dir that does not contain a `__init__.py` file (see L41). As the `*.html` files are in that directory, it is skipping them. Why not just tell it to copy everything inside the `themes` dir? Not sure what else would be needed.

waylan:
Ok, I got the tests to pass with the following changes:

```diff
diff --git a/mkdocs/tests/integration.py b/mkdocs/tests/integration.py
index d064da3..5c0b10e 100644
--- a/mkdocs/tests/integration.py
+++ b/mkdocs/tests/integration.py
@@ -38,7 +38,7 @@ def main(output=None):
     for theme in sorted(MKDOCS_THEMES):
         project_dir = os.path.dirname(MKDOCS_CONFIG)
         out = os.path.join(output, theme)
-        command = ['mkdocs', 'build', '--site-dir', out, '--theme', theme]
+        command = ['mkdocs', 'build', '-v', '--site-dir', out, '--theme', theme]
         subprocess.check_call(command, cwd=project_dir)
 
     for project in os.listdir(TEST_PROJECTS):
diff --git a/mkdocs/utils/__init__.py b/mkdocs/utils/__init__.py
index 4960762..37d2a31 100644
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -370,7 +370,7 @@ def get_themes():
 
         themes[theme.name] = theme
 
-    themes = dict((name, os.path.dirname(theme.load().__file__))
+    themes = dict((name, os.path.dirname(os.path.abspath(theme.load().__file__)))
                   for name, theme in themes.items())
 
     return themes
diff --git a/requirements/project.txt b/requirements/project.txt
index 0fed659..a07047c 100644
--- a/requirements/project.txt
+++ b/requirements/project.txt
@@ -2,7 +2,5 @@ click>=4.0
 Jinja2>=2.7.1
 livereload>=2.3.2
 Markdown>=2.5
-mkdocs-bootstrap>=0.1.1
-mkdocs-bootswatch>=0.1.0
 PyYAML>=3.10
 tornado>=4.1
diff --git a/setup.py b/setup.py
index e5a6980..a2b47b6 100755
--- a/setup.py
+++ b/setup.py
@@ -56,14 +56,12 @@ setup(
     author='Tom Christie',
     author_email='tom@tomchristie.com',  # SEE NOTE BELOW (*)
     packages=get_packages("mkdocs"),
-    package_data=get_package_data("mkdocs"),
+    include_package_data = True,
     install_requires=[
         'click>=4.0',
         'Jinja2>=2.7.1',
         'livereload>=2.3.2',
         'Markdown>=2.3.1,<2.5' if PY26 else 'Markdown>=2.3.1',
-        'mkdocs-bootstrap>=0.1.1',
-        'mkdocs-bootswatch>=0.1.0',
         'PyYAML>=3.10',
         'tornado>=4.1',
     ],
```

Note that I added a `-v` (verbose) flag to the integration tests. As a reminder, the DEBUG log includes a log of all the config settings and their values after validation. In the log I was able to see that the paths to the themes were not absolute, which clued me in to that problem. Once that was resolved, I got curious why it still couldn't find the files. But since the log told me the full path to the theme dir, it was easy for me to navigate to it and discover the problem. I would recommend that the verbose flag be used for the integration tests all the time. It makes debugging much easier and causes to harm.

waylan:
Also, I did not remove the now unused `get_package_data` function from `setup.py`. I just made the simplest changes to get the tests to pass. You may want to use a more selective method to define what package data gets included (see the docs [here][1])

[1]: https://pythonhosted.org/setuptools/setuptools.html#including-data-files

waylan:
And I just realized that you had the requirements:

```
mkdocs-bootstrap>=0.1.1
mkdocs-bootswatch>=0.1.0
```
because the integration tests loop through all themes and run with each one. Nice way to make sure changes to MkDocs doesn't break the other themes. Unfortunately, something is causing `mkdocs-bootswatch>=0.1.0` to fail for me locally. I have not investigated what that is. And its late and I'm tired. It'll have to wait till another day.

d0ugal:
Hmm, something has clearly gone wrong here. The goals were to move the themes out but still require them so we don't break users. They would then be removed fully in a few releases.

I had some conflicts and rebased, I'm starting to suspect I either messed up the rebase or conflict resolution. It at least fails for me locally now I've ran `tox -r`

However, given that there seem to be some issues with the theme packages themselves, I am going to push this to 0.15 and get 0.14 out the door today.

d0ugal:
@waylan Thanks for the diffs, that made it easy for me to fix. I didn't have to remove the theme dependencies tho' - they seem to work fine for me. I'd be interested to see what problem you had with them. They use some of the more modern setuptools stuff, like the change you suggested here.

It all works well for me locally, waiting to see what CI makes of it.

waylan:
Glad I could help. ALthough, I don't see any of the changes (except the verbose output). Did you forget to add a commit? Looking at Travis' output, the debug info shows the page paths are still not absolute and the failure is because the template is not found. Those relate to the missing changes to utils and setup.py respectively.

d0ugal:
The changes are all there, other than the removal of the themes.

The error is slightly different actually, this time it can't find the `sitemap.xml` file and the paths I have checked are absolute. It also always passes for me locally for some reason, even after removing the tox virtualenvs etc.

d0ugal:
This is the path used for the sitemap.xml file. https://travis-ci.org/mkdocs/mkdocs/jobs/66553887#L147

d0ugal:
Okay, I have split this into small steps to see if I can see where the failure comes in because I still can't reproduce this locally. I expect one unittest failure as I didn't keep them in sync all the way, but the integration test should always work, atleast until we hit the issue seen before.

d0ugal:
Oh, wow - I did not expect it to fail at https://github.com/d0ugal/mkdocs/commit/6b1c6db00579801f6c10254c48e67c802855a272. Perhaps master is broken?

d0ugal:
The really odd thing is that all the unittests should fail until you get to https://github.com/d0ugal/mkdocs/commit/07cbca14b05ac5b6c935705f851dddfe97225ad0 because that removes the references to the themes but they pass.

d0ugal:
and you  can see for example that https://github.com/d0ugal/mkdocs/commit/1cbb79de61524f8f53db00ad12f3587ebdb83b67 is still trying to build the Amelia theme despite it being removed in the previous commit.

So. Confused. :confused:.

d0ugal:
Yeah, so #643 confirmed that master is okay. For some reason, just removing one theme causes things to fail is strange ways. It looks to me like a Travis issue and for some reason the theme directory still exists? I'm really not sure.

d0ugal:
So, I think I have finally fixed this with https://github.com/d0ugal/mkdocs/commit/a57e8cafc5c70e3460e31d8eb69892d7387086ee. Sigh.

d0ugal:
Question raised by @tomchristie should the themes be named `mkdocs-$name-theme`? I just went with `mkdocs-$name` (like `mkdocs-bootswatch`). I had considered this at one point but decided that packages could have multiple themes and could contain extensions too (eventually).

Worth thinking about it more before I merge this.

trel:
or possibly, `mkdocs-theme-$name`, depends on how you want them sorted and what other items a package contains...

d0ugal:
I think for now I'll just stick with `mkdocs-$themename` - we can't really enforce this on users anyway, so we might as well keep it simple and the change isn't compelling enough for me to go and rename the packages I already created.

d0ugal:
I plan on merging this tomorrow - if anyone has feedback on the documentation I added that would be particularly good. We can always do that review after it is merged if it is easier.

d0ugal:
@waylan Thanks, sorry that was much rougher than I expected. This is what I get for writing docs at 1am after a long long day.

waylan:
Sorry, but I commented on the commit directly (2784a1e) rather than the PR as GitHub won't show all the files for such a large diff. Although, now I see that the relevant files are being shown. [Sigh].

d0ugal:
:heart: @appveyor's new speed.

d0ugal:
+697 −12,897. So happy.

tomchristie:
> +697 −12,897.

:sparkles: :relieved: :sparkles:

PR #528, opened 2015-05-15
Refactor Markdown Extension Options Config.

waylan:
Config validation now handles all extension processing:

* Builtin extensions are defined and within the default scheme.
* User extensions are defined only as a list in the config. Note, this is
  a backward incompatable change from the previous (short-lived) release.
* The users extensions are added to the list of builtins.
* Any duplicates are accounted for in validation.
* Extension options are supported by a child key/value pair on the ext name.
* All extension options are compiled into a format Markdown accepts
  within the validation process and are saved to the internal `mdx_configs`
  config setting.
* The `mdx_configs` setting is private and raises an error if set by the user.
* A whole suite of tests were added to test all aspects of ext validation.

All relevant build tests were updated to pass the config to
`build.convert_markdown` as the config now handles all extension data.

The relevant documentation was updated to reflect the changes. While I was
at it, I spellchecked the entire document and made a few additional formatting
changes.

This fixes #519 plus a lot more.

PR #533, opened 2015-05-18
Finish removing OrderedDict code

facelessuser:
OrderedDict support was recently removed in #528, but there were still
some references to OrderedDict that are causing failures.  This
finishes the removal of OrderedDicts.


d0ugal:
Thanks!

PR #537, opened 2015-05-19
Change to active status should recursively move up tree

JordanCheney:
Now that the page configuration can include multiple levels of headers and pages setting an item to active should recursively set all items in that tree to active.

### DIFF
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -200,6 +200,7 @@ class Header(object):
         for ancestor in self.ancestors:
             ancestor.set_active(active)
 
+
 def _path_to_page(path, title, url_context, use_directory_urls):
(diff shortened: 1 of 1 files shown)


d0ugal:
Good spot, that makes sense I think. I'll need to give it a test - ideally we'd have a unittests too

mkdocs/nav.py
```diff
@@ -195,6 +195,10 @@ def indent_print(self, depth=0):
             ret += item.indent_print(depth + 1)
         return ret
 
+    def set_active(self, active=True):
+        self.active = active
+        for ancestor in self.ancestors:
+            ancestor.set_active(active)
```
d0ugal: We need an extra newline here.

PR #539, opened 2015-05-21
Deprecated config

d0ugal:


waylan:
Why are you passing the key to the run_validation method? Seems unnecessary to me. The only value I perceive is for using in warning messages. But why not do the same as with errors [here]? The key is known there. No need to pass it around so much and muddy of the API.

[here]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/base.py#L49

d0ugal:
@waylan Fair point, fixed that. I think it is looking much better now.

waylan:
Yes, looking much better.

PR #542, opened 2015-05-21
Unicode files

d0ugal:


mkdocs/tests/integration.py
```diff
@@ -14,79 +14,39 @@
...
+        command = ['mkdocs', 'build', '--site-dir', out, '--theme', theme]
+        subprocess.check_call(command, cwd=project_dir)
+
+    for project in os.listdir(TEST_PROJECTS):
 
-        run_build(theme, output, config_file, quiet)
+        project_dir = os.path.join(TEST_PROJECTS, project)
+        out = os.path.join(output, project)
+        command = ['mkdocs', 'build', '--site-dir', out]
+        subprocess.check_call(command, cwd=project_dir)
```
d0ugal: This is where I ran into #543. Otherwise whatever config I specified, it build the MkDocs documentation as I was at the root of the repo.

PR #375, opened 2015-03-25
Add copyright footer to base.html

lbenet:
Add copyright notice and note that the documentation was built using MkDocs,  on the footer, following
the model of mkdocs theme

PR #549, opened 2015-05-23
Add (centered) copyright footer Bootstrap and Bootswatch themes

lbenet:
This also includes a note that the documentation was built with MkDocs

This is related with a suggestion in #375


d0ugal:
Thanks!

Issue #554, opened 2015-05-25
Update ReadTheDocs to use latest MkDocs


d0ugal:
This is really an upstream issue, but I want to track it here for downstream users.

d0ugal:
readthedocs.org is currently running `0.11.1` this means that they are [missing all of these lovely commits](https://github.com/mkdocs/mkdocs/compare/0.11.1...master) which is roughly half of the MkDocs git history (based on number of commits).

d0ugal:
After I roll a 0.13.0 release in the next few days I will be working on updating readthedocs to use it, currently there are problems due to the theme customisations they make.

ericholscher:
:+1: Happy to merge & deploy the new version when it's ready.

d0ugal:
@ericholscher Thanks! Can we define what blocks are needed from themes so they work on ReadTheDocs? See https://github.com/mkdocs/mkdocs/issues/355 and https://github.com/rtfd/readthedocs.org/issues/978#issuecomment-62574931. I want to add blocks and reduce the duplication of the theme that you guys have - this should then reduce the risk of us breaking something if you only inject into small defined areas. We can then document that and add support in other themes.

d0ugal:
See https://github.com/rtfd/readthedocs.org/pull/1305

d0ugal:
This just landed in ReadTheDocs! w00t!

deejay1:
Seems like this has regressed in the latest master...

d0ugal:
@deejay1 What has?

deejay1:
@d0ugal mkdocs documents have problems with links not having trailing slashed and resulting in having "somedirectoryindex.html" in them

d0ugal:
@deejay1 That sounds unrelated to this issue - can you open a new one with an example? It might be that this is a problem with RTD itself and it needs to be reported upstream (we have not changed anything lately)

deejay1:
@d0ugal Ah sorry, got a bit confused with the links back and forth on that issue and commented on the wrong project :/

PR #569, opened 2015-05-28
Ensure all errors and warnings are logged before exiting.

### LINKED ISSUE #536: Improve error/warning handling in Config
While working on moving the markdown extension handling to the config, I noticed a few oddities with how errors and warnings are handled within the config. And taking a closer look at the code I see even more. I could just do a PR, but wanted some feedback on where you want to take this.

1.  I noticed that you need to hardcode the config key into any warnings (see this [example][1]) . This is bad as the same ConfigOption class could be used for more than one key (for example, if the Type or URL types raised a warning). `base.Config._validate` does the right thing with errors and uses the current key (see [L51][2]). The same thing should happen for errors two lines above.

2. Why don't warnings actually use warnings? Why not follow the pattern used for errors and define a `ValidationWarning` which subclasses [`warnings.Warning`][3]. Then within the code call `raise ValidationWarning('some message...')` and in `base.Config._validate` add a catch for warnings in addition to the current catch for errors. Using the same API for both warnings and errors seems more sensible to me (and easier to remember).

3. Why are the errors being obscured? Only the message from the error is saved (see [L51][2]) and then later a new error is raised (see [L148][4]). Wouldn't it be better if the errors themselves were saved and re-raised? It would certainly help with debugging. Yes, I realize that if multiple errors were raised, only one could get re-raised. Not sure what the best solution is here for errors. Maybe log all the errors and then raise the first one? At least that way the user can work through them one at a time if the traceback is needed to figure out what went wrong.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/config_options.py#L224
[2]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/base.py#L51
[3]: https://docs.python.org/2/library/exceptions.html#exceptions.Warning
[4]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/base.py#L148

waylan:
Also ensure all debug messages are logged. And exit without
creating a new, unhelpful traceback by raising SystemExit.
Note that `sys.exit` raises SystemExit. However, by raising it
explicitly, we can also output a helpful explaination of why
the program exited. The program exists with an exit code of 1
to indicate is failed.

Fixes #536.

### DIFF
--- a/mkdocs/cli.py
+++ b/mkdocs/cli.py
@@ -10,6 +10,7 @@ from mkdocs import gh_deploy
 from mkdocs import new
 from mkdocs import serve
 from mkdocs import utils
+from mkdocs import exceptions
 from mkdocs.config import load_config
 
 log = logging.getLogger(__name__)
@@ -71,13 +72,17 @@ def serve_command(dev_addr, config_file, strict, theme, livereload):
 
     logging.getLogger('tornado').setLevel(logging.WARNING)
 
-    serve.serve(
-        config_file=config_file,
-        dev_addr=dev_addr,
-        strict=strict,
-        theme=theme,
-        livereload=livereload,
-    )
+    try:
+        serve.serve(
+            config_file=config_file,
+            dev_addr=dev_addr,
+            strict=strict,
+            theme=theme,
+            livereload=livereload,
+        )
+    except exceptions.ConfigurationError as e:
+        # Avoid ugly, unhelpful traceback
+        raise SystemExit('\n' + str(e))
 
 
 @cli.command(name="build")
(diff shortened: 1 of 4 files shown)


d0ugal:
Awesome, this looks great. FYI, I'm holding off on new features to make rolling a 0.13.2 easier if I need to do it for #554 but I consider this a bug fix, so I'll include it if I make another point release.

Just waiting for CI, then I'll happily merge. Thanks again!

d0ugal:
Looks like it needs a small test update to take into account the string -> exception change.

waylan:
I've fixed the failures and moved the "clean exit on error" handling to the CLI as it should only ever be needed there.

d0ugal:
Thanks!

PR #579, opened 2015-06-02
CLI improvements

waylan:
* Added short options (`-a`) to most options.
* Added a `--quiet` option which silences INFO and WARNINGS.
* Made `--quiet` and `--verbose` common options to all subcommands (yet only defined once). For backward compat, they also work on the `mkdocs` parent command - although perhaps they should be removed from there. If you agree, let me know and I'll add that change to the PR.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -15,6 +15,8 @@ You can determine your currently installed version using `mkdocs --version`:
 
 ## Version 0.14.0 (2015-??-??)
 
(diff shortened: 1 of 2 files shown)


d0ugal:
Looks good to me. There is a minor flake8 nit and can you add to the release notes. I created an entry for 0.14 in master.

waylan:
@d0ugal I've made the requested changes and this should be ready to go.

d0ugal:
Thanks!

Issue #583, opened 2015-06-02
Either don't use six or support an older version (like 1.4)


d0ugal:
This is causing quite a few users problems, particularly on OSX.

See #576, #590 and [a mailing list post](https://groups.google.com/forum/#!topic/mkdocs/6IXq0PWVt6I). There are possible workarounds for this in the issues and email thread.

samhatfield:
For what it's worth, I'm working on Linux Mint and I also had some issues. My system was using the six stored in `/usr/lib/python2.7/dist-packages`, which was about 1.5, even if a newer one was installed to `/usr/local/lib/python2.7/dist-packages` via pip. Eventually I had to rename the six files in the first directory to force it to use pip's version.

d0ugal:
@samhatfield Thanks, we are trying to resolve this ASAP in #585. I'll probably roll a new release shortly after as this is causing issues for lots of people. I had no idea six was treated specially by different OSes.

d0ugal:
This is now fixed in master, I'll be working on releasing very soon.

PR #585, opened 2015-06-03
Removed six dependency.

waylan:
See #583

Locally, on Python 2 I'm getting two test failures (both exhibit same issue) with relative urls that I don't understand. I can't see how the changes I made would effect this. The weird thing is the tests pass in Python 3. If anyone has any suggestions, I'm happy to fix it.

Otherwise, everything should be good to go.

As I suspected, the parts of six that MkDocs was using was very minimal. Mostly `string_types` and one instance of `text_type` -- both of which are easy enough to replicate in `utils.py`. The rest is just a few different import paths, which is easy to deal with. I think six is way too much overhead/bloat for the amount of use MkDocs gets from it.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -15,6 +15,7 @@ You can determine your currently installed version using `mkdocs --version`:
 
 ## Version 0.14.0 (2015-??-??)
 
+* Remove dependancy on the six library. (#583)
 * Add `--quiet` and `--verbose` options to all subcommands.
 * Add short options (`-a`) to most command line options.
 
--- a/mkdocs/build.py
+++ b/mkdocs/build.py
@@ -1,5 +1,6 @@
 # coding: utf-8
 
+from __future__ import unicode_literals
 from datetime import datetime
 import io
 import logging
@@ -8,7 +9,6 @@ import os
 from jinja2.exceptions import TemplateNotFound
 import jinja2
 import json
-import six
 
 from mkdocs import nav, search, utils
 from mkdocs.relative_path_ext import RelativePathExtension
@@ -109,7 +109,7 @@ def get_page_context(page, content, toc, meta, config):
         base = config['site_url']
         if not base.endswith('/'):
             base += '/'
-        canonical_url = six.moves.urllib.parse.urljoin(
+        canonical_url = utils.urljoin(
             base, page.abs_url.lstrip('/'))
     else:
         canonical_url = None
--- a/mkdocs/cli.py
+++ b/mkdocs/cli.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # coding: utf-8
 
+from __future__ import unicode_literals
 import logging
 import click
 
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -1,10 +1,10 @@
+from __future__ import unicode_literals
 import logging
 import os
-
-import six
 import yaml
 
 from mkdocs import exceptions
+from mkdocs import utils
 
 
 log = logging.getLogger('mkdocs.config')
@@ -14,7 +14,7 @@ class ValidationError(Exception):
     """Raised during the validation process of the config on errors."""
 
 
-class Config(six.moves.UserDict):
+class Config(utils.UserDict):
     """
     MkDocs Configuration dict
 
@@ -106,7 +106,7 @@ def _open_config_file(config_file):
     log.debug("Loading configuration file: %s", config_file)
 
     # If it is a string, we can assume it is a path and attempt to open it.
-    if isinstance(config_file, six.string_types):
+    if isinstance(config_file, utils.string_types):
         if os.path.exists(config_file):
             config_file = open(config_file, 'rb')
         else:
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -1,6 +1,5 @@
+from __future__ import unicode_literals
 import os
-
-import six
 
 from mkdocs import utils, legacy
 from mkdocs.config.base import Config, ValidationError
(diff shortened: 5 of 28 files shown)


facelessuser:
Something is definitely wrong, but I can't see what broke it.  If I get some time later, I will try pulling it down and seeing if I can figure it out.

facelessuser:
It appears to be related to `# from __future__ import unicode_literals`. I removed it from some key files, and the error went away.

Personally, I am all for `unicode_literals`, and I think they should be enforced throughout the project.  Mkdocs should consistently work in unicode everywhere.  I just think there must be some areas that aren't yet handling unicode properly.

Also it looks like you left a `six` in `build_tests.py`.  That should get cleared up with a common reference for `zip`.

d0ugal:
I'd hazard a guess that the issue is here: https://github.com/mkdocs/mkdocs/blob/0.13.3/mkdocs/utils.py#L184-L259

I somewhat confirmed that by removing the uncode_literals import from `utils.py` and `tests/build_tests.py` and then the tests pass. Some of that code is really hairy and I can't say that I fully understand it!

Given that it works on 3, could it be that the 2.X stdlib doesn't play well with unicode? We already hack around a Python 2.6 bug here.

facelessuser:
> I somewhat confirmed that by removing the uncode_literals import from utils.py and tests/build_tests.py and then the tests pass. Some of that code is really hairy and I can't say that I fully understand it!

Yeah those were the two places I disabled it as well.

> Given that it works on 3, could it be that the 2.X stdlib doesn't play well with unicode? We already hack around a Python 2.6 bug here.

Something isn't playing nice with unicode, but I haven't had enough time yet to analyze exactly what.   It is much easier starting a project from the beginning with proper unicode support than it is sometimes to retroactively enable it as you will catch the bug a lot earlier in the former scenario.  Many libraries these days force unicode internally to avoid py2X and py3X issues.  I have no doubt we can get mkdocs using unicode internally without issues, but we are just going to have to decode some its black magic first to understand why it is currently broken before we can fix it.  I think once I can pinpoint *exactly* why it fails, a fix will be straight forward.

d0ugal:
Yeah, I agree - unfortunately MkDocs had Python 3 and unicode both retroactively added and it has left a few areas needing love. We are slowly getting there. I'll try and find some time to look at this soon if somebody else doesn't beat me too it. I have a lot going on at the moment tho'

facelessuser:
No worries, I plan on digging more into the issue over the coming days anyways.  Last night I was content with just classifying the problem; didn't really feel up to digging into it deep last night.

facelessuser:
As I was feeling more clear this morning, I think I found the issue.  What wasn't obvious last night, is now painfully obvious:

```python
    if (nav.file_context.current_file.endswith("/index.md") is False and
            nav.url_context.base_path is not '/' and
            relative_url.startswith("./")):
        relative_url = ".%s" % relative_url
```

Should be


```python
    if (nav.file_context.current_file.endswith("/index.md") is False and
            nav.url_context.base_path != '/' and
            relative_url.startswith("./")):
        relative_url = ".%s" % relative_url
```

d0ugal:
Aha! Good spot. Took me two reads to see what changed :)

facelessuser:
I know, right? But when you see it, it is oh so clear :).

d0ugal:
Looks like this change will be good after that is updated. Given the six requirement has caused so many problems already, I want to release a fix. Should I roll 0.14 which would be a very small release? Do this as a 0.13.4 bug fix release?

I'm most tempted by a 0.14 release as removing requirements seems bigger than a 0.13.4 and it is also easiest probably.

waylan:
Good catch @facelessuser . Interestingly, I suspect that another way to solve the same issue would be to force pyyaml to return the config in all Unicode. But the `is not` => `!=` change should definitely still happen. The all Unicode config should happen also as a step toward all-unicode internally--but I see that as a separate issue from removing six so I didn't address it here.

Also, thanks for pointing out the `six` reference in `build_tests.py`. My not-so-good grep skills missed that one. I'm curious why that didn't show up as a failure in the tests as six is no longer listed as a dependency. Maybe Travis includes that as a standard lib? I know I forgot to remove it when testing locally (I should have forced tox to rebuild the envs from scratch).

I won't have any time to work on this for a few days. But if anyone else wants to take a crack at it, feel free.

facelessuser:
I'll take care of the Unicode YAML stuff in a separate commit later today.

d0ugal:
> Maybe Travis includes that as a standard lib?

That is an interesting question. Since Tox creates virtualenvs on Travis some black magic must be happening to make it get there. I believe virtualenv copies the system Python, so it is probably included in the distro.

waylan:
I just pushed what should be a fix for this (and updated the release notes). I'm not able to run any tests locally from where I am, so we'll have to wait for the CI builds to run to confirm it's fixed.

d0ugal:
Thanks @waylan I can probably do a minor update if needed.

waylan:
Hmm, the Travis docs don't list six as a [pre-installed package][1]. Not sure whats going on there. In any even, it looks like all tests are passing now.

[1]: http://docs.travis-ci.com/user/languages/python/#Pre-installed-packages

facelessuser:
I'll wait until this is merged before I provide the YAML unicode update.  For my unittest, I will need to check for a unicode type, and I'll have to rebase if I don't pick up this change.  Looks like it will drop in clean though.

d0ugal:
Thanks @waylan and @facelessuser!

d0ugal:
@facelessuser Do you think you'll get to the YAML change soon? I'm planing to release an early 0.14.0 but I can wait for the unicode change too.

facelessuser:
Yes, I will put it together today.

d0ugal:
Fantastic, thanks - no pressure tho' :)

facelessuser:
Pull has been made, it took me a bit longer as my windows system (even without my changes) was having problems accessing config files because they weren't being closed; so I had to debug that a bit.  But everything should be working well now. The pull is #592.

PR #592, opened 2015-06-04
YAML now loads all strings as unicode

facelessuser:
- All strings are redirected to be loaded as unicode
- I was experiencing an issue on my system where I was getting errors
trying to access config file objects because they were not being closed.
In all cases, once the YAML content was loaded, we were done accessing
the file content.  YAML object will close file access in the objects
once processing the content is complete.

### DIFF
--- a/mkdocs/utils.py
+++ b/mkdocs/utils.py
@@ -38,28 +38,26 @@ else:                           # pragma: no cover
 
 def yaml_load(source, loader=yaml.Loader):
     """
-    Custom yaml loader.
+    Wrap PyYaml's loader so we can extend it to suit our needs.
 
     Load all strings as unicode.
     http://stackoverflow.com/a/2967461/3609487
     """
 
-    class Loader(loader):
-
-        """Custom Loader."""
-
-        pass
-
     def construct_yaml_str(self, node):
         """Override the default string handling function to always return unicode objects."""
-
         return self.construct_scalar(node)
 
+    class Loader(loader):
+        """Define a custom loader derived from the global loader to leave the global loader unaltered."""
+
+    # Attach our unicode constructor to our custom loader ensuring all strings will be unicode on translation.
     Loader.add_constructor('tag:yaml.org,2002:str', construct_yaml_str)
 
     try:
         return yaml.load(source, Loader)
     finally:
+        # TODO: Remove this when external calls are properly cleaning up file objects.
         # Some mkdocs internal calls, sometimes in test lib, will load configs
         # with a file object but never close it.  On some systems, if a delete
         # action is performed on that file without Python closing that object,
--- a/mkdocs/utils.py
+++ b/mkdocs/utils.py
@@ -12,6 +12,7 @@ import os
 import sys
 import shutil
 import markdown
+import yaml
 
 from mkdocs import toc
 
@@ -33,6 +34,39 @@ if PY3:                         # pragma: no cover
 else:                           # pragma: no cover
     string_types = basestring,  # noqa
     text_type = unicode         # noqa
+
+
+def yaml_load(source, loader=yaml.Loader):
+    """
+    Custom yaml loader.
+
+    Load all strings as unicode.
+    http://stackoverflow.com/a/2967461/3609487
+    """
+
+    class Loader(loader):
+
+        """Custom Loader."""
+
+        pass
+
+    def construct_yaml_str(self, node):
+        """Override the default string handling function to always return unicode objects."""
+
+        return self.construct_scalar(node)
+
+    Loader.add_constructor('tag:yaml.org,2002:str', construct_yaml_str)
+
+    try:
+        return yaml.load(source, Loader)
+    finally:
+        # Some mkdocs internal calls, sometimes in test lib, will load configs
+        # with a file object but never close it.  On some systems, if a delete
+        # action is performed on that file without Python closing that object,
+        # there will be an access error. This will process the file and close it
+        # as there should be no more use for the file once we process the yaml content.
+        if hasattr(source, 'close'):
+            source.close()
 
 
 def reduce_list(data_set):
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -15,6 +15,7 @@ You can determine your currently installed version using `mkdocs --version`:
 
 ## Version 0.14.0 (2015-??-??)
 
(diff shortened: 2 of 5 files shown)


facelessuser:
Looks like everything checked out in CI.  I probably won't be able to address any issues until late tonight my time, but I think this is a pretty straight forward pull.

waylan:
As an interesting test of this, I would like to see if the tests still pass with the changes to `mkdocs/config/config_options.py` in 6bb45cb reverted. I think the tests added in that commit should probably remain--only  remove the `str` checks.

Also, this PR should probably have a note added to the release notes.

One other potential concern: what happens when a user passes a value in on the command line? Specifically, is it Unicode? I realize only a few of the config settings are even available from the command line, but it might be worth looking into. I have not checked this myself.

facelessuser:
> As an interesting test of this, I would like to see if the tests still pass with the changes to mkdocs/config/config_options.py in 6bb45cb reverted. I think the tests added in that commit should probably remain--only remove the str checks.

Yes removing `str` allows it to still pass the test suites with this pull.

> Also, this PR should probably have a note added to the release notes.

I can do that later today.

> One other potential concern: what happens when a user passes a value in on the command line? Specifically, is it Unicode? I realize only a few of the config settings are even available from the command line, but it might be worth looking into. I have not checked this myself.

I have not checked this myself either.  Though that is a general Unicode support thing.  This issue is specifically just enforcing an all Unicode yaml *load*, not to bring mkdocs up to a 100% Unicode safe state.  I think eventually an in-depth analysis of Unicode handling should be performed, but this is probably not the place for that. What happens after the load needs to be looked at separately, but it is definitely something important to look into, especially as we are now enforcing `unicode_literals` everywhere via #585.  One thing is for sure, if there are holes in Unicode support, they will eventually pop up in the issue tracker :).

mkdocs/utils.py
```diff
@@ -35,6 +36,39 @@
...
+    """
+    Custom yaml loader.
+
+    Load all strings as unicode.
+    http://stackoverflow.com/a/2967461/3609487
+    """
+
+    class Loader(loader):
+
+        """Custom Loader."""
```
d0ugal: This comment can be removed, it doesn't tell me anything.

mkdocs/utils.py
```diff
@@ -35,6 +36,39 @@
...
+
+        pass
```
d0ugal: I'd also remove `pass`

waylan: Actually I believe it would be an error to remove `pass` as nothing is defined within the class. Instead the class method `add_constructor` is called later (L58) which modifies the class. We can't just call `add_constructor` on the parent class (`yaml.Loader`) as that would modify the parent class (defined in the yaml lib) rather than a copy/ subclass. I agree, it is a little odd when reading the code, but the yaml lib's API requires it be that way. I suspect the comment was an attempt to clarify this (which it apparently did not do very well).

d0ugal: Ah, so it is. I thought it was a class method below. That's a confusing API.

mkdocs/utils.py
facelessuser: Sure, I'm just in a habit now of providing  comments for classes and methods, I kill it in the next rev.

mkdocs/utils.py
facelessuser: I don't want to monkey patch the global loader.  I don't think that is ever a good approach.  Instead, I inherit from the global loader and use my own leaving the global intact in its default state.  This I feel is a better approach.  I can, instead of removing the comment, expound on this more in the class comment.  I apologize as sometimes what is obvious in my head isn't obvious outside of it :).

d0ugal: Sounds good. Expand on what is going on in the comment and remove the pass. We don't need a docstring and `pass` :)

facelessuser: I think you will find PyLint angry if I remove the pass.  A matter of fact, I think you will find Python angry if I remove the pass and refuse to compile.

Just like you can't have an empty if block without a pass, you can't have an empty class without a pass.

I know it may seem odd, but this is the proper way.  The pass needs to stay.

facelessuser: I explain a little further.  If I remove the pass, Python will think everything under the class is supposed to be part of it and complain about unexpected indentation (or something similar).  Pass tells python "it's okay, there's nothing else here".

d0ugal: I'm really aware what pass is used for. It isn't required if a docstring is provided and pylint is absolteuly file with it. We already do this here: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/exceptions.py

facelessuser: Ahh, then I have learned something new today :).

facelessuser: Apologies if I came off sounding like you didn't know what a `pass` was for.  I was more just trying to wrap my head around the response from my understanding albeit incorrect understanding.  I've just always used the `pass` explicitly.  It makes sense though thinking about it, as docstrings are parsed as part of the function/class and not just as a stray comment.

d0ugal:
Other than the comment update, this looks great. Thanks very much! I'll merge it right after you update.

I'd like to remove the finally that closes files, but I'll do that in a follow up. It should be the responsibility of the tests to clean up.

Thanks!

mkdocs/utils.py
waylan: @facelessuser perhaps the code would be clearer if the `construct_yaml_str` function was defined before the `Loader` subclass (jsut reorder them in the source). It is easy to miss that it is not indented and not a class method (especially as its first argument is `self`). Reordering might help eliminate that confusion for future readers.

d0ugal: No problem, I was a bit snarky - so, sorry about that too! Reordering does seem like it would make it extra clear.

facelessuser: @waylan yes, I think that is an excellent idea; much more clear.

facelessuser:
> I'd like to remove the finally that closes files, but I'll do that in a follow up. It should be the responsibility of the tests to clean up.

Yeah, my intentions wasn't to have it there at all, but I realized I would have to massage some other things outside the scope of this pull to not have it not there, and I was trying to get this in today as I knew you were prepping for a bug fix, major release?

I will wrap this up tonight.

d0ugal:
Great! Thanks!

PR #600, opened 2015-06-05
Cleaned up socket.error stack trace for mkdocs serve.

samhatfield:
Only the error message is displayed for socket errors.


d0ugal:
Looks good to me! Test coverage is unfortunately very poor here. So I'll give a manual test and then merge. Thanks!

d0ugal:
Thanks!

Issue #248, opened 2014-11-27
Better document interlinking.


tomchristie:
Currently we only support direct links to other documents.
We ought to support a richer style of interlinking that makes it easy to link to particular sections.
This _could_ involve some kind of index, but would probably prefer if that was handled automatically.

The sort of thing I was originally think for this is:

```
See the [Parsers](ref:) documentation.
```

Creates a link to the section named 'Parsers'.

```
See the [Parsers documentation](ref:Parsers).
```

Creates a link to the section named 'Parsers'.

We could then just generate the index based on heading names, and raise errors if an interlink does not match.

d0ugal:
<ericholscher> I would at least recommend trying to implement it in a generic way, similar to the commonmark spec
<ericholscher> having yet another markdown flavor makes me full of sadness
<ericholscher> but pushing forward a spec of generic inline directives is really useful and moves the world to a better place
<ericholscher> http://talk.commonmark.org/t/generic-directives-plugins-syntax/444

waylan:
I have two things to say about this:

1. What's wrong with the existing solution (as @fedelibre pointed out)? `[link to section](file.md#section)` works just fine.

2. If some sort of "reference" feature/directive was added, then each section in the entire site would need to be unique. People already have difficulty keeping all sections in a single page unique. See point number 1.

robodude666:
@waylan I'm personally in favor of this approach:

 * It allows the author to move a section of their documentation to another file without having to worry about breaking any links.
 * It'll make it easy to link to a specific section w/o it being a ## header or using \<a\> html tags, for example linking to a specific paragraph.
 * Currently you need to figure out how a specific header would get sluggified, which can cause invalid hyperlinks if you guessed wrong.
 * Sometimes the slug generated by `toc` or `headerid` is not very author-friendly to type out every time they want to link to it: `how-the-parser-is-determined`. Kind of a PITA to link to `parsers.md#how-the-parser-is-determined`. I'd rather link to `ref:Parsers_determined`

It'll also make it easier once extensions :wink: are supported for developers to create API-documenting extensions that use refs instead of having to remember relative paths of API md files.

waylan:
@robodude666 if you want to define your own slugs, this is already possible with the [Attribute Lists][1] Extension. And I know that you are already aware that you can [define your own slugify function][2].

[...]

PR #602, opened 2015-06-06
remove more lorem ipsum

trel:
additional cruft...


d0ugal:
Thanks! This example might be useful for #248.

PR #605, opened 2015-06-06
Updated documentation on pages configuration.

samhatfield:
It now better reflects the new pages configuration format. The pages section
in `configuration.md` was also reduced to a summary, with most of the
details now in `writing-your-docs.md`. Also fixed a few typos.


d0ugal:
Thanks, that makes sense to me. The duplication before wasn't ideal. At some point we need to expand this to show that the pages can be nested further than two levels.

PR #606, opened 2015-06-07
Make all internal imports relative.

waylan:
Let's make MkDocs a better citizen. Relative imports were added in
Python 2.6 and there is no reason to not be using them. See [pep-0328][1].

Also updated the Release Notes.

[1]: https://www.python.org/dev/peps/pep-0328/

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -15,6 +15,7 @@ You can determine your currently installed version using `mkdocs --version`:
 
 ## Version 0.14.0 (2015-??-??)
 
+* Made all internal imports relative imports.
 * Improve Unicode handling by ensuring that all YAML strings are loaded as Unicode.
 * Remove dependancy on the six library. (#583)
 * Add `--quiet` and `--verbose` options to all subcommands.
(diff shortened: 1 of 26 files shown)


d0ugal:
I don't use them on purpose, I don't see how they make us a better citizen. I don't work on any projects that use relative importing everywhere. As far as I can tell, it only really makes sense to use them in complex/deep package namespaces which we thankfully don't have.

Specifying the full path is nice and explicit.

waylan:
Thiis obviously your choice, but since I've been using them in Markdown, I've gotten used to them. Is makes it really easy to identify which important statements are importing internal modules and which are importing third party code. Not confusing at all to me.

As far a being a good citizen, it means that the package is not required to be on the root of the PYTHONPATH. True, that is more important for a library than a tool. However, I can envision environments where a user could not import mkdocs in a traditional way but still want to use it. By using relative imports they could potentially include it as a submodule within their project and import it nested within there projects code. This is a must-have feature for many users of the Markdown lib and it has just become standard practice for me. I pause every time I see nonrelative imports which import internal modules, which motivated me to do the work. Buy, hey this is your project.

waylan:
By the way, if it is still a goal to build a self-contained install that does not need to have Python previously installed, I expect this would be a big help in getting there. While not nessecary, I expect it would make things easier. But maybe that previously stated goal has been abandoned. Just saying.

PR #612, opened 2015-06-07
Handle (pass-through) character and entity references in TOC parser

zmousm:
I have noticed the ampersand in a header like this:
```
### Tips & tricks
```
converted to `&amp;` by markdown, is dropped in the resulting TOC:
```
Tips  tricks
```

This patch fixes this by adding `handle_entityref` (and `handle_charref`, while we're at it) to `mkdocs.toc.TOCParser` which pass through such entities unchanged.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -24,6 +24,7 @@ You can determine your currently installed version using `mkdocs --version`:
 * Stack traces are no longer displayed on socket errors, just an error message.
 * Bugfix: Fix a JavaScript encoding problem when searching with spaces. (#586)
 * Bugfix: gh-deploy now works if the mkdocs.yml is not in the git repo root (#578)
+* Bugfix: Handle (pass-through instead of dropping) HTML entities while parsing TOC.
 
 ## Version 0.13.3 (2015-06-02)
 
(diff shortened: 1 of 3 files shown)


d0ugal:
Interesting, good catch! Would you be able to add a test and a quick mention in the release notes?

zmousm:
Testing for more entities would require the input to be unicode...

d0ugal:
This is great, thanks very much!

waylan:
> Testing for more entities would require the input to be unicode...

It should be Unicode already. The TOC is initially returned by Markdown and the Markdown library is all Unicode all the time. Additionally, due to a recent change, all literal strings defined in MkDocs are also Unicode as every module uses Unicode literals (`from __future__ import unicode_literals`), including the test modules. So, unless MkDocs converts the Unicode string to a byte string somewhere (which would be a bug), then it should be Unicode.

Actually, I see that the tests wrap the output with `str()`. They should probably be using `mkdocs.utils.text_type()` which will always return a Unicode string regardless of Python version.

zmousm:
I tried your suggestion, and you are right, however contrary to my belief markdown does not seem to convert e.g. `©` to an entity, but rather returns unicode in the toc, so a test like this fails:

```python
    def test_charref(self):
        md = dedent("""
        # Heading © 1
        """)
        expected = dedent("""
        Heading &copy; 1 - #heading-1
        """)
        toc = markdown_to_toc(md)
        self.assertEqual(text_type(toc).strip(), expected)
```

facelessuser:
I don't think Python Markdown has ever auto converted things like that. As I recall, you need to use the correct HTML entity.  Or you can write a Python Markdown extension to convert those for you.

PR #619, opened 2015-06-09
Allow README.md to be used as the index page

### LINKED ISSUE #608: Allow non-index.md files to be used as the homepage
ReadTheDocs allows README.md files, GitHub also renders those as the root. We should support that and/or a way for users to pick.

SEJeff:
If index.md exists, it will be
preferenced over README.md, to match the featureset of [readthedocs](https://github.com/rtfd/readthedocs.org/blob/347104f25c82bba0c42029a2226cbbb71a0bf627/readthedocs/doc_builder/base.py#L89-L103) as the index page.

Fixes #608

I'm not sure what you require as far as documentation goes, but first wanted to make sure this, or something like it is reasonable. I used the maxsplit argument to `str.split()`, so there shouldn't be any unexpected surprises.

### DIFF
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -141,12 +141,17 @@ def get_html_path(path):
     Map a source file path to an output html path.
 
     Paths like 'index.md' will be converted to 'index.html'
+    Paths like 'README.md' will be converted to 'index.html' if index.md doesn't also exist
     Paths like 'about.md' will be converted to 'about/index.html'
     Paths like 'api-guide/core.md' will be converted to 'api-guide/core/index.html'
     """
     path = os.path.splitext(path)[0]
     if os.path.basename(path) == 'index':
         return path + '.html'
+    # To match github-esque repositories where the webui renders README.md
+    # as the page index in the repository browser.
+    elif os.path.basename(path).split('.', 1)[0] == 'README':
+        return path.replace('README', 'index') + '.html'
     return "/".join((path, 'index.html'))
 
 
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -13,6 +13,7 @@ class UtilsTests(unittest.TestCase):
     def test_html_path(self):
         expected_results = {
             'index.md': 'index.html',
+            'README.md': 'index.html',
             'api-guide.md': 'api-guide/index.html',
             'api-guide/index.md': 'api-guide/index.html',
             'api-guide/testing.md': 'api-guide/testing/index.html',


mkdocs/utils/__init__.py
```diff
@@ -141,12 +141,17 @@ def get_html_path(path):
...
     Paths like 'about.md' will be converted to 'about/index.html'
     Paths like 'api-guide/core.md' will be converted to 'api-guide/core/index.html'
     """
     path = os.path.splitext(path)[0]
     if os.path.basename(path) == 'index':
         return path + '.html'
+    # To match github-esque repositories where the webui renders README.md
+    # as the page index in the repository browser.
+    elif os.path.basename(path).split('.', 1)[0] == 'README':
+        return path.replace('README', 'index') + '.html'
```
SEJeff: The `path.replace` could likely be better, but again, I figured it was easier to get a yay or nay on the approach and go from there before this gets merged.

d0ugal:
This looks reasonable, I am not 100% sure what will happen if you have both an index and a README file tho' - presumably one will write over the other and the page will be lost?

mkdocs/utils/__init__.py
d0ugal: That seems reasonable to me.

d0ugal:
Okay, so I have tested this with the MkDocs documentation.

- If I just rename the `index.md` it works pretty great
- We probably need to [update this](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L171-L172) too. I forget exactly where it is used, but it will show Home in the nav rather than README.
- If I duplicate the `index.md` to `README.md` we run into some issues.
    - We have two navigation items that point to the same location.
    - The homepage shows the `README.md` and the `index.md` isn't accessible.

I think having both is going to be a relatively uncommon use-case, but probably still a valid one.

FWIW, this is what I see. Regardless of if I click on Home or README the README nav is highlighted and the README page is shown (I modified the title so I knew).

screenshot

d0ugal:
I am also trying to figure out what documentation changes are required. It seems we never explicitly state that an index.md is required. So I would suggest just adding a brief [in the docs here](https://github.com/mkdocs/mkdocs/blob/master/docs/user-guide/configuration.md#pages) to say something like...

> MkDocs will use the index.md file for your home page, if that isn't found it will also try README.md

d0ugal:
I am trying to think how we can better approach this issue. The problem is that [where we set the output path](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/nav.py#L144), we don't have any context of what other pages exist.

SEJeff:
I'm going to close this for now. This clearly will need a lot more work.

d0ugal:
@SEJeff Thanks for your effort, it might be something like this would be easier with @tomchristie's factoring. I don't think it is specifically a goal of his, but he might have some ideas.

tomchristie:
Actually the 'refactor' branch does already handle this. I'm not too concerned about clashing 'index.md' and 'README.md' files, simply warning if this occurs would be sufficient, and also be easy to do.

d0ugal:
Awesome. My main motivation for this is that it is a common GitHub pattern, ReadTheDocs hacks support in for people as it is.

PR #620, opened 2015-06-09
Update mkdocs utils coverage

SEJeff:
I feel kind of bad making a module of such pretty table (dict) based tests look so ugly, but this does up the coverage to 100% for `mkdocs.utils`. While I was in there futzing with the README stuff related to #608, I decided to take a stab at this.

I'm relatively new to mock and normally use it as a context manager instead of a decorator like most of the rest of the tests. If you've got any better ways or strong preferences on how to do this, I'm totally happy to change things.

### DIFF
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -101,6 +101,7 @@ class UtilsTests(unittest.TestCase):
         assertPathGenerated("img.png", "./img.png")
         assertPathGenerated("./img.png", "./img.png")
         assertPathGenerated("/img.png", "../img.png")
+        assertPathGenerated("https://www.example.com/img.png", "https://www.example.com/img.png")
 
     def test_reduce_list(self):
         self.assertEqual(
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -1,9 +1,8 @@
 #!/usr/bin/env python
 # coding: utf-8
 
-from __future__ import unicode_literals
-
 import mock
+from __future__ import unicode_literals, with_statement
 import os
 import unittest
 
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -4,11 +4,17 @@
 import mock
 from __future__ import unicode_literals, with_statement
 import os
+import sys
 import unittest
 
 from mkdocs import nav, utils, exceptions
 from mkdocs.tests.base import dedent
-from mock import patch
+from mock import mock_open, patch
+
+if sys.version_info.major == 2:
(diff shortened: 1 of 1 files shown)


SEJeff:
Gah, when I'm running:

``bash
(ve)[jeff@omniscience mkdocs] (update-mkdocs-utils-coverage)$ nosetests -vs mkdocs/tests/utils/ --with-coverage --cover-package mkdocs mkdocs 2>&1 | grep 'mkdocs\.utils '
mkdocs.utils                     149      0   100%   
``

It shows 100% code coverage, but indeed, there doesn't appear to be coverage for 112 and 310. Additionally, I didn't test it on py2.6 knowing travis would. (I'm developing on Fedora 21 / python 2.7.x).

I'll try to finish the rest of this in the next day or two.

mkdocs/tests/utils/utils_tests.py
```diff
@@ -1,12 +1,19 @@
...
+from __future__ import unicode_literals, with_statement
 import os
+import sys
 import unittest
 
 from mkdocs import nav, utils
 from mkdocs.tests.base import dedent
+from mock import mock_open, patch
+
+if sys.version_info.major == 2:
```
facelessuser: Why don't you just check `utils.PY3`?

d0ugal: +1, that will most likely fix the failure in 2.6

SEJeff:
So I rebased this on master, which is horribly broken and fails a lot of tests. When it fails, it will be due to master being currently broken. All I did was changed the tests to use `utils.PY3`, and then rebased against master to remove merge conflicts.

d0ugal:
> So I rebased this on master, which is horribly broken and fails a lot of tests. When it fails, it will be due to master being currently broken. All I did was changed the tests to use utils.PY3, and then rebased against master to remove merge conflicts.

Master is fine for me, Travis and Appveyor...

mkdocs/tests/utils/utils_tests.py
```diff
@@ -1,14 +1,20 @@
 #!/usr/bin/env python
 # coding: utf-8
 
-from __future__ import unicode_literals
-
 import mock
+from __future__ import unicode_literals, with_statement
```
d0ugal: Moving the __future__ import here causes a syntax error.

d0ugal:
Once I fix the error I commented on, all the tests pass. If you are seeing other errors I'd be curious to see them

mkdocs/tests/utils/utils_tests.py
```diff
@@ -1,14 +1,20 @@
...
 import os
+import sys
```
d0ugal: Adding sys causes a lint error because it isn't used.

d0ugal:
Closing this for now, I would be happy to consider an updated version.

PR #627, opened 2015-06-12
Loading all Markdown in memory

d0ugal:
This change updates the Markdown loading to load all the documents at startup. This gives us much more flexibility when we want to reference other Markdown files or use the metadata from a Markdown file earlier in the build.

Remaining tasks:

- [ ] Fix a few test  failures on older Python versions
- [ ] Fix the code in a few places that is wonky. Commented on inline.
- [ ] Add to release notes

Original text below the line.

---

This needs a bunch of works and it breaks loads of the tests, but it works quite well. It gets the page title from the Markdown if it isn't provided in the mkdocs.yml.

Feedback very welcome, it is a bit rough at the moment tho'

See #316

### DIFF
--- a/mkdocs/build.py
+++ b/mkdocs/build.py
@@ -10,7 +10,9 @@ from jinja2.exceptions import TemplateNotFound
 import jinja2
 import json
 
-from mkdocs import nav, search, utils
+from mkdocs import nav
+from mkdocs import search
+from mkdocs import utils
 from mkdocs.relative_path_ext import RelativePathExtension
 import mkdocs
 
@@ -164,30 +166,16 @@ def build_template(template_name, env, config, site_navigation=None):
 
 def _build_page(page, config, site_navigation, env, dump_json):
 
-    # Read the input file
-    input_path = os.path.join(config['docs_dir'], page.input_path)
-
-    try:
-        input_content = io.open(input_path, 'r', encoding='utf-8').read()
-    except IOError:
-        log.error('file not found: %s', input_path)
-        raise
-
-    # Process the markdown text
-    html_content, table_of_contents, meta = convert_markdown(
-        markdown_source=input_content,
-        config=config,
-        site_navigation=site_navigation
-    )
+    html_content, toc = convert_markdown(page.markdown, config, site_navigation)
 
     context = get_global_context(site_navigation, config)
     context.update(get_page_context(
-        page, html_content, table_of_contents, meta, config
+        page, html_content, toc, page.meta, config
     ))
 
     # Allow 'template:' override in md source files.
-    if 'template' in meta:
-        template = env.get_template(meta['template'][0])
+    if 'template' in page.meta:
+        template = env.get_template(page.meta['template'][0])
     else:
         template = env.get_template('base.html')
 
@@ -208,7 +196,7 @@ def _build_page(page, config, site_navigation, env, dump_json):
     else:
         utils.write_file(output_content.encode('utf-8'), output_path)
 
-    return html_content, table_of_contents, meta
+    return html_content, toc, page.meta
 
 
 def build_extra_templates(extra_templates, config, site_navigation=None):
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -42,7 +42,9 @@ def paths_to_pages(pages_config, use_directory_urls, docs_dir):
             pages_config[i] = _path_to_page(entry, use_directory_urls, docs_dir)
             continue
         elif not isinstance(entry, dict):
-            raise Exception("Broken 1")
+            msg = ("Line in 'page' config is of type {0}, dict or string "
+                   "expected. Config: {1}").format(type(entry), entry)
+            raise exceptions.ConfigurationError(msg)
 
         next_cat_or_title, subpages_or_path = next(iter(entry.items()))
 
@@ -184,7 +186,26 @@ class Page(object):
         self.next_page = None
         self.ancestors = []
 
-        self.load_markdown()
+        self.markdown, self.meta = self.load_markdown()
+
+    def __eq__(self, other):
+
+        def sub_dict(d):
+            return dict((key, value) for key, value in d.items()
+                        if key in ['title', 'input_path', 'abs_url'])
+
+        return (isinstance(other, self.__class__)
+                and sub_dict(self.__dict__) == sub_dict(other.__dict__))
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __str__(self):
+        return self.indent_print()
+
+    def __repr__(self):
+        return "nav.Page(title='{0}', input_path='{1}', url='{2}')".format(
+            self.title, self.input_path, self.abs_url)
 
     def load_markdown(self):
 
@@ -196,7 +217,7 @@ class Page(object):
             log.error('file not found: %s', input_path)
             raise
 
-        self.markdown, self.meta = meta.get_data(input_content)
+        return meta.get_data(input_content)
 
     @property
     def title(self):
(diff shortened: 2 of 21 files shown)


waylan:
Looks like a good start.

One thing that concerns me is that the Markdown is rendered to HTML too soon. If/when a plugin API is added, I anticipate a hook being provided prior to each page being built which means the HTML should not be rendered until build time. I realize that it is rendered early because that is how you are getting the meta-data for each page. And yes, that info should already be available before the pre-build hook.

Of course, those two things are in direct opposition to each other. However, I recommend that the Python-Markdown's Meta-Data Extension not be used. Instead extract the meta-data early (perhaps using [docdata][1]), and save the meta-data and the markdown source (with the meta-date removed) as two separate attributes in the Page instance. The way, the Markdown source can be used later to render the HTML just before building the page, but after the pre-build hook.

If you don't want to add another dependency, I give permission to include the relevant parts of docdata directly in MkdDocs. Or you are welcome to use some other implementation of your choice or creation.

An interesting side-affect will be that the code which calls Markdown will still be in `build.py` so you won't need to move the Markdown wrapper (leave it in `build.py`) and you won't need to update those tests (which are many of the currently failing tests).

[1]: https://github.com/waylan/docdata

facelessuser:
> However, I recommend that the Python-Markdown's Meta-Data Extension not be used.

+1 to this.  Personally, I have never found the meta extension to be all that useful as I want the meta data **before** I parse.  Usually, by time I've parsed is too late for me to use the meta for what I want.

d0ugal:
Great, I'm glad you both said this because I was struggling with it too. It did indeed feel a bit strange to render the Markdown so early.

> If you don't want to add another dependency, I give permission to include the relevant parts of docdata directly in MkdDocs. Or you are welcome to use some other implementation of your choice or creation.

@waylan Great - thanks! I am a bit hesitant to include another dependency, just because of the effort recently we had to do to remove two (six and ghp-import). So I might take this approach, initially I'll try a prototype with the dependency and see how that goes.

d0ugal:
Updated. I think this generally does make much more sense. Unfortunately, most of the tests still break as they are broken due to how I changed the internal pages structure. This PR changes it so that we convert from string paths to page objects as soon as possible. The previous was was quite delayed and the tests took advantage of this to easily define a set of pages as a list of strings.

d0ugal:
... and because we load the pages when we do now, it verifies they all exist. A bunch of tests have fake paths to markdown files because they never tested the build process. /me *sighs*

mkdocs/build.py
```diff
@@ -164,30 +166,16 @@ def build_template(template_name, env, config, site_navigation=None):
...
-        markdown_source=input_content,
-        config=config,
-        site_navigation=site_navigation
-    )
+    html_content, toc = convert_markdown(page.markdown, config, site_navigation)
 
     context = get_global_context(site_navigation, config)
     context.update(get_page_context(
-        page, html_content, table_of_contents, meta, config
+        page, html_content, toc, page.meta, config
```
waylan: Within the template I would expect the html_content, toc, and meta to all be attributes of the page object. I realize that may involve so template refactoring and this is prototyping right now, but I think that is we're this needs to go.

d0ugal: Yeah, I agree. In my previous version I had added it to the template, but left it available via the old names too to avoid updating all the templates at the same time.

mkdocs/nav.py
```diff
@@ -26,6 +29,37 @@ def filename_to_title(filename):
...
+        next_cat_or_title, subpages_or_path = next(iter(entry.items()))
+
+        if isinstance(subpages_or_path, list):
+            paths_to_pages(subpages_or_path, use_directory_urls, docs_dir)
+        elif isinstance(subpages_or_path, utils.text_type):
+            pages_config[i] = _path_to_page(subpages_or_path,
+                                            use_directory_urls, docs_dir,
+                                            title=next_cat_or_title)
+        else:
+            raise Exception('Broken 2')
```
d0ugal: This needs fixed.

mkdocs/nav.py
```diff
@@ -148,6 +186,63 @@ def __init__(self, title, url, path, url_context):
         self.next_page = None
         self.ancestors = []
 
+        self.markdown, self.meta = self.load_markdown()
+
+    def __eq__(self, other):
```
d0ugal: It feels a bit like a hack adding this, but it made the tests much easier to do. Need to decide if we want to keep it or not.

mkdocs/serve.py
```diff
@@ -71,7 +71,7 @@ def builder():
             theme=theme,
         )
         config['site_dir'] = tempdir
-        build(config, live_server=True)
+        build(config, live_server=True, clean_site_dir=True)
```
d0ugal: This is an incidental change and should be moved out.

mkdocs/tests/config/config_tests.py
```diff
@@ -151,32 +149,33 @@ def test_default_pages_nested(self):
...
-            open(os.path.join(tmp_dir, 'subB', 'index.md'), 'w').close()
+            open(os.path.join(tmp_dir, 'subB', 'page2.md'), 'w').close()
             conf = config.Config(schema=config.DEFAULT_SCHEMA)
             conf.load_dict({
                 'site_name': 'Example',
                 'docs_dir': tmp_dir
             })
             conf.validate()
+
             self.assertEqual([
```
d0ugal: This is so gross. Is there a better way to do this?

mkdocs/utils/mdutils.py
```diff
@@ -0,0 +1,4 @@
+
+
+def get_markdown_title(markdown):
+    return
```
d0ugal: This needs to be implemented :smile:

mkdocs/nav.py
```diff
@@ -26,6 +29,37 @@ def filename_to_title(filename):
...
+
+def paths_to_pages(pages_config, use_directory_urls, docs_dir):
+
+    for i, entry in enumerate(pages_config):
+
+        if isinstance(entry, utils.text_type):
+            pages_config[i] = _path_to_page(entry, use_directory_urls, docs_dir)
+            continue
+        elif not isinstance(entry, dict):
+            msg = ("Line in 'page' config is of type {0}, dict or string "
```
d0ugal: This is the error message we had before, it is terrible still. Should improve.

mkdocs/nav.py
```diff
@@ -35,6 +69,9 @@ def __init__(self, pages_config, use_directory_urls=True):
         self.homepage = self.pages[0] if self.pages else None
         self.use_directory_urls = use_directory_urls
 
+        for page in self.pages:
+            page.url_context = self.url_context
```
d0ugal: Because the page object is created at a different time, we need to add this here. It seems quite hacky tho - maybe add a method to Page to set the context?

tkirill:
@d0ugal This is that required refactoring for #316?  If so then what is the status of this PR?  Is it waiting for review or there are some tasks to be done?  If there are any free tasks to be done to merge this PR I would be happy to do them.

waylan:
@tkirill, yes this is the refactoring for #316 among other things. I believe this is waiting on the issues mentioned in the comments above.

d0ugal:
@tkirill Yup, this re-factoring would enable #316 (it actually [almost includes the implementation](https://github.com/mkdocs/mkdocs/pull/627#discussion-diff-33427199)) and a number of other things we want to do. @waylan is correct that the above comments and check list in the description [here](https://github.com/mkdocs/mkdocs/pull/627#issue-87650999) are the outstanding tasks.

If you want to try any of them, I'd suggest making pull requests against my fork and then they can be merged in and included here. Stick to one per commit to make things easier to review if you can.

d0ugal:
Closing in favour of @tomchristie's work. https://github.com/mkdocs/mkdocs/tree/restructure

PR #628, opened 2015-06-12
Absolute md links

d0ugal:


mkdocs/relative_path_ext.py
```diff
@@ -68,6 +71,10 @@ def path_to_url(url, nav, strict):
         # If the site navigation has been provided, then validate
         # the internal hyperlink, making sure the target actually exists.
         target_file = nav.file_context.make_absolute(path)
+
+        if target_file.startswith(os.path.sep):
+            target_file = target_file[1:]
+
         if target_file not in nav.source_files:
```
d0ugal: This breaks on windows because `source_files` are local filesystem paths. I'm not sure about the best approach to fix this.

PR #629, opened 2015-06-13
Remove six imports and test against oldest allows requirement versions

d0ugal:


d0ugal:
@waylan You might be curious to know, it seems that python-livereload pulls in six (but only in the latest version) so that is probably why we had it before and couldn't figure out why. So probably not something weird with Travis.

PR #644, opened 2015-06-19
Generate search index entries for all headings, not just H1 and H2

zmousm:
Currently search index generation is limited to H1 and H2, which is hard-coded. I have found it necessary to extend this to H3 and H4, as I wanted to provide more specific search results. Not sure what a good balance would be (e.g. stop at H4), as the level of headings included in the TOC, for example, is up to the template, so I opted for no limits.

### DIFF
--- a/mkdocs/search.py
+++ b/mkdocs/search.py
@@ -11,7 +11,7 @@ except ImportError:                     # pragma: no cover
 
 class SearchIndex(object):
     """
-    Search index is a collection of pages and sections (H1 and H2
+    Search index is a collection of pages and sections (heading
     tags and their following content are sections).
     """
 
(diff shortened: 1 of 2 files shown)


d0ugal:
This looks great! I'm really busy at the moment, planning to look at it at the start of next week.

mkdocs/search.py
```diff
@@ -26,9 +26,9 @@ def _find_toc_by_id(self, toc, id_):
         for toc_item in toc:
             if toc_item.url[1:] == id_:
                 return toc_item
-            for toc_sub_item in toc_item.children:
-                if toc_sub_item.url[1:] == id_:
-                    return toc_sub_item
+            toc_item_r = self._find_toc_by_id(toc_item.children, id_)
+            if toc_item_r is not None:
+                return toc_item_r
```
d0ugal: This if statement could be simplified with `return toc_item_r`

d0ugal:
Thanks!

thumbs

mkdocs/search.py
zmousm: That wouldn't work because recursion breaks as soon as it reaches the first empty list of children. For example:

```
() recursing [u'#_1', u'#_2'] for dnsmasq
(_1) recursing [u'#capwap-split-mac', u'#local-switching-vs-overlay-tunnels', u'#vlans-vlan-profiles-overlayvlanwla-tunnels'] for dnsmasq
(capwap-split-mac) recursing [] for dnsmasq
(_1) recursion returned <type 'NoneType'>
() recursion returned <type 'NoneType'>
```

So `_find_toc_by_id` returns `None`.

d0ugal: Oh, sorry. Good point. I totally miss-read it.

PR #651, opened 2015-06-23
Update index.md

RussellTaylor83:
Updated documentation, mkdocs help is no longer valid, suggested mkdocs --version instead.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -51,10 +51,10 @@ Install the `mkdocs` package using pip:
 
     $ pip install mkdocs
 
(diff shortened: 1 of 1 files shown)


d0ugal:
Good spot! I think we should go for `mkdocs --help`. It is a bit more useful than `--version` as it helps users discover the other commands.

RussellTaylor83:
I had thought that, but mkdocs --help isn't listed as a command, and it does exactly the same as just mkdocs. Either way is fine, it proves it's run. --version was consisten with testing Python and Pip though :)

d0ugal:
Yeah, okay - that seems reasonable.

d0ugal:
Thanks! 
thumbs

PR #654, opened 2015-06-25
Remove a "hack" which fixes "![Image alt text](images/image.jpg)" to …

ivanz:
…generate "./images/image.jpg" instead of "../../images/image.jpg". 

However it breaks a test?


d0ugal:
So, I can see that it causes you an issue, but it also fixes another. What we really need is a better way to handle both cases.

The test that fails is the one that proved the hack originally.

d0ugal:
Closing this for now, would be happy to consider an update.

PR #660, opened 2015-06-26
Ensure consistent order of auto-populated pages.

### LINKED ISSUE #638: Navigation not in alphanumeric order (when pages config is automatic)
##### mkdocs Version:

0.14.0

##### Summary:

I have a problem with the table of content of my docs. I want that my chapters are in the correct order.

##### Steps To Reproduce:

In the docs folder there is the following structure:

- docs/
    - book/
        - Chapter 1/
            - something.md
        - Chapter 2/
            - something.md

...

Now when I build the docs the chapters won't be in the correct order.

The problem exists regardless of the theme.

##### Expected Results:

- book
    - Chapter 1
    - Chapter 2
    - Chapter 3
    - Chapter 4
    - Chapter 5
    - Chapter 6
    - Chapter 7
    - Chapter 8
    - Chapter 9
    - Chapter 10
    - Chapter 11
    - Chapter 12

##### Actual Results:

- book
    - Chapter 8
    - Chapter 6
    - Chapter 5
    - Chapter 2
    - Chapter 12
    - Chapter 9
    - Chapter 4
    - Chapter 7
    - Chapter 3
    - Chapter 1
    - Chapter 11
    - Chapter 10

waylan:
Both files and directories should be sorted to ensure consistent ordering of
pages in the menus across builds and systems. Turns out to be a simple fix.
As `os.walk` provides a reference to the list of dirs it uses internally,
we can modify that list in place and it will use the modified list.

While this may not give proper alphanumeric ordering for all languages,
it goes ensure consistent ordering. For more control over ordering,
it is expected that a plugin could be used once that API becomes available.

This fixes #638.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -19,6 +19,7 @@ You can determine your currently installed version using `mkdocs --version`:
 * Fix issues when using absolute links to Markdown files. (#628)
 * Add support for [site_description] and [site_author] to the [ReadTheDocs]
   theme. (#631)
+* Bugfix: Ensure consistent ordering of auto-populated pages. (#638)
 
 [site_description]: /user-guide/configuration.md#site_description
 [site_author]: /user-guide/configuration.md#site_author
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -328,7 +328,8 @@ class Extras(OptionallyRequired):
         if self.file_match is None:
             raise StopIteration
 
-        for (dirpath, _, filenames) in os.walk(docs_dir):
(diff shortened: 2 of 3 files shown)


d0ugal:
Awesome. This looks great. I wonder if this would work:

    sorted(os.walk(docs_dir)):

I think it does, but yours is clearer. So we can stick with that. Can you add a Bugfix to the release notes? I know I missed a few recently (#662).

waylan:
Do'h! Just pushed a new commit with the release notes updated.

d0ugal:
Thanks! :cake:

PR #673, opened 2015-06-28
Perform a clean build when serving

d0ugal:


mkdocs/serve.py
```diff
@@ -71,7 +71,7 @@ def builder():
             theme=theme,
         )
         config['site_dir'] = tempdir
-        build(config, live_server=True)
+        build(config, live_server=True, clean_site_dir=True)
```
tomchristie: What does this actually change if we're building to a temporary directory?

d0ugal: 1. Start `mkdocs serve`
2. view a markdown file in the browser
3. delete/rename that markdown file, refresh.

It wont be removed until `mkdocs serve` is restarted. We might be building in a temp directory, but we build in it multiple times.

waylan: Would it perhaps make more sense to remove/delete the temp dir when the server is shut down? Then the tmp dir wouldn't be hanging around littering the file system when not in use and the next time the `serve` command is used it will always be "clean."

d0ugal: We remove the tempdir when the server shutdown already. https://github.com/mkdocs/mkdocs/blob/master/mkdocs/serve.py#L88

This is to do with multiple builds happening within one "session" of running `mkdocs serve`. We could create a new tempdir on each build and change where we are serving from... but that sounds like hard work.

waylan: Ah, okay, that makes sense to me.

PR #674, opened 2015-06-29
Remove `build_sitemap()` function.

tomchristie:
Minor clean up that looks like it doesn't functionally affect anything.
Any reason we can't do this?

### DIFF
--- a/mkdocs/build.py
+++ b/mkdocs/build.py
@@ -131,17 +131,6 @@ def get_page_context(page, content, toc, meta, config):
     }
 
 
-def build_sitemap(config, env, site_navigation):
-
-    log.debug("Building sitemap.xml")
-
-    template = env.get_template('sitemap.xml')
-    context = get_global_context(site_navigation, config)
-    output_content = template.render(context)
-    output_path = os.path.join(config['site_dir'], 'sitemap.xml')
-    utils.write_file(output_content.encode('utf-8'), output_path)
-
-
 def build_template(template_name, env, config, site_navigation=None):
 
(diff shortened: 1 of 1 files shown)


mkdocs/build.py
```diff
@@ -247,7 +236,8 @@ def build_pages(config, dump_json=False):
         log.debug("Search is enabled but the theme doesn't contain a "
                   "search.html file. Assuming the theme implements search "
                   "within a modal.")
-    build_sitemap(config, env, site_navigation)
+
+    build_template('index.xml', env, config, site_navigation)
```
d0ugal: This would need to be 'sitemap.xml'

d0ugal:
Other than the comment I just added, no, I can't think why it wouldn't work. I am not sure why I didn't use this originally.

mkdocs/build.py
```diff
@@ -131,17 +131,6 @@ def get_page_context(page, content, toc, meta, config):
     }
 
 
-def build_sitemap(config, env, site_navigation):
-
-    log.debug("Building sitemap.xml")
```
d0ugal: We lose the log message, but that isn't a big deal.

tomchristie: We have "Building template: sitemap.xml" instead.

tomchristie:
Inline comment addressed. Brain-fail.

mkdocs/build.py
d0ugal: Oh, yeah - cool. That works for me then :)

d0ugal:
Thanks!

PR #678, opened 2015-07-02
Move all commands into submodule

tomchristie:
For discussion...

Move all the commands into their own modules to keep the complexity more tightly scoped.

Related future stuff to consider (for context, rather than as part of this PR)

* `gh_deploy` -> `deploy`
* `cli.py` -> `main.py` and runnable from `python main.py` for local testing purposes.


tomchristie:
Alternately, use full directories instead so that we can keep *all* the related modules inside the relevant command...

    /commands/build/...
    /commands/deploy/...
    etc.

d0ugal:
Seems reasonable to me. I don't like the idea of packages for each command - that can happen after, but some really don't need it.

waylan:
> `gh_deploy` -> `deploy`

There is already a ticket for that: #431, although it is not named very well.

> `cli.py` -> `main.py` and runnable from `python main.py` for local testing purposes.

I'm assuming you mean "`cli.py` -> `__main__.py` and runnable from `python -m mkdocs` for local testing purposes."

Changing it to `main.py` gives you no advantage over `cli.py`. You can run `python cli.py` just as easily as you can run `python main.py`. You only get special treatment when you use `__main__.py`.

tomchristie:
"Changing it to main.py gives you no advantage over cli.py"

The suggested name change is simply because one of those is meaningful to me and the other is not. :)
The other implied suggestion is that we'd add a `if __name__ == "__main__":` block to allow it to be run as `python mkdocs/main.py`  Anyways I should prob keep those suggestions out of this seperate thread, just felt they added a little wider context.

d0ugal:
Changing to `__main__.py` makes sense. I'll never use that entry-point tho'

PR #681, opened 2015-07-03
Remove `include_nav` and `include_next_prev`

### LINKED ISSUE #679: Drop `include_nav`, `include_next_prev` from config and context.
We should consider dropping the `include_nav` and `include_next_prev` config.
The only theme that currently references them is the builtin `mkdocs` theme, and this could presumbly instead use...

    {% if nav|length > 1 %}

...and...

    {% if next_page or prev_page %}

For contextual switching, instead of using the current context variables.

tomchristie:
Closes #679.

I still need to eyeball this to verify that the template `nav|length>1` filtering really does behave as expected.

### DIFF
--- a/docs/user-guide/custom-themes.md
+++ b/docs/user-guide/custom-themes.md
@@ -87,8 +87,6 @@ site_url          | [site_url]            |
 extra_css         | [extra_css]           |
 extra_javascript  | [extra_javascript]    |
 extra             | [extra]               |
-include_nav       | [include_nav]         |
-include_next_prev | [include_next_prev]   |
 copyright         | [copyright]           |
 google_analytics  | [google_analytics]    |
 
@@ -102,8 +100,6 @@ google_analytics  | [google_analytics]    |
 [extra_css]: ./configuration.md#extra_css
 [extra_javascript]: ./configuration.md#extra_javascript
 [extra]: ./configuration.md#extra
-[include_nav]: ./configuration.md#include_nav
-[include_next_prev]: ./configuration.md#include_next_prev
 [copyright]: ./configuration.md#copyright
 [google_analytics]: ./configuration.md#google_analytics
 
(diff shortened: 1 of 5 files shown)


tomchristie:
Works as expected...

single page:

image

multiple pages:

image

tomchristie:
I leave the decision on this to you @d0ugal. Personally I think we should just go ahead with it and document in release notes that those two bits of config/context have been removed (anyone with a custom theme who's updating *might* need to act on that.)

tomchristie:
Would close these:

* https://github.com/mkdocs/mkdocs-bootstrap/issues/5
* https://github.com/mkdocs/mkdocs-bootswatch/issues/4

d0ugal:
Seems reasonable. Away until Sunday, so won't take a proper look until then.

There is a way in config_options to deprecate options - which means we can show people a specific error rather than one like "this doesn't exist" - it is undocumented and unused because of $reasons. I can show you an example next week.

The purpose of it was for moving non-behavioural config options under extra. I pulled the move from the last release but left the code to do it still.

It might need some tweaking for config options that are being just removed vs moved.

d0ugal:
@tomchristie See #684. I searched a bit on GitHub and see a few people using it so I think it is worth showing something a bit more descriptive.

It's a bit confusing as your dev branch is on the mkdocs repo, the PR is for that.

tomchristie:
> It's a bit confusing as your dev branch is on the mkdocs repo, the PR is for that.

I guess it'd be worth me starting to use `tomchristie/mkdocs` instead, right?

d0ugal:
> I guess it'd be worth me starting to use tomchristie/mkdocs instead, right?

I don't have any strong opinions, so it is up to you. I just prefer to keep dev branches in my fork. I am not even sure GitHub will let you as https://github.com/tomchristie/mkdocs redirects to `mkdocs/mkdocs`

d0ugal:
Two further things.

- `mkdocs.config.config_options.NumPages` (and related tests) can be removed with this change.
- Needs an addition to the release notes.

d0ugal:
I plan to come back to this and update it for the next release, but anyone else interested should feel free to have a go first.

waylan:
I've updated this and went with graceful deprecation in #1026. I'm closing this in favor of #1026.

PR #684, opened 2015-07-07
Provide a deprecation warning for include_nav and include_next_prev

d0ugal:


d0ugal:
@tomchristie Does this seem reasonable? Should we merge it? It just goes into your branch.

tomchristie:
@d0ugal I'm deferring this for now. Will get the internal refactoring sorted, then we can make a clearer judgment on what, if any, deprecations we make towards a 1.0 release.

Issue #686, opened 2015-07-08
ManPage


williamjmorenor:
Hi

I am working in package mkdocs in Fedora, I am getting a warning because mkdocs do not ship a manpage,:

rpmlint rpmbuild/RPMS/noarch/mkdocs-0.14.0-1.fc22.noarch.rpm 
mkdocs.noarch: W: no-manual-page-for-binary mkdocs

Can please add a mange to mkdocs tarball

d0ugal:
Hi, sure - we can do this. I just need to first learn how  :)

waylan:
There are a few tools out there that will convert optparse/argparse help to man page. I know Click is built on top of optparse (sort of), but not sure if they would work.

williamjmorenor:
https://admin.fedoraproject.org/pkgdb/package/mkdocs/

waylan:
I took another look at this and found [click-man], which is a tool that creates man pages for Click applications. You install the tool and run one command and all your man pages are generated (one per command/subcommand). Interestingly, the [docs] for the package explain why we shouldn't automate that on build (rather than documenting how to automate it). The short of it is that Python builds are system agnostic while man pages are system specific. Also, if installed in a virtualenv, the manpage should not be installed (a reason they miss).

So if you are building deb/rmp/etc packages, then you can use click-man to generate the man page, It's just one more line in your build script. Or, if you installed MkDocs via pip and want a man page, you can `pip install click-man` and then run that one command manually.

Therefore, my proposed resolution of this issue is to document how to use click-man in the installation documentation and be done with it.

[click-man]: https://github.com/click-contrib/click-man
[docs]: https://github.com/click-contrib/click-man#automatic-man-page-installation-with-setuptools-and-pip

d0ugal:
👍🏼 great outcome.

PR #687, opened 2015-07-09
manpage.1  for #686

williamjmorenor:


williamjmorenor:
captura de pantalla de 2015-07-09 17-39-27

mkdocs.1
```diff
@@ -0,0 +1,27 @@
...
+\fBbuild\fP
+make a HTML website from sources
+.TP
+\fBserve\fP
+start a server to view in real time the documentation
+.TP
+\fBgh\-deploy\fP
+commit changes to a GitHub Page
+.SH "AUTHOR"
+William Moreno Reyes
```
d0ugal: Is the author that of MkDocs or the manpage?

mkdocs.1
```diff
@@ -0,0 +1,27 @@
+.TH "MKDOCS" "1" "0.1" "William Moreno Reyes" ""
+.SH "NAME"
+mkdocs \- Create HTML documentation from markdown sources
+.SH "SYNOPSIS"
+mkdocs [help|new|build|serve|gh\-deploy] {options}
```
d0ugal: `mkdocs help` doesn't work. This is incorrect on mkdocs.org I see. This would be correct

    mkdocs [build|gh\-deploy|new|serve] {options}

mkdocs.1
```diff
@@ -0,0 +1,27 @@
...
+.SH "DESCRIPTION"
+MkDocs is a fast and simple way to create a website from source files written in Markdown, and configured with a YAML configuration file, the documentation  can be hosted anywhere, even in free hosting services like Read the Docs and GitHub Pages.
```
d0ugal: Can we improve the wrapping of this? I see in your screenshot it doesn't look great. Maybe we need to manually wrap it?

mkdocs.1
```diff
@@ -0,0 +1,27 @@
...
+.SH "OPTIONS"
+.TP
+\fBhelp\fP
+display this help and exit
```
d0ugal: I'd remove this one, if you want to show users how to get help - `mkdocs --help` is correct.

mkdocs.1
```diff
@@ -0,0 +1,27 @@
...
+.TP
+\fBserve\fP
+start a server to view in real time the documentation
+.TP
+\fBgh\-deploy\fP
+commit changes to a GitHub Page
+.SH "AUTHOR"
+William Moreno Reyes
+.SH "COPYRIGHT"
+CC\-BY
```
d0ugal: Assuming the Copyright is for MkDocs, and not this file, it should say Tom Christie as per the LICENSE file.

d0ugal:
The test failures can be ignored, they were resolved by 4cb864a66f20334c0573f434d62d03113d4065cb

mkdocs.1
williamjmorenor: This is the author of the manpage, you can edit everthing you want and add you, or rewritte all the manpage and only let your name here.

mkdocs.1
williamjmorenor: This is the licence of the manpage

mkdocs.1
tomchristie: http://man7.org/linux/man-pages/man7/man-pages.7.html

> A list of authors of the documentation or program.

mkdocs.1
tomchristie: A license for the manpage itself makes no sense to me - wouldn't be comfortable with this pull request as it stands.

robodude666:
I think it would be better to have the manpage file automatically built as part of setup.py using the click help info and setup.py project info; that way everything stays in sync.

d0ugal:
@robertogerson It would be better to do that, any ideas how?

waylan:
> I think it would be better to have the manpage file automatically built as part of setup.py using the click help info and setup.py project info; that way everything stays in sync.

I agree as well. As I commented [here](https://github.com/mkdocs/mkdocs/issues/686#issuecomment-119949161):

> There are a few tools out there that will convert optparse/argparse help to man page. I know Click is built on top of optparse (sort of), but not sure if they would work.

I know nothing about any of the libraries out there (I have no personal experience with any of them) and didn't provide any links as I can't recommend anything (and enough time has gone by since doing the research that I don't recall what they were). A quick search will turn them up easy enough. However, in my searches, I could find nothing which specifically stated it worked with Click.

robodude666:
I found a pretty good article [here](https://andialbrecht.wordpress.com/2009/03/17/creating-a-man-page-with-distutils-and-optparse/) on how to add a `build_manpage` command to `setup.py`. Probably what @waylan was referring to.

Instead of using `optparse` you can extract the commands from `click`:

```python
from mkdocs.cli import cli as mkdocs_cli
for name, command in mkdocs_cli.commands.items():
    print "%s - %s" % (name, command.short_help)
```

waylan:
I just took another look as this. Here are my notes for future reference.

> I found a pretty good article [here](https://andialbrecht.wordpress.com/2009/03/17/creating-a-man-page-with-distutils-and-optparse/) on how to add a `build_manpage` command to `setup.py`.

The basic idea of the linked article is to create a custom `HelpFormatter` class which outputs the help in Manpage format rather than a format appropriate for  the `--help` flag. While the article discusses optparse, which is now deprecated in favor of argparse (since Python 2.7), both libraries share the `HelpFormatter` API. Interestingly, Click also has a [`HelpFormatter`][1] class which has essentially the same API. It shouldn't be too much work to adapt the example script in that article to work with Click.

The rest of the article is about how to call that new formatter from a Distutils sub-command. That should carry over to a Setuptools subcommand with little change. There are a few benefits to this approach: The manpage is only generated on an appropriate OS (skip this step on Windows, etc), and some of the distutils/setuptools' meta data is available to include in the manpage.

[1]: https://github.com/mitsuhiko/click/blob/master/click/formatting.py

waylan:
Hmm, perhaps this is not as easy as I thought. Click's `HelpFormatter` is [hard-coded][1] into the `Context` class and use of the `Context` class is [hard-coded][2] into the `BaseCommand` class. Which means to use a different `HelpFormatter` we would need to subclass both classes as well as any of the existing subclasses of `BaseCommand`. And then there is the issue that we want the current default behavior for the `--help` flag.

It is true that the above linked article avoids these problems by a monkey patch:

```python
self._parser.formatter = ManPageFormatter()
```

However, Click doesn't offer an API that can be money-patched like this. There is no instance of the Formatter stored on a class anywhere. A method simply returns it and that method is buried multiple levels deep. I suppose that method could be replaced with a money-patch, but that feels like pretty rocky ground at this point.

It might make more sense to simply iterate over the various parts of the Click API and build up the manpage without trying to tap into its own renderer. It might even make sense to use a template. In fact, the current patch provides a base for the template. Just swap out the pieces of text (which are available from Click) with the appropriate template variables and pass in the Click instance as the template's context.

[1]: https://github.com/mitsuhiko/click/blob/master/click/core.py#L373
[2]: https://github.com/mitsuhiko/click/blob/master/click/core.py#L592

waylan:
This may be relevant to installation: <http://stackoverflow.com/a/15469948/866026>

The man page should not be installed (globally) when Mkdocs is installed in a virtual env.

d0ugal:
Closing this, because although I think it is a great idea we need to figure out a way to automatically generate it. Keeping this updated with the code will otherwise eventually cause issues.

Issue #653, opened 2015-06-25
Relative paths in sub-directory - shouldn't they get converted to absolute on build?


ivanz:
Given the following directory and file structure:

```
│   index.md
│
└───Section
    │   section.md
    │
    ├───files
    │       test.zip
    │
    └───images
            image.jpg
```

I expect that in `section.md` when I use: 

`![Image alt text](images/image.jpg)`

It will generate: `/Section/images/image.jpg`

However instead it generates: `../../images/image.jpg`

I would expect to see `../../images/image.jpg` generated only if I were doing `![ (/images/image.jpg)`

d0ugal:
Yeah, that is annoying. I've been meaning to look into this.

waylan:
I suspect this is a duplicate of #192. 

Also see [my comment][1] for an explanation of why all internal paths **must** be relative.

[1]: https://github.com/mkdocs/mkdocs/issues/192#issuecomment-75322787

ivanz:
Well my issue is that I want me and my team to be able to use an editor such as MarkdownPad for editing documentation and for this we want to use relative paths in markdown, so that we get to get preview of the images/diagrams in the editor (otherwise it wouldn't know what the root "/" is if absolute paths are used)

The changes in pull request #654 fix this for me but break one test. :(

d0ugal:
Closing in favour of #192.

Issue #683, opened 2015-07-03
Specifying 1.0 behavior.


tomchristie:
The following is a sketch of my thinking on an ideal configuration for a 1.0 release.
This doesn't take into account any compatibility concerns, I just wanted to start by outlining my thoughts on how we could get something really restricted, simple and consistent.

---
## Build process

Simplify the build process and specify it properly so that we can describe it in the documentation, and so that process is obvious to end users.
- Copy media files from theme dir.
- Copy media files from docs dir.
- Build static pages from theme dir. (HTML, JSON, XML files. Templated, passed global context.)
- Build static pages from docs dir. (HTML, JSON, XML files. Templated, passed global context.)
- Build documentation pages.

Template directories are `<docs_dir>/templates` and `<theme_dir>/templates`. Anything in those directories not considered as static build pages. Base template for documentation pages is `base.html` unless overridden with `template: ...` in the page metadata.

**Notes**:
- Move `sitemap.xml` into themes, for simpler, more consistent build process.
- Move `search_index.json` and search assets into themes for simpler more consistent build process. (search info in provided in context, use something like `{{ as_json(search) }}` in the template.
- Need to strictly specify how markdown paths are mapped to html paths.

---
## Template context

Properly separate the global and per-page context. Only build the global context once. Scope the context more nicely.
#### Global context

`site_title`
`homepage_url`

`media`:
&nbsp;&nbsp;&nbsp;&nbsp;`extra_javascript`
&nbsp;&nbsp;&nbsp;&nbsp;`extra_css`

`build`:
&nbsp;&nbsp;&nbsp;&nbsp;`utc_time`
&nbsp;&nbsp;&nbsp;&nbsp;`mkdocs_version`

`extra`:
&nbsp;&nbsp;&nbsp;&nbsp;`<variables from 'extra' config>`

`search`
`nav`
#### Per page context

Per-page context should be scoped on the `page` context variable:

`page`

Eg:

Use: `{{ page.title }}`, `{{ page.meta }}`, `{{ page.toc }}`, `{{ page.previous_url }}`
Not : `{{ page_title }}` and `{{ meta }}`, `{{ toc }}`, `{{ previous_url }}`.

**Notes:**
- Include a `{{ url(pathname) }}` function instead of using `{{ base_url }}/pathname`.
- Do we provide something like an `{{ is_active(nav_element) }}` in order to keep `nav` itself entirely static?
- We should be able to avoid the "walking" of the nav that we currently do.

---
[...]

Issue #192, opened 2014-11-06
Allow absolute URLs


jodeleeuw:
One thing I'm finding frustrating in crafting my documentation is the lack of support for absolute URLs. Say I've got a directory structure like this:

docs
|-- tutorial/
|---- hello-world.md
|-- api /
|---- foo.md

If I want to link to foo.md from hello-world.md then my link needs to be `[foo function](../api/foo.md)` but it would be easier if I could just do `[foo function](/api/foo.md)` and not worry about where I am in the directory tree as I'm writing the docs.

waylan:
Personally, I prefer relative URLs only. For example, when documenting a Python project, I may want to upload a set of static HTML files to http://pythonhosted.org (as provided by PyPI). I generate the static HTML, zip it up and upload it. However, the document root on pythonhosted.org is `http://pythonhosted.org/packagename/` (as opposed to readthedocs which uses a subdomain: `http://packagename.readthedocs.org`). That means that when serving the files locally (`http://localhost/`) the root is different than on pythonhosted.org. And of course, if I'm browsing the local static files (using `file://`) the root is different there and could be different for the same project on different systems (depending on where my project is locally stored on each system). Additionally, as @tomchristie demonstrated in a previous comment, the root is also different on GitHub (as well as other source code hosting sites) The only way that I have found to get consistent results is to use relative links. Yes, it can be a little annoying to always know exactly where you are in relation to the entire document structure, but it works everywhere.

A few possible workarounds:

1. Expect all URLs that start with `/` to be internal absolute URLs from the doc root and convert them to relative URLs saving the author the hassle.  The author can still use relative URLs (which do not get altered) for total control if desired. Of course, external links would not start with `/` but `http` or similar and also not be altered.

2. Same as (1) except that a preface is prepended to absolute URLs from the configuration settings keeping the URLs absolute. But that preface could be different in different contexts, which means the setting would need to change for development (eg: `http://localhost/`) and production (eg: `http://pythonhosted.org/packagename/`).

3. Add some special syntax to Markdown (through a Markdown extension) which is used for internal links only. As it knows about internal links it always does the "right thing" (which may be absolute or relative URLs). See the [wikilinks][1] extension for an example (it even accepts a callable to custom format the URLs which could be used here).

[...]

Issue #167, opened 2014-10-21
Only warn the user about files in the output directory that wont be overwritten


d0ugal:
As per @tomchristie's [comment](https://github.com/tomchristie/mkdocs/pull/154#issuecomment-59035198), we can expand on the additions in #154 by only warning the user when files wouldn't be overwritten (i.e. stale files that no longer exist in the docs).

waylan:
Now that "clean" is the default (see #997), we issue a warning by default when the user uses `--dirty`. Therefore, I'm closing this as it offers no additional value.

PR #689, opened 2015-07-10
Internal refactoring.

tomchristie:
Builds site in memory.

Presents a clean internal API, that is more comprehensible, allows us to make improvements more easily, and paves path to fully documented build behaviors and context available to themes.

In broader context this is a big step towards #683, but at this point I just want to present a fully compatible internal refactoring that should allow us to sort out a bunch of issues.

TODO:

- [x] Set `page.url` and `nav.is_active` states on each page.
- [ ] Build `404.html`, search and other template pages.
- [ ] Support multiple theme directories, as per current build style.
- [ ] Support `use_directory_urls`.
- [ ] Support `extra_css`, `extra_javascript`, `extra_templates`.
- [ ] Review #77, particularly search URLs, and anywhere `abs_url` is used. 
- [ ] Warn on duplicate output files in build.
- [ ] print statements -> warn/error.
- [ ] Tests, tests, tests.
- [ ] Cleanup unused bits in `utils`.
- [ ] Verify identical site build before/after this change.
- [ ] Document internal structure.
- [ ] Build but warn on .md files not in `pages`.
- [ ] Ensure markdown extensions in config are supported.
- [ ] OrderedDict for yaml loading of config.

Resolves: 
#653 - Relative paths to media sometimes broken.
#316 - Determine default titles from page table of contents.
#192 - Allow absolute URL links in source pages.
#608 - Allow README.md or others as homepage.

Makes these viable:
#658 - Validate links to anchors.
#167 - Only warn on properly stale files in site dir.
#699 - Build all markdown files.
No ticket - Support HTML files in nav.
No ticket - Support alternate layout eg `/path/to/file.html` vs `path/to/file/index.html`.


mkdocs/compat.py
```diff
@@ -0,0 +1,21 @@
+import sys
```
d0ugal: I'm really not a fan of re-creating this file. We got rid of it originally because it became a unmaintainable mess. The few compatability bits we do need are in:

https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L32

To give further context, I moved us to six which was much nicer from a code point of view, but it led us to having tons of install problems for users - it seems some distros special case six and you can't rely on having the version you specify in setup.py. I never quite understood why, but it caused a number of issues and emails on varying platforms.

Then when we removed six, @waylan opted for managing the different import paths within each file. I honestly am not a huge fan of that, but it is less tangled up than the compat.py was originally.

I'd personally be tempted to vendor six, if we did really want a file like this, then we would at least have something sorta-standard and maintained/well tested but that wasn't particularly popular when I suggested it.

mkdocs/structure/files.py
```diff
@@ -0,0 +1,128 @@
...
+            if self.root in ('index', 'README'):
+                directory = os.path.dirname(self.input_path)
+            else:
+                directory = os.path.splitext(self.input_path)[0]
+            return os.path.join(directory, 'index.html')
+        return self.input_path
+
+    def is_documentation_page(self):
+        return self.extension in (
+            '.markdown',
```
d0ugal: This is duplicated in the urls.py, would it be better to expand/change the is_markdown utility method and use that?

mkdocs/structure/pages.py
```diff
@@ -0,0 +1,143 @@
...
+        return title
+
+
+def build_pages(files):
+    pages = []
+    for file in files.documentation_pages():
+        md = markdown.Markdown(
+            extensions=[
+                _RelativePathExtension(file, files, strict=False),
+                'meta', 'toc', 'tables', 'fenced_code'
```
d0ugal: Looks like we have lost user extensions here.

mkdocs/compat.py
d0ugal: ... I don't have a strong preference which way we go, but we just need to be consistent. So if we really go back to compat.by there are a bunch of imports that need to be updated.

d0ugal:
Looks like a good start, but obviosuly still tons to do. So I just gave it a quick once over. My primary concern with wide sweeping refactors like this is what we might break. I tend to do these things in much smaller steps.

mkdocs/structure/pages.py
waylan: Also, the default extensions are defined in the config. That way, duplicates are handled properly, etc (a number of people have reported setting the permalink config setting on the toc extension (including the MkdDocs docs; see [here][1]), which would define it a second time. The current solution handles that properly).  The config would need to be passed to this build_pages function. Or at least both the `markdown_extensions` key and the (private) `mdx_configs` key (which holds extension configs).

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs.yml#L23

tomchristie: Yup - this part is incomplete. I believe it's on the todo list.

tomchristie:
> I tend to do these things in much smaller steps.

Likewise. Clearly breaking the rules here.

tomchristie:
> I tend to do these things in much smaller steps.

Likewise. Clearly breaking all the rules here.

d0ugal:
> > I tend to do these things in much smaller steps.
> 
> Likewise. Clearly breaking all the rules here.

Maybe we can treat this branch as an incubator and slowly move changes into master? Trying to think of a decent workflow. Next week I hope to have time for MkDocs as I'll be at EuroPython and plan on sprinting. However, it will be tricky to do much without conflicting badly with this work and I don't want to cause you a massive headache. I'd be happy to help drive this forward too, but not sure I can easily do that yet.

Open to ideas.

d0ugal:
How can I help push this work forward?

tomchristie:
@d0ugal Possibilities:

* There's some items on the checklist above that'd be obviously actionable. Eg print statements -> logging.
* There's some items on the checklist that I could take you through if you wanted to take them on.

**Alternatively**

* Taking a more radical tack on what we're expecting from 1.0  (eg. not necessarily 100% compatible against custom theme templates) would allow us to treat this work as a cleaner slate, and not get too bogged down in figuring out the gradual migration approach.

d0ugal:
> * Taking a more radical tack on what we're expecting from 1.0  (eg. not necessarily 100% compatible against custom theme templates) would allow us to treat this work as a cleaner slate, and not get too bogged down in figuring out the gradual migration approach.

If we call the next release 1.0, then I am happy to break things so long as we provide users with clear docs to upgrade. I wouldn't be happy about a `0.15.0` release breaking all the things.

I can try and look at some of the check list items - I guess I'll send a pull request against this branch if I get something together.

tomchristie:
> If we call the next release 1.0, then I am happy to break things so long as we provide users with clear docs to upgrade. I wouldn't be happy about a 0.15.0 release breaking all the things.

My preference would be to work on a 1.0 release that provides an easy upgrade path for users of the existing themes.

d0ugal:
I think this work is really valuable, and needs to happen at some point - but to tidy up a bit I'll close this as it has fallen out of date. Happy for a new PR to be opened.

PR #698, opened 2015-07-26
Corrected lint errors in Markdown docs.

waylan:
Also added README.md and CONTRIBUTING.md to the linter.

Note, that I am still getting one failer (in two locations). However
I consider that failer a bug in the linter and have reported it
upstream. We could disable that Rule (MD031), but as we are not
requiring the lint rules to pass presently, I just left it alone.

Also, while the code linter is set to allow lines 119 chars long,
I am using the Markdown linter's default of 80. Prose is easier to
read with shorter line lenghts, so I think it makes more sense to
use the default. Also, changing the default would have required
adding a config file. Adding a Ruby file for only one minor setting
seems silly, so I left it alone.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -11,14 +11,6 @@ generator that's geared towards building project documentation. Documentation
 source files are written in Markdown, and configured with a single YAML
 configuration file.
 
-!!! warning
-
-    MkDocs is currently still in development.
-
-    We're progressing quickly, but the documentation still needs filling in, and
-    there are a few rough edges. The 1.0 release is planned to arrive in the
-    next few months.
-
 ### Host anywhere
 
 Builds completely static HTML sites that you can host on GitHub pages, Amazon
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,37 +6,47 @@ Project documentation with&nbsp;Markdown.
 
 ## Overview
 
-MkDocs is a **fast**, **simple** and **downright gorgeous** static site generator that's geared towards building project documentation. Documentation source files are written in Markdown, and configured with a single YAML configuration file.
+MkDocs is a **fast**, **simple** and **downright gorgeous** static site
+generator that's geared towards building project documentation. Documentation
+source files are written in Markdown, and configured with a single YAML
+configuration file.
 
----
+!!! warning
 
-**MkDocs is currently still in development.**
+    MkDocs is currently still in development.
 
-We're progressing quickly, but the documentation still needs filling in, and there are a few rough edges.  The 1.0 release is planned to arrive in the next few months.
+    We're progressing quickly, but the documentation still needs filling in, and
+    there are a few rough edges. The 1.0 release is planned to arrive in the
+    next few months.
 
----
+### Host anywhere
 
-#### Host anywhere.
+Builds completely static HTML sites that you can host on GitHub pages, Amazon
+S3, or anywhere else you choose.
 
-Builds completely static HTML sites that you can host on GitHub pages, Amazon S3, or anywhere else you choose.
+### Great themes available
 
-#### Great themes available.
+There's a stack of good looking themes included by default. Choose from
+bootstrap, readthedocs, or any of the 12 bootswatch themes.
 
-There's a stack of good looking themes included by default. Choose from bootstrap, readthedocs, or any of the 12 bootswatch themes.
+### Preview your site as you work
 
-#### Preview your site as you work.
+The built-in devserver allows you to preview your documentation as you're
+writing it. It will even auto-reload whenever you save any changes, so all you
+need to do to see your latest edits is refresh your browser.
 
-The built-in devserver allows you to preview your documentation as you're writing it. It will even auto-reload whenever you save any changes, so all you need to do to see your latest edits is refresh your browser.
+### Easy to customize
 
-#### Easy to customize.
-
-Get your project documentation looking just the way you want it by customizing the theme.
+Get your project documentation looking just the way you want it by customizing
+the theme.
 
 ---
 
 ## Installation
 
-In order to install MkDocs you'll need [Python] installed on your system, as well as the Python package manager, [pip].  You can check if you have these already installed like so:
+In order to install MkDocs you'll need [Python] installed on your system, as
+well as the Python package manager, [pip]. You can check if you have these
+already installed like so:
 
 ```bash
 $ python --version
@@ -52,10 +62,11 @@ On Windows we recommend that you install Python and pip with [Chocolatey].
 Install the `mkdocs` package using pip:
 
 ```bash
-$ pip install mkdocs
+pip install mkdocs
 ```
 
-You should now have the `mkdocs` command installed on your system.  Run `mkdocs --version` to check that everything worked okay.
+You should now have the `mkdocs` command installed on your system. Run `mkdocs
+--version` to check that everything worked okay.
 
 ```bash
 $ mkdocs --version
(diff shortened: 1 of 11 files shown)


waylan:
The upstream bug report is [here](https://github.com/mivok/markdownlint/issues/100) if anyone is interested.

docs/index.md
```diff
@@ -6,37 +6,47 @@ Project documentation with&nbsp;Markdown.
...
-MkDocs is a **fast**, **simple** and **downright gorgeous** static site generator that's geared towards building project documentation. Documentation source files are written in Markdown, and configured with a single YAML configuration file.
+MkDocs is a **fast**, **simple** and **downright gorgeous** static site
+generator that's geared towards building project documentation. Documentation
+source files are written in Markdown, and configured with a single YAML
+configuration file.
 
----
-
-**MkDocs is currently still in development.**
+!!! warning
```
waylan: The linter complained that bold was being used instead of a header. While understandable, this is **not** a header. However, it is a warning to I changed it to an admonition (which I suspect was not an option when this was originally authored).

docs/index.md
```diff
@@ -64,38 +75,47 @@ mkdocs, version 0.14.0
...
 
-MkDocs comes with a built-in webserver that lets you preview your documentation as you work on it. We start the webserver by making sure we're in the same directory as the `mkdocs.yml` config file, and then running the `mkdocs serve` command:
+MkDocs comes with a built-in webserver that lets you preview your documentation
+as you work on it. We start the webserver by making sure we're in the same
+directory as the `mkdocs.yml` config file, and then running the `mkdocs serve`
+command:
 
 ```bash
 $ mkdocs serve
 Running at: http://127.0.0.1:8000/
```
waylan: Unlike in other code blocks, I left the `$` in here because output from the command it also shown. See [MD014](https://github.com/mivok/markdownlint/blob/master/docs/RULES.md#md014---dollar-signs-used-before-commands-without-showing-output)

docs/user-guide/configuration.md
```diff
@@ -158,36 +180,47 @@ Set a list of CSS files to be included by the theme. For example, the
...
+will be passed to the template. This allows for great flexibility when creating
+custom themes.
 
-For example, if you are using a theme that supports displaying the project version, you can pass it to the theme like this:
+For example, if you are using a theme that supports displaying the project
+version, you can pass it to the theme like this:
 
-    extra:
-        version: 1.0
+```yaml
```
waylan: I added language identification to many of the code blocks. The linter was not complaining, but it makes for better output, so I did it anyway. Can't guarantee I got them all though. There is a proposal to add such a rule to the linter which is not yet implemented (see mivok/markdownlint#29).

docs/user-guide/configuration.md
```diff
@@ -205,25 +240,36 @@ index.md     | index.html           | /                        | /index.html
...
+to use another port, or allows you to make the service accessible over your
+local network by using the `0.0.0.0` address.
 
-As with all settings, you can set this from the command line, which can be useful, for example:
+As with all settings, you can set this from the command line, which can be
+useful, for example:
 
-    mkdocs serve --dev-addr=0.0.0.0:80  # Run on port 80, accessible over the local network.
+```bash
+mkdocs serve --dev-addr=0.0.0.0:80  # Run on port 80, on the local network.
```
waylan: Unfortunately, the linter seems to enforce the line length on code blocks also. While generally not  a bad idea, sometimes, there may be a legitimate reason for it. In any event, I was able to shorted the comment without losing the meaning (I think), so no great loss here. If you disagree, please advise and we can improve it.

docs/user-guide/custom-themes.md
```diff
@@ -148,32 +150,39 @@ folder on all pages you would do this:
...
 #### build_date_utc
+
 A Python datetime object that represents the date and time the documentation
 was built in UTC. This is useful for showing how recently the documentation
 was updated.
 
-
 ### Page Context
+
 The page context includes all of the above Global context and the following
```
waylan: Someone has a bad habit of adding an extra blank line before headers (and occasionally not including one after). If this is you, you should probably stop. ;-)

docs/user-guide/custom-themes.md
```diff
@@ -302,7 +316,13 @@ full search implementation to your theme.
...
 
 <form action="search.html">
-  <input name="q" id="mkdocs-search-query" type="text" class="search_input search-query ui-autocomplete-input" placeholder="Search the Docs" autocomplete="off">
+  <input
+      name="q"
+      id="mkdocs-search-query"
+      type="text"
+      class="search_input search-query ui-autocomplete-input"
+      placeholder="Search the Docs"
+      autocomplete="off">
```
waylan: This was an unusually long line in a code block. Actually, I think it is easier to read now that it is reformatting. Wasn't sure if I did the indenting the best way though. Open to suggestions.

docs/user-guide/deploying-your-docs.md
```diff
@@ -77,15 +75,16 @@ where `<projectname>` is the name you used to register your project with PyPI.
...
+1. You must be using [Setuptools] in your `setup.py` script ([Distutils] does
+   not offer an `upload_docs` command).
+1. Your project must already be registered with PyPI (use `python setup.py
+   register`).
+1. Your `mkdocs.yml` config file and your "docs" directory (value assigned to
+   the [docs_dir] configuration option) are presumed to be in the root directory
+   of your project alongside your `setup.py` script.
+1. It is assumed that the default value (`"site"`) is assigned to the [site_dir]
+   configuration option in your `mkdocs.yaml` config file. If you have set a
+   different value, assign that value to the `--upload-dir` option.
```
waylan: The default behavior for the linter is to use `1.` for every order-list item. I prefer the consecutive items better as it is easier (clearer) to read the source. This is [configurable][1], but I couldn't justify creating a "style" file for this one instance of a relatively short ordered-list.

[1]: https://github.com/mivok/markdownlint/blob/master/docs/RULES.md#md029---ordered-list-item-prefix

docs/user-guide/styling-your-docs.md
```diff
@@ -119,17 +117,17 @@ h1 {
 }
 ```
 
-!!! Warning
+!!! note
```
waylan: This seemed more appropriate as a "note" than a "warning" to me. If you disagree, I can change it back. Was not actually raised by the linter.

docs/user-guide/writing-your-docs.md
```diff
@@ -179,25 +213,27 @@ Left         | Center        | Right
 Left         | Center        | Right
 ```
 
-#### Fenced code blocks
+### Fenced code blocks
 
-Start with a line containing 3 or more backtick \` characters, and ends with the first line with the same number of backticks \`:
+The first line should contain 3 or more backtick (`` ` ``) characters, and the
+last line should contain the same number of backtick characters (`` ` ``):
```
waylan: I reworked this paragraph. It made no sense in the previous form. Not specifically a linter issue. Although, the backticks being in code spans rather than escaped seems better to me.

waylan:
Sorry for all the noise, but there were a number of things I wanted to highlight as possible concerns for review to make sure they didn't get missed. I was rather liberal with my changes (perhaps some of the rules should be disabled and the changes backed out?). So I went through the diff and commented on a bunch of stuff. If anyone has any comments or criticisms, let me know.

docs/user-guide/custom-themes.md
d0ugal: We could remove some of it. The CSS class is irrelevant. I copied it from a theme which is why it is longer than we need.

docs/index.md
d0ugal: Seems reasonable, perhaps we could just remove the message in general. It doesn't really give enough information to be helpful.

docs/user-guide/custom-themes.md
waylan: I agree. The docs should probably show the minimal amount required to get things working. But it is not clear what that is in this case. What should this look like?

d0ugal: This should do it. `<input name="q" id="mkdocs-search-query" type="text" >`

docs/index.md
d0ugal: I think we should just remove this line. Telling people that it is still in development isn't really helpful and we don't want to scare people away.

tomchristie: Seconded. Yup we're not at 1.0 yet, but we're using it on major projects and a wide userbase.
If stuff does change between versions we'll make sure to do so with a clear upgrade path.

d0ugal:
Looks good - I can't see any other issues than the last [two](https://github.com/mkdocs/mkdocs/pull/698#discussion_r36953442) [comments](https://github.com/mkdocs/mkdocs/pull/698#discussion_r36953538). This is a great change, while it doesn't totally match my preferences it is good to see the formatting be a bit more standard.

Does this resolve all the warnings?

d0ugal:
Ah, I see there are still [4 warnings](https://travis-ci.org/mkdocs/mkdocs/jobs/72757215). Can include fixes for those and then remove the [exclude from the Travis config](https://github.com/mkdocs/mkdocs/blob/master/.travis.yml#L32-L33).

docs/user-guide/deploying-your-docs.md
tomchristie: This one does marginally bug me :p

d0ugal: Yeah, likewise. If there is a way to easily disable it I'd be happy to see that happen.

tomchristie:
I guess the two things here that it'd be nice to tweak if possible are the code line lengths (be lenient) and the slightly dumbass numbering. Tho I don't know if I care enough about those to justify an extra file in the root directory. (Any other way around that?)

tomchristie:
Also some of the heading changes eg `####` -> `###` make sense, although I suspect that I deliberately chose the alternative depth for styling purposes. Theres more consistency in them now, but hopefullly not too the detriment of the docs look & layout.

waylan:
I just pushed the requested changes (removed the "in development" warning and simplified the form example). And now to address the other concerns raised...

> Ah, I see there are still 4 warnings. Can include fixes for those and then remove the exclude from the Travis config.

As I stated [previously][0] (it probably got lost in all the noise), those failures are a [bug][1] in the linter which needs to be fixed upstream. Unfortunately, my Report has received no response of any kind, so I have no idea if the devs consider it a bug or what kind of timeline we can expect for a fix.The only thing we can do on our end is not use fenced code blocks, which means we loose the ability to tell the code highlighter to not highlight those code blocks (they will look ugly/wrong).

I suppose we could disable that rule, but it is actually a useful rule in most cases, and then we have to have the config file (see below).

[0]: https://github.com/mkdocs/mkdocs/pull/698#issue-97357303
[1]: https://github.com/mivok/markdownlint/issues/100

> I guess the two things here that it'd be nice to tweak if possible are the code line lengths (be lenient) and the slightly dumbass numbering.

I completely agree regarding the numbering. Thankfully, it is configurable...

Unfortunately, there is no way to adjust the "code" line length separately from the rest of the document's line length. We could just increase the line length for everything...

> Tho I don't know if I care enough about those to justify an extra file in the root directory. (Any other way around that?)

Which is why I didn't do it. And no, there is no way around it. Unfortunately, the command line tool doesn't accept individual settings as options. It only has the option to pass in a filename pointing to a config file. And the config file must be a Ruby file, which is even less attractive IMO.

> Also some of the heading changes eg #### -> ### make sense, although I suspect that I deliberately chose the alternative depth for styling purposes. Theres more consistency in them now, but hopefullly not too the detriment of the docs look & layout.

I suspected that was the reason. However, I took a look at the generated output after making the changes (before submitting the PR), and there wasn't any obvious formatting issues with the changes. I think it will be fine.

d0ugal:
Looks good. Let's roll with this for now - hopefully later we can sort out the remaining lint errors so we can enable checking and then over time we might want a config for it to customise things but this is a great step forward.

d0ugal:
Interesting to note that GitHub doesn't support the fenced code syntax. https://github.com/mkdocs/mkdocs/blob/master/docs/user-guide/writing-your-docs.md#fenced-code-blocks

PR #706, opened 2015-08-13
Rename cli.py to __main__.py

d0ugal:
This uses the standard entrypoint name and allows `python -m mkdocs` to
work.


waylan:
:+1:

PR #707, opened 2015-08-13
Add the current page abs_url for RTD integration

d0ugal:


d0ugal:
For more context see https://github.com/rtfd/readthedocs.org/issues/1480

PR #713, opened 2015-08-15
Refactor markdown loading

d0ugal:
Looking again at this work I started in #627 to try and push forward some much needed internal re-factoring.

---

If you want to review this, the second commit had all the internal updates and is the interesting one. The first just includes a dependency and the last updates all the tests so they match the internal changes.

The general goal here is to load the Markdown files much earlier in the process and split the markdown from the metadata. This then means the metadata is more widely available as an attribute of the Page object, and thus the metadata for a page is available to other pages during the build.

---
##### TODO:
- [x] A number of comments inline on the code need to be addressed, see below for those
- [ ] Fix failing tests on Windows - I can't figure out what I've done to break these tests. It is somewhat strange, I am sure it is obvious but I am missing it.

### DIFF
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -10,13 +10,16 @@ from __future__ import unicode_literals
 import datetime
 import logging
 import os
+import io
 
 from mkdocs import utils, exceptions
+from mkdocs.utils import mdutils
+from mkdocs.utils import meta
 
 log = logging.getLogger(__name__)
 
 
-def filename_to_title(filename):
+def _filename_to_title(filename):
     """
     Automatically generate a default title, given a filename.
     """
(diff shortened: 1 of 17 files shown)


mkdocs/nav.py
```diff
@@ -26,6 +29,37 @@ def filename_to_title(filename):
...
+        next_cat_or_title, subpages_or_path = next(iter(entry.items()))
+
+        if isinstance(subpages_or_path, list):
+            paths_to_pages(subpages_or_path, use_directory_urls, docs_dir)
+        elif isinstance(subpages_or_path, utils.text_type):
+            pages_config[i] = _path_to_page(subpages_or_path,
+                                            use_directory_urls, docs_dir,
+                                            title=next_cat_or_title)
+        else:
+            raise Exception('Broken 2')
```
d0ugal: This needs to be completed.

mkdocs/nav.py
```diff
@@ -148,6 +186,63 @@ def __init__(self, title, url, path, url_context):
         self.next_page = None
         self.ancestors = []
 
+        self.markdown, self.meta = self.load_markdown()
+
+    def __eq__(self, other):
```
d0ugal: `__eq__` or `__ne__` were added primarily to make testing easier, I am not sure if that is a good reason or not.

mkdocs/nav.py
```diff
@@ -26,6 +29,37 @@ def filename_to_title(filename):
...
+
+def paths_to_pages(pages_config, use_directory_urls, docs_dir):
+
+    for i, entry in enumerate(pages_config):
+
+        if isinstance(entry, utils.text_type):
+            pages_config[i] = _path_to_page(entry, use_directory_urls, docs_dir)
+            continue
+        elif not isinstance(entry, dict):
+            msg = ("Line in 'page' config is of type {0}, dict or string "
```
d0ugal: This is the same error we had before, but while I am moving it around we should improve it.

waylan:
I'm reopening this as we need to address this and there is no specific issue open to track it. I've add the "update needed" label to remind us that the current code in his PR needs work.

Issue #711, opened 2015-08-14
Support other extensions than `yml` for `mkdocs.yml`


brutus:
This is just a minor cosmetic enhancement, but it would be neat, if other extensions where recognized for the settings file (`mkdocs.yml`). Since I usually use `.yaml` for all my YAML files, it would makes organizing stuff in projects slightly easier and more organized.

tomchristie:
I take it by "other extensions" you *just* mean *.yaml* and *.yml* right?

waylan:
You can name your config file whatever you want. Just pass the filename with the appropriate flag:

```sh
mkdocs build --config-file whatever_name_you_want.anything
```

My point is, I'm not sure there is any benefit to expanding the default to two possible options. What happens when a third option is requested, and a fourth ...

brutus:
@tomchristie: never came across any other. Just thought i keep it general.

@waylan: Yes this works. The annoying part is, that the `--config-file` option comes after the `build` and `serve` arguments… makes it harder to write shell aliases. But as I said, minor cosmetic thing.

waylan:
It is a nitpick, but `build` and `serve` are not arguments, but sub-commands. Although, I suppose to the OS, they are seen as arguments (which is why git sub-commands are actually `git-commandname` under the hood).

Unfortunately, Click does not give us the flexibility to allow options defined on a sub-command to be passed by the user to the parent command and then passed through to the sub-command (annoyingly, Click would require the options to be defined twice, once on the parent command and again on the sub-command--which is not very dry and creates weird output for the `--help` flag).

brutus:
If you would support multiple possibilities for the default filename, it could look something like this:

https://github.com/mkdocs/mkdocs/compare/master...brutus:patch-1

I thought `.yaml` is a really common ending. But I agree, it might be too much hassle, since it is just cosmetics.

brutus:
My original thought was kinda like this: 

Since the _official file extension_ for YAML files (as stated by the maintainers of the spec) is `.yaml`, it should be supported (also see [here](http://stackoverflow.com/questions/21059124/is-it-yaml-or-yml)).

I thick the three letter form `yml` comes from old MS Windows restrictions, but I think that even there you can use extensions like `json` and `yaml` nowadays, so it is no longer needed.

If you dislike the small overhead to check for two different extensions, you could just support `yaml`. Might be not that big of a thing since mkdocs is not at version 1 yet?

tomchristie:
[...]

PR #717, opened 2015-08-18
Support `yaml` as extension for the default configuration file

brutus:
Since the _official file extension_ for YAML files (as stated by the maintainers of the spec) is `.yaml`, it should be supported (also see [here](http://stackoverflow.com/questions/21059124/is-it-yaml-or-yml)).

The last of the supported file names is "the default" and will be reported if missing and no other name is given.

See: https://github.com/mkdocs/mkdocs/issues/711

### DIFF
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -99,9 +99,17 @@ class Config(utils.UserDict):
 
 def _open_config_file(config_file):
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks for the input, but on balance closing as per #711.

PR #721, opened 2015-09-07
Provide the source file name to the JS in readthedocs theme

gregmuellegger:
readthedocs.org needs the file name in order to identify the correct source file to link to the same file in other versions of the same documentation or to the source on GitHub.

See rtfd/readthedocs.org#1480 for more information.

### DIFF
--- a/mkdocs/themes/readthedocs/base.html
+++ b/mkdocs/themes/readthedocs/base.html
@@ -27,6 +27,7 @@
   <script>
     // Current page data
     var mkdocs_page_name = "{{ page_title }}";
+    var mkdocs_page_input_path = "{{ current_page.input_path }}";
     var mkdocs_page_url = "{{ current_page.abs_url }}";
   </script>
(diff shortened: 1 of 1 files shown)


d0ugal:
Looks fine to me. Was mkdocs_page_url not useful? We can remove that if it didn't help.

gregmuellegger:
Hi, the `mkdocs_page_url` was kind of the right thing, but not completely. We use the input_path now by splitting away the file suffix. And what is left is the value we need in readthedocs.

With the page url we would propably be able to get to the same value by stripping the leading and trailing slash, but that feels to dirty and might break if mkdocs once decides that it will append `index.html` to every URL or anything similiar.

You can remove the `mkdocs_page_url` if you want, but I assume it won't hurt to keep it around.

d0ugal:
Sorry for the delay in getting back to this. Thanks - I'll try and get a release out soon.

PR #729, opened 2015-09-22
correct site_url in sitemap

samuelcolvin:
I'm currently getting `<loc>None/</loc>` in my sitemap, see http://help.tutorcruncher.com/sitemap.xml

I believe this should fix it however we should probably add a test, where would be the best place for the test to go?

I can see that the real solution is to set `site_url`, either mkdocs should deal better with no `site_url` or it should be required.

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -163,7 +163,12 @@ class URL(OptionallyRequired):
     Validate a URL by requiring a scheme is present.
     """
 
+    def __init__(self, default='', required=False):
+        super(URL, self).__init__(default, required)
+
     def run_validation(self, value):
+        if value == '':
+            return value
 
         try:
             parsed_url = utils.urlparse(value)


waylan:
I wonder if a better fix would be to set the default to an empty string (`''`) rather than `None`. Change [`mkdocs/config/defaults.py L27`][1] as follows:

```diff
- ('site_url', config_options.URL()),
+ ('site_url', config_options.URL(default='')),
```

My only hesitation is that perhaps somewhere the code is relying on the `None` value for behavior. Hopefully the tests would catch that. Unfortunately, I'm not able to check into it right now.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/defaults.py#L27

d0ugal:
I'd agree with @waylan, we should set the default - then when this is used in other places we wont ever get None, which is never correct. An empty string with a relative path could sometimes be correct at least.

I can't think of anywhere this would cause problems, this value is mostly just passed on I think.

samuelcolvin:
Ok, I've done that but I've actually changed the default on the `URL` class to be an empty string since I thought that made more sense.

Issue #728, opened 2015-09-22
Using a period in a directory name causes the rendered title to include only the left hand side of the directory


justinkinney:
I noticed that using a period in a directory name causes the rendered title to lose everything to the right of the period.

![cursor_and_seespn_docs](https://cloud.githubusercontent.com/assets/466759/10026426/95e8c814-6116-11e5-8226-01e90346c682.png)

To replicate (assuming default docs directory path):
```
mkdir -p docs/Project\ with\ a.dot
echo "#testing" > docs/Project\ with\ a.dot/test\ page\ with\ a.dot.md
mkdocs build
```

I'm guessing the problem comes from [filename_to_title()](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L387), but I don't have a suggestion for how to fix it at this point.

d0ugal:
Hmm, interesting. So Python see's this as all being part of the extension. That is annoying. I am not sure what a good fix would be.

waylan:
`filename_to_title()` calls `os.path.splitext(filename)[0]` on every name given to it. However, that should only be called on filenames, but not on directory names. Given `filename_to_title()` is only passed the base name (not the path) there is no way internally for the function to check whether it has a dir or file name (can't call `os.path.isdir()`). Therefore, that `isdir` test needs to happen in the calling code. Perhaps a second function `dirname_to_title()` could be created. Or require the ext to be removed prior to passing to `filename_to_title()`.

justinkinney:
A naive solution I played with was essentially adding the following to `filename_to_title()`:
```
    for (dirpath, dirnames, filenames) in os.walk('docs'):
        for dirname in dirnames:
            if os.path.isdir('%s/%s' % (dirpath, dirname)):
                title = filename
            else:
                title = os.path.splitext(filename)[0]
```
it "works" but it's obviously pretty frail.

PR #730, opened 2015-09-22
fixes https://github.com/mkdocs/mkdocs/issues/728

justinkinney:
Just a possible solution.

### DIFF
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -393,6 +393,17 @@ def filename_to_title(filename):
     return title
 
 
+def dirname_to_title(dirname):
+
+    title = dirname
+    title = title.replace('-', ' ').replace('_', ' ')
+    # Capitalize if the dirname was all lowercase, otherwise leave it as-is.
+    if title.lower() == title:
+        title = title.capitalize()
+
+    return title
+
+
 def find_or_create_node(branch, key):
     """
     Given a list, look for dictionary with a key matching key and return it's
@@ -431,7 +442,7 @@ def nest_paths(paths):
 
         branch = nested
         for part in parts:
-            part = filename_to_title(part)
+            part = dirname_to_title(part)
             branch = find_or_create_node(branch, part)
 
         branch.append(path)


mkdocs/utils/__init__.py
```diff
@@ -431,7 +435,10 @@ def nest_paths(paths):
 
         branch = nested
         for part in parts:
-            part = filename_to_title(part)
+            dir = False
+            if os.path.basename(part) != path:
```
waylan: Shouldn't this line read:

```python
if os.path.basename(path) != part
```

mkdocs/utils/__init__.py
```diff
@@ -431,7 +435,10 @@ def nest_paths(paths):
...
+                dir = True
```
waylan: Why not change the line above from `!=` to `==` and then change this line to:

```python
part = os.path.splitext(part)[0]
```

Then you can remove the call to `os.path.splitext` in `filename_to_title` and the new `dir` parameter becomes unnecessary.

waylan: Oh right, you can't modify `part` here because the later call to `find_or_create_node` needs the full filename. The general idea should still be doable though.

justinkinney: I'm not sure that works since filename_to_title is called later from `nav.py` to add files to the TOC when `_path_to_page` is called. I tried a few variations on what you described, but to be honest I think they just complicate things a bit more (at least the ways I tried).

waylan: Ah, okay. I wasn't sure if the function was called from anywhere else in the code. Then we definitely need two separate code paths, one for filenames and one for dirnames.

justinkinney:
Ok, based on previous comments, I've added a `dirname_to_title()` method to handle the code path in `nest_paths()`. I also simplified my changes since I realized that `nest_paths()` is only working on paths not filenames due to `directory, _ = os.path.split(path)`.

d0ugal:
Works great, thanks! Sorry for the long delay.

PR #735, opened 2015-09-26
Fix IOError with livereload + custom config file

vi4m:
When using livereload with custom config file, after first refresh you got IOError. 

Steps to reproduce:

`mkdocs serve --config-file=mkdocs-local.yml --livereload`

after first refresh you get IOError (end of stream).

That's because closed descriptor was reused by load_config(), but not reinitialized properly, resulting in
stream IO Error (end of file).  Proposed fix is - to reinitialize closed file description again every time livereload happens.

### DIFF
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -112,7 +112,11 @@ def _open_config_file(config_file):
         else:
             raise exceptions.ConfigurationError(
                 "Config file '{0}' does not exist.".format(config_file))
-
+    else:
+        # reopen config file descriptor when needed
+        if config_file.closed:
+            config_file = open(config_file.name, config_file.mode)
+        config_file.seek(0)
     return config_file
 
 


vi4m:
Any news about it?

d0ugal:
This change looks good, I've not had a chance to test it. A unit test would help to verify this behaviour and get it merged sooner.

I am very busy at the moment due to life, I am unlikely to be active in MkDocs for around a month.

robodude666:
Looks like this patch makes the assumption that the config file still exists on reload.

waylan:
@robodude666 makes a good point. That change which prompts the autoreload could be a change to the config file. If a user inadvertently deleted the config file, they should get the same clean error message on a reload as they would get on the initial load. That doesn't happen with the current patch.

d0ugal:
I think we should take into account the problem @robodude666 spotted. Otherwise happy to accept this.

waylan:
Fixed in #883.

Issue #640, opened 2015-06-18
Warn users if they are using an old/different version


d0ugal:
pip has a nice feature where they warn users if they are using an old version and prompt them to update. I think it would be neat to add something like this. The code they use for it is here: https://github.com/pypa/pip/blob/7.0.3/pip/utils/outdated.py#L95

d0ugal:
I actually think this is becoming more important. Otherwise multiple users could end up deploying the same docs with different MkDocs versions.

For example, Alice is working on Project X and installed MkDocs six months ago. Then Bob joins the project and installs MkDocs today when he wants to deploy the documentation, he gets a more recent version than Alice did. Now when they deploy the documentation, the deployed docs will switch between the old and new version - possibly adding or removing bugs and features.

d0ugal:
Given that we include the MkDocs version in the [git commit](https://github.com/mkdocs/mkdocs/commit/4af6066d62bbbaf8b82ea289bbc87bc62c63e4ae) we could use that information to warn about using older versions than a previous deploy.

waylan:
Ah, so you are suggesting that the `gh-deploy` command check the `gh-pages` branch for the latest commit, extract the MkDocs version from the commit message, do a version comparison with the currently installed version of MkDocs, and issue a warning if they don't match? That might work.

What happens when doth Alice and Bob upgrade? Should they still get the warning the first time they deploy? Or should the warning only issue for Alice before she upgrades? In order words, should we allow newer versions to deploy silently but warn on older versions, or should we always warn on mismatched versions? This is what is not clear to me.

d0ugal:
> In order words, should we allow newer versions to deploy silently but warn on older versions, or should we always warn on mismatched versions? This is what is not clear to me.

Good question. I am not sure. I am not sure I even like the idea in general. I guess we should allow newer versions - otherwise that could be annoying. Although, if it is just a warning, that would be fine.

So, how about:
- If it is an older version, error with a flag to do it anyway.
- It it is a newer version, display a warning/info message "The previous deploy was done with MkDocs version X you are deploying with version Y"
- If it is the same version, don't do anything.

waylan:
That works for me. I like the "info" message in a newer version.

waylan:
[...]

PR #740, opened 2015-10-17
Check if newer versions of MkDocs are available

d-dorazio:
Warn the user if newer version of MkDocs are available. This solves Issue #640.


d0ugal:
This is neat, it is good to see that it is quite easy to do. However, this does a few things that we cannot do...

- This essentially makes pip a dependency - I am not sure I want to do that. The version of people people have installed will vary hugely.
- It also accessed private packages under `pip._vendor`
- Are all the other pip imports considered public? They don't feel public.

I've not reviewed the code beyond this yet.

waylan:
In addition to the concerns raised by @d0ugal, is appears to me that there is a lot of code duplication from pip. A few classes were just copied over wholesale (possibly with a few tweaks - I haven't looked that close). Regardless of any potential licencing issues, would it be possible to just import them and subclass (if necessary)? Of course, that still maintains a dependency on pip, so its not a workable solution, but it needed to be pointed out. Is there not any independent package that serves this purpose?

This is just my opinion, but it seems to me that maintaining such functionality internally is outside the scope of MkDocs. Such a feature should be maintained as a third party dependency (which could be used by any Python package hosted on PyPI). If no such third party package exists, then my inclination is that MkDocs shouldn't include it as a feature.

d0ugal:
> This is just my opinion, but it seems to me that maintaining such functionality internally is outside the scope of MkDocs.

I mostly agree. However, I wasn't able to find such a package. Maybe I'll take another look, it would be quite a fun and useful thing to build on it's own if I can't find one.

d-dorazio:
@waylan just subclassing those two classes isn't enough, because unfortunately in pip the name of the files where to store the metadata is hardcoded.

I'd like to work on this brand new package, because I found myself using the previous solution a couple of times.

d0ugal:
Cool, I'd certainly be interested in seeing what you come up with!

d-dorazio:
Don't expect nothing fancy. I'm not an experienced developer as you certainly have already understood.

waylan:
> just subclassing those two classes isn't enough, because unfortunately in pip the name of the files where to store the metadata is hardcoded.

I suspected it was something like that. What that tells me is that that code was never intended to be used outside of pip and therefore should not be relied upon to work in future releases.

robodude666:
Why not simply detect if pip is installed, and then enable/disable the functionality accordingly?

May be more appropriate discussion for #640, but why the auto-check? If a user is happy with the current feature set, they shouldn't be nagged to upgrade. If they find a bug, they can visit mkdocs.org to see if it's been fixed in a newer version or report it. If you want to communicate new features or releases to a user, start a newsletter/blog and let them subscribe (see [GitLab blog](https://about.gitlab.com/blog/)).

At the very least there should be a command line flag to disable the check. May also not be worth implementing until 1.0.0 when the mkdocs.yml format stabilizes.

waylan:
> Why not simply detect if pip is installed, and then enable/disable the functionality accordingly?

To use pip requires using various private internal pip API calls. For that reason, if such a feature is implemented, it should not rely on pip at all.

> If a user is happy with the current feature set, they shouldn't be nagged to upgrade.

Personally, I agree. I don't really see any value in such a feature myself for a project like MkDocs. That  (among other reasons) is why I suggested that such a feature should be its own third-party library.

d0ugal:
> >If a user is happy with the current feature set, they shouldn't be nagged to upgrade.

> Personally, I agree. I don't really see any value in such a feature myself for a project like MkDocs. That (among other reasons) is why I suggested that such a feature should be its own third-party library.

It was just a wishlist feature of mine and an idea, it is far from an important feature. So, yeah, since it seems to be fairly complicated we shouldn't do it in MkDocs itself. If we can have it provided by a third party package, we should go down that route.

This doesn't seem like the correct direction, it is an interesting POC tho' - thanks for putting it together.

Issue #607, opened 2015-06-07
Improve theme customization


waylan:
In both #597 and #467 possible solutions include making changes to the existing themes so that users can override part of the template without having to re-implement the entire template. While those issues address specific requests, this is intended as an attempt to define a general model/policy for how this should work. Once the decisions are made, someone(s) can go through the templates and start to implement it.

I see three potential approaches:
1.  **Includes**:
   
   Break out each overridable piece of the template into a separate file and use an [include](http://jinja.pocoo.org/docs/dev/templates/#include) in the parent template to pull it in. The parent template would look like this:
   
   ```
   <div>
       <!-- some content -->
   </div>
   
   {% include 'somefeature.html' %}
   
   <div>
       <!-- some more content -->
   </div>
   ```
   
   Then the file `somefeature.html` would simply contain the template code to implement "somefeature".
   
   If a user wants to override "somefeature," she can create a directory and assign it to the [`theme_dir`](http://www.mkdocs.org/user-guide/configuration/#theme_dir) config setting. Within that dir, she then creates a file `somefeature.html` which contains the replacement template for that feature. When the build process runs, Jinja should check the `theme_dir` first for each "included" template and fall back to the dir of the theme assigned to the [`theme`](http://www.mkdocs.org/user-guide/configuration/#theme) config setting. That way, the user defined template gets precedence and a user can override a given feature without messing with the entire template.
   
   For example, to disable analytics, a user would simply create a blank file with the appropriate name (`analytics.html`) and save it to the `theme_dir`. Or if a user wanted to use a service other than Google's, she could create the appropriate file (`analytics.html`) which contains the HTML required by her service of choice.
2.  **Blocks** (template inheritance):
   
   Wrap each overridable piece in a separate [block](http://www.mkdocs.org/user-guide/configuration/#extra), like this:
   
   ```
   <div>
       <!-- some content -->
   </div>
   
   {% block somefeature %}
       <!-- implementation here -->
   {% endblock %}
   
   <div>
       <!-- some more content -->
   </div>
   ```
   
   The default implementation would then be defined right there in the built-in template (as it is now).
   
   To override "somefeature," the user would create a directory and assign it to the `theme_dir` config setting. Within that dir, she then creates a template file which "extends" the theme's base template:
   
   ```
   {% extends "base.html" %}
   
   {% block somefeature %}
       <!-- custom implementation here -->
   {% endblock %}
   ```
   
[...]

PR #745, opened 2015-10-26
Add option to hide Sphinx footer

svogt:
Add an option to hide the Sphinx footer. The official theme also has such an option so I thought to port it

https://github.com/snide/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/footer.html#L45

### DIFF
--- a/mkdocs/themes/readthedocs/footer.html
+++ b/mkdocs/themes/readthedocs/footer.html
@@ -19,5 +19,7 @@
     {% endif %}
   </div>
 
+  {% if hide_sphinx %}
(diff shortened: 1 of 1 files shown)


waylan:
First of all, MkDocs does not have a `hide_sphinx` setting defined (see the [docs][1] for a list of settings), so this won't work. While an additional setting could be added, my understanding it that the goal is to keep the number of settings to a minimum.

Parhaps it would be preferable to address this through theme customization (see #607) rather than adding another setting.

[1]: http://www.mkdocs.org/user-guide/configuration/

svogt:
Sounds good, I'll have a look

Issue #269, opened 2014-12-03
'Edit on Github' should link to the page, not the project


weitzman:
IMO, the Github link should be smartened so that its text is 'Edit on Github' and the hyperlink points to the current page. For example, see the link in upper right of http://read-the-docs.readthedocs.org/en/latest/builds.html

PIP docs [do this well](http://pip.readthedocs.org/en/latest/news.html)

jimporter:
Is this just for the Readthedocs theme? At least for me (using one of the Bootstrap themes), I'd find it strange for "Edit on GitHub" to get higher billing than a link to the project's main page on GitHub. As a user, I think I'd be a lot more likely to want to go to GitHub to download a tarball than I would be to edit the docs.

d0ugal:
@jimporter Yup, that's correct, it is only in the read the docs theme at the moment.

jails:
:+1:, yeah otherwise people click on the edit button and just bail out because there's nothing to edit. Finding the correct `.md` in a repository is just a pain in the ass when you are a newcomer. Will greatly help people to get involved into documentation maintenance.

laserb:
I got this working by changing the following line in the breadcrumbs.html from
`<a href="{{ repo_url }}" class="icon icon-github"> Edit on GitHub</a>`
to
```
<a href="{{ repo_url }}{{ current_page.abs_url.rstrip('index.html').rstrip('/') }}.md" class="icon icon-github"> Edit on GitHub</a>
```

d0ugal:
Neat, I didn't expect it to be this easy - we should mabe come up with a nicer method tho and add a property to the current_page and remove the stripping.

jails:
:+1: would be great to have this in !

d0ugal:
@laserb Did you change your repo_url to point to the docs directory? That's the only way I can get this to work and that seems a little odd. Otherwise I have to do something like:

    <a href="{{ repo_url }}blob/master/docs/{{ current_page.input_path }}" class="icon icon-github"> Edit on GitHub</a>

d0ugal:
Incidently, @laserb you can replace `{{ current_page.abs_url.rstrip('index.html').rstrip('/') }}.md` with `{{ current_page.input_path }}` :)

Mopster:
[...]

PR #752, opened 2015-11-17
Make 'Edit on GitHub' link point at individual pages. Addresses issue #269.

dmehra:
This implements the last suggestion at https://github.com/mkdocs/mkdocs/issues/269#issuecomment-157217372: the 'Edit on GitHub' link takes the user to the specific page they were on (but in view, not edit mode; the user may click on the edit pen to edit). 

No change for 'Edit on Bitbucket' as their model of accessing individual files is not clear to me.

No new tests added, I couldn't think of a useful test here; docs updated. Let me know if this works for upstream, meanwhile I'm using this approach for my docs site.

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -50,6 +50,43 @@ When set, provides a link to your GitHub or Bitbucket repository on each page.
 **default**: `'GitHub'` or `'Bitbucket'` if the `repo_url` matches those
 domains, otherwise `null`
 
+### repo_branch
+
+When set, points "Edit on GitHub" or "Edit on BitBucket" links to the specified branch
+in the repository. Common defaults are 'master' for GitHub and 'default' for BitBucket.
+
+```yaml
+repo_branch: release
+```
+
+**default**: 'master'
+
+### repo_docs_dir
+
+If your docs_dir does not reside at the root of your GitHub repository,
+and repo_link is desired, set repo_docs_dir to be the path from the repository root
+to docs_dir (inclusive). This option is not used for Bitbucket as providing
+direct view/edit links to individual files on Bitbucket is not supported here.
+
+For example, if your repository is `https://github.com/example/repository` with
+project structure of
+
+```
+stuff
+ \_ src
+ \_ docs
+    \_ concepts
+    \_ examples
+```
+
+For properly formed 'Edit on GitHub' links, set repo_docs_dir to
+
+```yaml
+repo_docs_dir: stuff/docs
+```
+
+**default**: `null` (assumes docs_dir to be at the top of the repository)
+
 ### site_description
 
 Set the site description. This will add a meta tag to the generated HTML header.
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -68,6 +68,9 @@ def get_global_context(nav, config):
         # gets passed to themes.
         'repo_url': config['repo_url'],
         'repo_name': config['repo_name'],
+        'repo_branch': config['repo_branch'],
+        'repo_docs_dir': config['repo_docs_dir'],
+        'docs_dir': config['docs_dir'],
         'nav': nav,
         'base_url': nav.url_context.make_relative('/'),
         'homepage_url': nav.homepage.url,
--- a/mkdocs/config/defaults.py
+++ b/mkdocs/config/defaults.py
@@ -75,6 +75,14 @@ DEFAULT_SCHEMA = (
     # "GitHub" or "Bitbucket" for known url or Hostname for unknown urls.
     ('repo_name', config_options.Type(utils.string_types)),
 
+    # Path from the repo root to the docs dir (inclusive), for use with repo_url;
+    # defaults to expecting docs_dir to reside at the top of repo.
+    ('repo_docs_dir', config_options.Type(utils.string_types, default='docs')),
+
+    # Repo branch to use in the "Edit on GitHub" links,
+    # for use with repo_url; defaults to master branch.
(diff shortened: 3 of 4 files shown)


waylan:
Looks good. However, I don't like the name for the `docs_root` setting. It is not obvious from the name that it applies to the repo (MkDocs is all about "docs" after all). I wonder if perhaps a better approach would be to use a nested config setting. Something like:

```yaml
repo:
    url: http://github.com/user/project
    docs_dir: path/to/docs
    branch: master
```

I realize there is a potential backward comparability issue with existing projects, but it is better to get it right now before 1.0.

dmehra:
The nesting didn't work for me right off the bat, so for now, I renamed the option to be clear that it goes with the repo. 

If you'd rather go with the nested approach, I'll give it another try. I think I'd need to extend SubConfig to accept defaults, then subclass it for Repo and move RepoURL validation into that. My first attempt kept running into 
```
ValueError: dictionary update sequence element #15 has length 4; 2 is required
```
so clearly I was doing it wrong. Let me know if this seems worth pursuing.

d0ugal:
> The nesting didn't work for me right off the bat, so for now, I renamed the option to be clear that it goes with the repo.

Yeah, that bit is a bit complicated to do in the config. I'll see if I can get an example of how to make it work.

Generally I think this change looks good, it is a feature I've wanted to see land for some time. I'll try and make some time to try it out soon.

dmehra:
As suggested in https://github.com/mkdocs/mkdocs/issues/269#issuecomment-160955801 I added support for direct "Edit on BitBucket" links. The config options are still flat, not nested. @d0ugal please let me know the next steps here.

waylan:
I pushed some documentation cleanup in #802.

However, I'm not doing any more work on this. It works as-is (did not confirm BitBucket URLs). But I'm not crazy about it.

First of all, there is no way to ensure valid URLs. For example, if any parts of the URL begin or end with a slash (like the [`repo_url`](https://github.com/mkdocs/mkdocs/blob/master/mkdocs.yml#L6)), then you get double slashes in the URL (i.e.; `https://github.com/mkdocs/mkdocs//blob/master/docs/index.md`). While this still generally results in the correct page being displayed, it feels sloppy. I searched for an easy solution in Jinja, and didn't come up with anything. Of course, this is easy in Python and a custom filter could work, but why not just build the "edit_url" in Python when building each page and pass that in as an attribute on the page object (for example: `{{ current_page.edit_url }}`)? No need to build the URL in the template.

Secondly, I still feel like the various config settings for the repo should all be a nested group. However, this means more custom work on the config settings code.

So, I think this could be implemented much better (in Python, rather than the template), but it would take more work than I have the time to spend on it (right now). While the existing solution works, if it is not backward compatible later (when the better solution become available), do we really want to release the feature now?

waylan:
Had a few minutes to play with this and I came up with this:

```diff
diff --git a/mkdocs/commands/build.py b/mkdocs/commands/build.py
index 614173d..c94252a 100644
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -114,6 +114,25 @@ def get_page_context(page, content, toc, meta, config):
     else:
         canonical_url = None
 
+    if config['repo_url'] and config['repo_name'].lower() == 'github':
+        edit_url = util.urljoin(
+            config['repo_url'],
+            'blob',
+            config['repo_branch'],
+            config['repo_docs_dir'],
+            page.input_path
+        )
+    elif config['repo_url'] and config['repo_name'].lower() == 'bitbucket':
+        edit_url = util.urljoin(
+            config['repo_url'],
+            'src',
+            config['repo_branch'],
+            config['repo_docs_dir'], 
+            page.input_path
+        )
+    else:
+        edit_url = None
+
     return {
         'page_title': page_title,
         'page_description': page_description,
@@ -124,6 +143,7 @@ def get_page_context(page, content, toc, meta, config):
 
 
         'canonical_url': canonical_url,
+        'page_edit_url': edit_url,
 
         'current_page': page,
         'previous_page': page.previous_page,
```

It should be noted that this uses the settings config in the existing patch and the template still needs to be changed. It just seems like a much better approach. If someone else wants to run with it that is fine with me. Not sure when I'll have the time to finish it.

d0ugal:
I think we are almost there with this. I have two concerns

- Adding two new settings for such a small feature isn't ideal. I want the total number of settings in MkDocs to be as small as possible, this goes against that.
- I do agree with @waylan's suggestion, this makes much more sense to be done in Python. Then it will be easier for other themes to use.

Ultimately I didn't want to rush it into 0.15 as I mentioned here: https://github.com/mkdocs/mkdocs/issues/796#issuecomment-173525254

marcelstoer:
Shouldn't this be closed in favor of #802?

waylan:
@marcelstoer as per [this comment](#issuecomment-173527465), we are rejecting both and going with the approach roughly outlined in [this comment](#issuecomment-173363860).

I did note some interesting points you highlighted in your [comment](https://github.com/mkdocs/mkdocs/pull/802#issuecomment-186934686) on #802 though. When I have the time, I'll dig in. Of course, if someone beats me to it...

marcelstoer:
> we are rejecting both

Fair enough, sounds like a good plan. Still, shouldn't you close both PRs then?

d0ugal:
Yeah, I probably should close them. I just lost track a bit of the work being done and I didn't want the context to be lost. I think we have an original issue somewhere that we need to make sure has a way forward from these.

marcelstoer:
> I think we have an original issue somewhere

Yep, it's #269.

d0ugal:
Closing this as above.

Issue #751, opened 2015-11-15
Test are failing with python3.5


williamjmorenor:
Hi

I want to report than test are failing with python3.5

Full log is here
https://kojipkgs.fedoraproject.org//work/tasks/4427/11844427/build.log

waylan:
I can confirm this as well. The relevant bits from the tests:

```
$ tox -e py35-min-req
GLOB sdist-make: /home/waylan/code/mkdocs/setup.py
py35-min-req create: /home/waylan/code/mkdocs/.tox/py35-min-req
py35-min-req installdeps: -rrequirements/project-min.txt, -rrequirements/test.txt
py35-min-req inst: /home/waylan/code/mkdocs/.tox/dist/mkdocs-0.15.0.dev0.zip
py35-min-req installed: certifi==2015.9.6.2,click==4.0,coverage==4.0.2,flake8==2.5.0,Jinja2==2.7.1,livereload==2.3.2,Markdown==2.5,MarkupSafe==0.23,mccabe==0.3.1,mkdocs==0.15.0.dev0,mkdocs-bootstrap==0.1.1,mkdocs-bootswatch==0.1.0,mock==1.0.1,nose==1.3.7,pep8==1.5.7,pyflakes==1.0.0,PyYAML==3.10,tornado==4.1,wheel==0.24.0
py35-min-req runtests: PYTHONHASHSEED='1147804715'
py35-min-req runtests: commands[0] | /home/waylan/code/mkdocs/.tox/py35-min-req/bin/nosetests --with-coverage --cover-package mkdocs mkdocs
..............................................................................................................................................FFF....F.......
======================================================================
FAIL: test_content_parser (mkdocs.tests.search_tests.SearchTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/waylan/code/mkdocs/mkdocs/tests/search_tests.py", line 35, in test_content_parser
    title="Title"
AssertionError: Lists differ: [<mkdocs.search.ContentSection object at 0x7fe466d2e710>] != [<mkdocs.search.ContentSection object at 0x7fe466d2e5c0>]

First differing element 0:
<mkdocs.search.ContentSection object at 0x7fe466d2e710>
<mkdocs.search.ContentSection object at 0x7fe466d2e5c0>

- [<mkdocs.search.ContentSection object at 0x7fe466d2e710>]
?                                                     ^^

+ [<mkdocs.search.ContentSection object at 0x7fe466d2e5c0>]
[...]

PR #755, opened 2015-11-17
Add support for Python 3.5

waylan:
Also include a fix for one failing test with HTMLParser. Three additional
tests are still failing for a different problem related to search.

If I work out that issue, I'll push up another commit to fix it here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python: '2.7'
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ env:
 - TOXENV=py34-integration
 - TOXENV=py34-min-req
 - TOXENV=py34-unittests
+- TOXENV=py35-integration
+- TOXENV=py35-min-req
+- TOXENV=py35-unittests
 - TOXENV=pypy-integration
 - TOXENV=pypy-min-req
 - TOXENV=pypy-unittests
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 sudo: false
 language: python
-python: '2.7'
+python: '3.5'
 env:
 - TOXENV=py26-integration
 - TOXENV=py26-min-req
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,9 @@ environment:
     - TOXENV: py34-integration
     - TOXENV: py34-min-req
     - TOXENV: py34-unittests
+    - TOXENV: py35-integration
+    - TOXENV: py35-min-req
+    - TOXENV: py35-unittests
     - TOXENV: flake8
 init:
   - "ECHO %TOXENV%"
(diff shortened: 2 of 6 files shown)


waylan:
Oh, for got to mention this addresses #751.

waylan:
Just pushed another commit which should resolve all issues with Python 3.5. Interestingly, all the issues revolved around HTMLParser.

waylan:
Groan, it appears that Travis (at least, still waiting on appveyor) does not yet support Python 3.5.

waylan:
Hmm, the [Travis build log][1] clearly states: `InterpreterNotFound: python3.5`, however the [docs][2] list 3.5 as being supported. I wonder if we need to migrate to the [new system][3]. Perhaps new Python versions are not being added to legacy?

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/91721711
[2]: http://docs.travis-ci.com/user/languages/python/#Choosing-Python-versions-to-test-against
[3]: http://docs.travis-ci.com/user/migrating-from-legacy/

waylan:
Well Travis runs a lot fast on the new system, but still no [Python 3.5 support][1].

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/91724663

facelessuser:
@waylan Try what I do here: https://github.com/facelessuser/pymdown-extensions/blob/master/.travis.yml

waylan:
Thanks @facelessuser, that did the trick. Doesn't really make sense to me that that would matter unless Py35 is not yet installed on all systems. Perhaps setting the system Python to 3.5 forces the builds to run on a system which has Py35 available.

Except now Flake8 gives an [error][1] because `unicode` is not defined. That was not a problem in Python2.7.

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/91727004

facelessuser:
You can probably try`python -m flake8 .`  but maybe use the correct python.  I think there is a way to call a specific python version in Travis...

facelessuser:
It's kind of a pain.  I'll be happy when travis supports py35 proper.

facelessuser:
The other alternative is to add `# noqa` to things line unicode (which should be in one place).

waylan:
Flake8 is called from Travis through a Tox env. So I just told Tox to use Python2.7 for the Flake8 env. Finally have all tests passing on Travis.

facelessuser:
Nice, that's a good trick to know.

waylan:
I just noticed this line in the Travis build logs:

```
3.5 is not installed; attempting download
```

Apparently Python 3.5 is not actually installed on any of Travis' systems. However, when the config sets that version as the system version, it gets installed as part of the build dependencies. I suppose when that line no longer appears in build logs, we'll know that it is available and we will no longer need to set the system Python to 3.5.

Regardless, in my attempts to work around this, I switched to the new infrastructure and *wow* is it faster. Glad for that.

d0ugal:
Thanks!

waylan:
You can close #751 now that this is merged.

PR #768, opened 2015-11-30
Document the `--help` flag.

### LINKED ISSUE #767: debugging mkdocs
I want to debug a plugin I wrote using Pycharm. The problem I have mkdocs is an executable. I can see build.py which probably gets called by mkdocs. But with what parameters?

waylan:
As of 40dc17e17e5, any mention of the `--help` flag had been removed from
the documentation. A breif summary has been added here. However, the output
of `--help` has not been included as that can change in the future and
updating the docs every time a option is updated become unnesecarily
duplicative (as the command line is self-documenting). This fixes #767.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -199,6 +199,23 @@ with the `--clean` switch.
 mkdocs build --clean
 ```
 
+## Other Commands and Options
+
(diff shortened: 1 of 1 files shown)


d0ugal:
Thanks!

PR #781, opened 2015-12-28
Fix readthedocs theme subnav padding

ksimka:
Padding is broken when using level-3 navigation, this change fixes it.

### DIFF
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -25,8 +25,12 @@ h3, h4, h5, h6 {
     color: #838383;
 }
 
-.wy-menu-vertical .subnav a {
+.wy-menu-vertical .subnav a,
+.wy-menu-vertical .subnav li a {
     padding: 0.4045em 2.427em;
+}
+.wy-menu-vertical .subnav li.toctree-l3 a {
+    padding: 0.4045em 3.641em;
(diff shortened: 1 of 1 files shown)


d0ugal:
I finally got around to testing this, sorry for neglecting it. The change only partly fixes it from my testing, making it slight worse. It indents only the first heading on a page at level-3.

ksimka:
You can see it in action [here ](http://developers.wamba.com/api/services/agencies_index/index.html) (warning: russian language :) The proposed css is in the end of http://developers.wamba.com/api/extra/overload.css . Works for me in the latest Chrome and FF.

d0ugal:
This is the issue I found. See how "Contributing to MkDocs" is indented further than the other titles. This doesn't happen without this patch applied. So I think this change works better on a second level of menu, but when you add a third one as in this screenshot it causes a new issue.

out

ksimka:
Ha, but it's the 4th level, "Contributing" is on the 3rd. Ok, got it, it breaks the 4th level. Btw, does that theme supports the 4th level of nav? I mean css classes.

d0ugal:
True, that is the fourth level in a way. The page is at the third level, then the table of contents is at the fourth level.

I'm not sure about CSS classes. This would require some investigation.

d0ugal:
I'll close this for now, happy to accept an updated version tho' :smile:

Issue #785, opened 2016-01-06
Readthedocs theme display error in IE 10


jdury:
Seems there's a bad detection of htmls "norms" on IE 10 while browsing docs generated by theme readthedocs (last version). See attached screenshots to appreciate difference with chrome. 

![rtd-ie](https://cloud.githubusercontent.com/assets/3641189/12145776/7d623bcc-b490-11e5-977a-0cba538fa70e.png)

![rtd-chrome](https://cloud.githubusercontent.com/assets/3641189/12145777/7d6423a6-b490-11e5-895b-6796dc68959e.png)

IE7 norms are detected by Internet Explorer, so it displays without using new css features, I think. 

Found a workaround by adding this tag in [base.html](https://github.com/mkdocs/mkdocs/blob/3c6a66b38825dc409f1f8e000bd041f57ee225be/mkdocs/themes/readthedocs/base.html) file :

```html
  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"/>
```

d0ugal:
This fix was merged and released in 0.15

PR #786, opened 2016-01-06
Added html meta X-UA-Compatible to properly display pages in IE

jdury:
Fix for #785 to have a proper display on IE of readthedocs theme


d0ugal:
Seems reasonable, but I am unable to verify this myself.

d0ugal:
I've just checked with my local Windows expert! Thanks!

Issue #796, opened 2016-01-13
Release 0.15.0


d0ugal:
I don't currently have time to work on the project (I am not even using MkDocs at the moment), however, I don't want to hold people back from getting a number of fixes in master. I would therefore like to try and see that the current master version works well. So I am looking for a few volunteers to help test and verify master and then I'll roll a release.

These are the outstanding tasks - I would need help with the first, and hopefully the second point if people have time.

- [x] Verify master
- [x] Resolve #764 and #693 
- [x] Update release notes
- [x] push release
- [x] update documentation on mkdocs.org

MadsRC:
I'm currently using MkDocs for my new project, so I'll try to squeeze in some time for helping you out.

d0ugal:
Thanks! If anyone else is interested but not sure how to do it, it's in the docs. http://www.mkdocs.org/about/contributing/#testing-the-development-version

d0ugal:
There are also a few bugs that might be blocking a release. https://github.com/mkdocs/mkdocs/milestones/0.15.0

laythun:
I am just about to start using mkdocs, once i get it working, i'll be available to contribute

waylan:
> There are also a few bugs that might be blocking a release. https://github.com/mkdocs/mkdocs/milestones/0.15.0

Well, #653 can't be a release blocker as it is *not* a bug as I explain in its duplicate (#192).

And for #610, a much simpler fix would be to raise an error when people put their site_dir in their docs_dir. I don't care if it breaks peoples stuff. It makes no sense, they've been told not to do it and it should never have been allowed.

That would remove two of the five potential blockers.

d0ugal:
> And for #610, a much simpler fix would be to raise an error when people put their site_dir in their docs_dir. I don't care if it breaks peoples stuff. It makes no sense, they've been told not to do it and it should never have been allowed.

Hah, I only care because tons of people pinged me about the error! However, given that we display a warning, maybe that is enough. If people are going to ignore warnings and complain at errors I can't please them :) I'll leave the issue for now, but it shouldn't block a release.

d0ugal:
Okay, I've removed the unimportant issues. That leaves us with #764 and #693. I think these are both bugs that were introduced since `0.14.0`. So these need to be verified as issues and fixed.

I might have time this weekend to try and fix #764 (I presume I broke this with the theme changes in master). Then I'll try and remember what was happening in #693.

d0ugal:
[...]

PR #799, opened 2016-01-15
Update release notes

waylan:
This should address #796. I didn't include every little change, but anything of note should be there.


d0ugal:
Thanks!

PR #802, opened 2016-01-19
Improved #752 (Edit on GitHub)

waylan:
So far I've cleaned up the documentation. <del>More may be coming...</del>

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -50,6 +50,43 @@ When set, provides a link to your GitHub or Bitbucket repository on each page.
 **default**: `'GitHub'` or `'Bitbucket'` if the `repo_url` matches those
 domains, otherwise `null`
 
+### repo_branch
+
+When set, points "Edit on GitHub" or "Edit on BitBucket" links to the specified branch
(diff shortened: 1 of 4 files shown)


marcelstoer:
I've got only one objection. Not sure though if it's a valid one as I'm an MkDocs novice and my only experience is with a "MkDocs & RTD-theme" project hosted at RTD.

The new `repo_branch` config parameter means the `mkdocs.yml` can no longer be merged from branch A to branch B (e.g. on GitHub) without some manual post-processing. When the changes in a hypothetical branch `dev` are merged back to `master` you've got to be really careful not to break the "Edit on GitHub" link. If you miss adjusting the `repo_branch` you'll totally confuse users who want to edit and end up on the wrong branch.

Is there no way it can be deduced automatically? At least for RTD environments it could be extracted from the URL.

waylan:
For the record, this is being rejected in favor of the approach outlined [here](https://github.com/mkdocs/mkdocs/pull/752#issuecomment-173363860). That said, the doc improvements here are probably still useful.

d0ugal:
Closing this as we should move forward with the plan here: https://github.com/mkdocs/mkdocs/issues/269#issuecomment-187589013

PR #816, opened 2016-01-31
readthedocs theme: Keep same font-size everywhere

### LINKED ISSUE #801: font-size in RDT theme is different in p and li
This works properly on readthedocs.org and is a CSS bug that can be understood with the following rules:

```
html { font-size: 100% }
body { font-size: 90% } /* only on mkdocs RTD theme */
p { font-size: 16px }
```

Since `li` is not inside `p`, it's not `16px` but `90%`. Since the default font size on (most?) browsers happens to be `16px`, `90%` of `16px` is smaller than the `16px` of `p`. On readthedocs.org, there's not `font-size` on `body`, so `font-size: 100%` from `html` is applied, which is `16px` on (most?) browsers, and all text uses the same `16px` size.

Is the CSS from mkdocs entirely different from readthedocs or is it just because mkdocs 0.14.0 is not in sync with RTD's CSS?

pquentin:
The RTD CSS uses both the `html` and `p` selectors to define font size.
MkDocs used to use `body` to reduce size, which only applied to anything
else than `p`. This patch restores compatibility with RTD by using html,
not body. Fixes #801.

### DIFF
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -1,7 +1,7 @@
 /*
  * Tweak the overal size to better match RTD.
  */
-body {
+html {
     font-size: 90%;
 }
 


pquentin:
I clicked on "Update branch" in GitHub because it was surprising that I was still waiting for "continuous-integration/appveyor", but it could be working as expected? (I can rebase my commit on top of master, by the way.)

waylan:
@pquentin I'm not sure why, but every PR is that way. I wouldn't worry about it.

d0ugal:
Yeah, it is a problem with appveyor. I don't know why it is reporting twice, one pass and one that never finishes. Maybe we should just disable it.

Issue #818, opened 2016-02-03
Collecting a list of MkDocs themes


d0ugal:
With 0.15 out the door, I am starting to hear about new themes - much faster than I expected. I know of three already.

- [Cinder](https://github.com/chrissimpkins/cinder)
- [Alabaster](https://github.com/iamale/mkdocs-alabaster)
- [Materializr](https://github.com/squidfunk/materializr) (not on PyPI yet)

To aid discovery and help promote the great work done by the community I would like to create a gallery for themes. Where should we do it?

- In the MkDocs documentation? We would have the issue of keeping this relevant and up to date
- In the MkDocs wiki? https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes - this is already stale since I first started it and I bet it is usually missed.

waylan:
The model that works for Python-Markdown extensions is to use the community editable (GitHub) wiki (that way I don't need to do anything). The trick is to link to that wiki page from the documentation (both in the instructions for theme developers and users). That way, everyone just follows the link to the "list of third party themes." Make is clear both in the documentation and on the wiki page that it is editable by anyone with a github account and encourage theme devs to add their own to the list.

d0ugal:
Good idea. Let's do that.

d0ugal:
Okay, cool - I think this is now handled! It would be good to make the wiki a bit nicer and include a screenshot for each theme. It should be possible for anyone to do that, but I'll leave this open to remind me to get to it at some point if nothing else.

d0ugal:
The major downside of the wiki is that we can't upload images directly to it. They need to be hosted somewhere and then referenced. I can't think of a nice way to do this - it's making me slightly favour having a dedicated third party themes page in the docs.

waylan:
The wiki is just a git repo. Are you sure its not possible to include images? I agree that committing them locally and pushing it less than ideal, but if it works...

Or why can't the images be hosted in the theme's repo and/or documentation and just linked from the wiki?

d0ugal:
Okay, I managed to add images. That was hard work. It looks a bit better now, but I am not a huge fan still.

Having themes host their images would work - it's just another barrier to entry.

d0ugal:
I do quite like all the screenshots showing the same documentation - it makes the difference in the themes more obvious.

d0ugal:
I'll close this issue, the wiki page is good enough for now. We can see how it evolves over time.

PR #824, opened 2016-02-12
some updates to docs

williamjmorenor:
1- Reference support of python3.5
2- Reference third party themes available (material, alabaster, cinder)

Signed-off-by: William Moreno Reyes <williamjmorenor@gmail.com>


docs/index.md
```diff
@@ -47,7 +47,7 @@ $ pip --version
 pip 1.5.2
 ```
 
-MkDocs supports Python 2.6, 2.7, 3.3 and 3.4.
+MkDocs supports Python 2.6, 2.7, 3.3, 3.5 and 3.5.
```
waylan: You might want to change that to "3.4 and 3.5" rather than "3.5 and 3.5"

docs/user-guide/styling-your-docs.md
```diff
@@ -100,6 +100,30 @@ the [Bootswatch] project.
 
 ![Yeti](http://bootswatch.com/yeti/thumbnail.png)
 
+## Third party themes
+
+There are many themes provides for the community like:
```
waylan: This is a nitpick, but I wouldn't use "many" here. Also, "provided" not "provides".

waylan:
I thought the decision was to list these in the wiki rather in the docs. See #818.

d0ugal:
Yup, happy to take the 3.5 addition in a new PR. Listing themes in the docs isn't sustainable - so we should go ahead with the plan discussed in #818.

PR #832, opened 2016-02-19
Add flag to add theme_dir from command line

pfultz2:
This adds a flag to specify the `theme_dir` from the command line.

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -73,6 +73,7 @@ dev_addr_help = ("IP address and port to serve documentation locally (default: "
                  "localhost:8000)")
 strict_help = ("Enable strict mode. This will cause MkDocs to abort the build "
                "on any warnings.")
+theme_dir_help = "The theme directory to use when building your documentation."
 theme_help = "The theme to use when building your documentation."
 theme_choices = utils.get_theme_names()
 site_dir_help = "The directory to output the result of the documentation build."
@@ -99,9 +100,10 @@ def cli():
 @click.option('-a', '--dev-addr', help=dev_addr_help, metavar='<IP:PORT>')
 @click.option('-s', '--strict', is_flag=True, help=strict_help)
 @click.option('-t', '--theme', type=click.Choice(theme_choices), help=theme_help)
+@click.option('-e', '--theme-dir', type=click.Path(), help=theme_dir_help)
 @click.option('--livereload/--no-livereload', default=True, help=reload_help)
 @common_options
-def serve_command(dev_addr, config_file, strict, theme, livereload):
+def serve_command(dev_addr, config_file, strict, theme, theme_dir, livereload):
     """Run the builtin development server"""
 
     logging.getLogger('tornado').setLevel(logging.WARNING)
@@ -112,6 +114,7 @@ def serve_command(dev_addr, config_file, strict, theme, livereload):
             dev_addr=dev_addr,
             strict=strict,
             theme=theme,
+            theme_dir=theme_dir,
             livereload=livereload,
         )
     except (exceptions.ConfigurationError, socket.error) as e:
@@ -124,15 +127,17 @@ def serve_command(dev_addr, config_file, strict, theme, livereload):
 @click.option('-f', '--config-file', type=click.File('rb'), help=config_help)
 @click.option('-s', '--strict', is_flag=True, help=strict_help)
 @click.option('-t', '--theme', type=click.Choice(theme_choices), help=theme_help)
+@click.option('-e', '--theme-dir', type=click.Path(), help=theme_dir_help)
 @click.option('-d', '--site-dir', type=click.Path(), help=site_dir_help)
 @common_options
-def build_command(clean, config_file, strict, theme, site_dir):
+def build_command(clean, config_file, strict, theme, theme_dir, site_dir):
     """Build the MkDocs documentation"""
     try:
         build.build(load_config(
             config_file=config_file,
             strict=strict,
             theme=theme,
+            theme_dir=theme_dir,
             site_dir=site_dir
         ), clean_site_dir=clean)
     except exceptions.ConfigurationError as e:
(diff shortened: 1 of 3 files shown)


waylan:
Just curios, what is the use case for this? Seems to me that you would always want the theme dir to be controlled by the config file. But maybe I'm missing something.

pfultz2:
> Just curios, what is the use case for this?

Its the same use case for the `--theme` flag. You want to be able to specify a different theme at build time. Since, I am using a custom theme, then it needs to be specified using `theme_dir` instead of just `theme`.

waylan:
I understand using a different theme at build time (with the `--theme` flag. However, under what situation would there be more than one `theme_dir` outside of the  `--theme` flag.

To clarify, as I understand it, a complete theme should be in a separate package and called with the  `--theme` flag. The `theme_dir` would be where the user would provide a few replacements/overrides/additions to the theme defined in  `--theme`. The only reason to override `theme_dir` at build time is if you have an entire interdependent theme there, which is not the intended purpose. In fact, I would recommend against it as the `theme` will still be loaded, with `theme_dir` overriding it. If you are not careful (or in future updates), you could get some remnants of `theme` still coming through.

So my point is that if it is bad practice to have an entire theme in `theme_dir`, then a `--theme_dir` flag should not be supported as a means to discourage it. Granted, this would be more clear if #607 was completed.

waylan:
Hmm, although perhaps if `theme=None`, then maybe `theme_dir` could be an independent theme. Not sure if that works though.

pfultz2:
> To clarify, as I understand it, a complete theme should be in a separate package and called with the --theme flag.

Exactly how is that accomplished with a custom theme? Where should custom themes be installed so they are found by the `--theme`? Because right now it looks like `--theme` only gives the ability to use a set prescribed themes.(ie `flatly|mkdocs|simplex|cosmo|readable|slate|amelia|cyborg|readthedocs|spacelab|united|bootstrap|cerulean|journal|yeti`)

pfultz2:
> Hmm, although perhaps if theme=None, then maybe theme_dir could be an independent theme. Not sure if that works though.

Thats exactly how I use `theme_dir`.

waylan:
> Where should custom themes be installed so they are found by the `--theme`? 

I don't believe it is documented, but the themes use setuptools entrypoints. When you install a theme as a separate Python package, it gets registered as a MkDocs theme and becomes available by the name it was registered as. In fact, everything except the `mkdocs` and `readthedocs` themes are separate packages which use this method (see [here][1] and [here][2]). The difference with them is that they are listed as dependencies and get installed by default when you install MkDocs. However, that may change in the future. And there are other [third-party themes][3] which you can install yourself.

[1]: https://github.com/mkdocs/mkdocs-bootswatch
[2]: https://github.com/mkdocs/mkdocs-bootstrap
[3]: https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes

waylan:
> > Hmm, although perhaps if theme=None, then maybe theme_dir could be an independent theme. Not sure if that works though.

> Thats exactly how I use theme_dir.

Okay, that is the use case I was looking for to justify this. I'm still not sure its a good idea, but I can at least understand it.

d0ugal:
> I don't believe it is documented, but the themes use setuptools entrypoints. When you install a theme as a separate Python package, it gets registered as a MkDocs theme and becomes available by the name it was registered as.

It is documented! http://www.mkdocs.org/user-guide/custom-themes/#packaging-themes

Granted the documentation may not be great. I'd love feedback suggestions (or maybe we just need to make it more visible somehow?)

d0ugal:
> Hmm, although perhaps if theme=None, then maybe theme_dir could be an independent theme. Not sure if that works though.

It does work and it is most useful when you are working on a theme itself. For example, this is how the official packaged themes use their own theme - https://github.com/mkdocs/mkdocs-bootswatch/blob/master/mkdocs.yml#L10

It means you don't need to worry about installing dev versions or maybe I am already using a MkDocs install that has the theme included.

This use case is why I could see this addition being useful. I want to quickly test the MkDocs documentation (or any other docs) against a theme that I am working on.

mkdocs/__main__.py
```diff
@@ -99,9 +100,10 @@ def cli():
 @click.option('-a', '--dev-addr', help=dev_addr_help, metavar='<IP:PORT>')
 @click.option('-s', '--strict', is_flag=True, help=strict_help)
 @click.option('-t', '--theme', type=click.Choice(theme_choices), help=theme_help)
+@click.option('-e', '--theme-dir', type=click.Path(), help=theme_dir_help)
```
d0ugal: Does `-e` make sense? `-d` seems more logical to me.

pfultz2: `-d` is already used for `--site-dir`. I didn't want to have different options between `serve` and `build`.

d0ugal:
> > > Hmm, although perhaps if `theme=None`, then maybe `theme_dir` could be an independent theme. Not sure if that works though.

> > Thats exactly how I use theme_dir.

> Okay, that is the use case I was looking for to justify this. I'm still not sure its a good idea, but I can at least understand it.

I think it is a reasonable use case. For example, a theme with company branding doesn't belong on PyPI. Then if it isn't going to be distributed, the effort required for packaging is higher than it is worth (particularly if you are not familiar with the lovely quirks of Python packaging.)

So, until we revisit theme customisation etc. more generally, I think we should make it easier to use the current options. I'd be happy for this change to be added. For that to happen, we need an addition in the release notes (under version 0.16.0).

Otherwise the change looks good to me, but I'd like to give it a quick test at some point too.

mkdocs/__main__.py
waylan: `-d` is used for [`--site_dir`][1] on the build command. And as `--theme_dir` is being added to both `build` and `serve`, the short flag should be consistent (`--site_dir` makes no sense for `serve`).

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/__main__.py#L127

d0ugal: So it is! Sorry, I forgot. I think this is probably the best option then.

pfultz2:
Ok, so I added a note about the new flag under the release notes for version 0.16.

d0ugal:
Thanks!

PR #836, opened 2016-02-23
Windows & Unicode

d0ugal:


mkdocs/utils/compat.py
```diff
@@ -0,0 +1,55 @@
...
+    # The rest of this function is a direct copy from cpython/Lib/nturl2path.py
+    # with only a few PEP8(!) fixes.
+
+    """OS-specific conversion from a file system path to a relative URL
+    of the 'file' scheme; not recommended for general use."""
+    # e.g.
+    #   C:\foo\bar\spam.foo
+    # becomes
+    #   ///C:/foo/bar/spam.foo
+    import urllib.parse
```
waylan: `urllib.parse` is Python 3 only. Just remove the `.parse` for Python 2.

mkdocs/utils/compat.py
```diff
@@ -0,0 +1,55 @@
...
+    if ':' not in p:
+        # No drive specifier, just convert slashes and quote the name
+        if p[:2] == '\\\\':
+            # path is something like \\host\path\on\remote\host
+            # convert this to ////host/path/on/remote/host
+            # (notice doubling of slashes at the start of the path)
+            p = '\\\\' + p
+        components = p.split('\\')
+        return urllib.parse.quote('/'.join(components))
```
waylan: This should be `urllib.quote(...)` in Python 2.

mkdocs/utils/compat.py
waylan: Maybe just do `from urllib import quote` as `quote` is the only thing actually used.

d0ugal:
Hah, so this failed with the same error. It seems the problem is in `quote` and not `pathname2url`. It looks substantially harder to port over.

waylan:
Well your test passed. But I don't see that the page was actually built in the logs. (see [L1061][1]). In fact I don't see any build log for that "project".

[1]: https://ci.appveyor.com/project/d0ugal/mkdocs/build/2.796/job/6engg8xobnv64paw#L1063

waylan:
Ah, the `-v` flag is missing on the "integration" tests (see [L48][2]), but is included on the "theme" tests (see [L41][1]). Therefore, we don't get verbose output for them. I noticed because I wanted to see what the generated URL actually is.

[1]: https://github.com/mkdocs/mkdocs/blob/14016b0da7d06ba30f24b879e13594abde92e309/mkdocs/tests/integration.py#L41
[2]: https://github.com/mkdocs/mkdocs/blob/14016b0da7d06ba30f24b879e13594abde92e309/mkdocs/tests/integration.py#L48

d0ugal:
Oh, well spotted. I'll fix that.

mkdocs/utils/__init__.py
```diff
@@ -321,7 +321,7 @@ def path_to_url(path):
     if os.path.sep == '/':
         return path
 
-    return pathname2url(path)
+    return pathname2url(path.encode('utf8'))
```
d0ugal: So, given the failures maybe we should only do the encode on windows and Python 3?

mkdocs/tests/integration.py
```diff
@@ -35,17 +35,20 @@
               required=True)
 def main(output=None):
 
+    print("Building themes.")
```
waylan: These print statements are not getting printed inline with the log statements. Instead they are all groupeds together at the end. These should be writing to the log rather than print statements. See the log [L1146-L1164](https://ci.appveyor.com/project/d0ugal/mkdocs/build/2.812/job/sy52anb8i5vtpo0r#L1146)

d0ugal: hah, whoops. That was a bit silly of me. Thanks - I'll take a look.

d0ugal: Looks better now: https://travis-ci.org/mkdocs/mkdocs/jobs/111585666#L995

PR #841, opened 2016-02-24
Remove external themes from MkDocs

d0ugal:
This change removes the dependency on mkdocs-bootstrap and mkdocs-bootswatch.
It also updates all references in the code. The only remaining connection is
when a user specifies one of these themes but they are not installed they get
slightly helpful error.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -52,7 +52,7 @@ details about these specific themes.
 [MkDocs Bootstrap]: http://mkdocs.github.io/mkdocs-bootstrap/
 [MkDocs Bootswatch]: http://mkdocs.github.io/mkdocs-bootswatch/
 
-They will be included with MkDocs by default until the 1.0 release. After that
+They will be included with MkDocs by default until a future release. After that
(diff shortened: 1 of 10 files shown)


d0ugal:
The original plan was to do this for 1.0, but I don't see an immediate 1.0 happening, so it is better to start moving things along.

docs/about/release-notes.md
```diff
@@ -51,7 +51,7 @@ details about these specific themes.
 [MkDocs Bootstrap]: http://mkdocs.github.io/mkdocs-bootstrap/
 [MkDocs Bootswatch]: http://mkdocs.github.io/mkdocs-bootswatch/
 
-They will be included with MkDocs by default until the 1.0 release. After that
+They will be included with MkDocs by default until a future release. After that
```
d0ugal: This is the bad change, it's a slight change in contract. However, given that `1.0` is somewhat mythical and hasn't been defined I think it is okay.

d0ugal:
I suspect it is okay to do this, and nobody has shouted at me so I'll probably go ahead with it soon.

waylan:
No objections from me.

PR #845, opened 2016-02-26
It had Bootstrap instead of Bootswatch in documentation

lesleyauk:
I also improved readability and reduced confusing repetition of the word "theme" here.

PR #847, opened 2016-02-27
Fixed typo so that Bootswatch theme properly id'ed

hartzell:
Fixed typo so that Bootswatch theme properly identified in documentation, Bootstrap => Bootswatch.


waylan:
This is a duplicate of #845 which also fixes the same issue.

PR #851, opened 2016-03-01
Allow using anchors when specifying pages

pfultz2:
This adds a patch where I can write the pages with anchors like this:

```yml
pages:
- section:
    - A: 'page.md#a'
    - B: 'page.md#b'
```

This allows for me to write a more detailed table of contents in some instances.

### DIFF
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -206,6 +206,9 @@ def _path_to_page(path, title, url_context, use_directory_urls):
     if title is None:
         title = filename_to_title(path.split(os.path.sep)[-1])
     url = utils.get_url_path(path, use_directory_urls)
+    if '#' in path:
+        path, anchor = path.split('#')
+        url = url + '#' + anchor
     return Page(title=title, url=url, path=path,
                 url_context=url_context)
 


d0ugal:
I understand what this feature does, but I don't understand why you would ever want to do this. A table of contents that contains some links to pages and some links to anchors would be confusing.

I don't think this belongs in MkDocs. It migth be that changes like this would be good plugin ideas, when #206 is done.

pfultz2:
> but I don't understand why you would ever want to do this.

So the user can access more of the documentation directly from the table of contents. The user can't see the other headers in files and I would like to make that visible to more easily find that information. 

Ideally, it would be nice if we could define another level down for the anchors like this:

```yml
pages:
- section:
    - page:
        - A: 'page.md#a'
        - B: 'page.md#b'
```

Then the themes could decide whether they want to support this deeper nesting or not in some way.

However, I understand currently that mkdocs doesn't support this deeper nesting, so I am just moving the anchors up one level. This way I can still make the anchors visible to the user in the toc.

> A table of contents that contains some links to pages and some links to anchors would be confusing.

I don't see that as confusing at all. It could be a little confusing if I mix them under the same section, like this:

```yml
pages:
- section:
    - A: 'page1.md'
    - B: 'page2.md#b'
```

Of course, this is partly dependent on the theme, and/or how the documentation is presented. For example, if I have a reference section in the documentation, it would make sense that this section would jump to different pages in the documentation.

d0ugal:
> However, I understand currently that mkdocs doesn't support this deeper nesting

You can nest the pages as deep as you like, you just need a theme that supports it.

I'm still not convinced, but I'll leave this open and see if there is interest from others.

pfultz2:
> You can nest the pages as deep as you like, you just need a theme that supports it.

Ok cool, I didn't know that. Are there themes that currently support this?

Of course, I still need a way to point to the anchors in a page.

> I'm still not convinced, but I'll leave this open and see if there is interest from others. 

It seems to be a natural extension of the deeper nesting. So instead of a menu just to the page, I can have a submenu to each of the page's main headers.

d0ugal:
> > You can nest the pages as deep as you like, you just need a theme that supports it.

> Ok cool, I didn't know that. Are there themes that currently support this?

Yeah, most themes do. mkdocs and readthedocs have done so for a long time now, the Bootswatch themes just gained support very recently (in the last few weeks) and many of the external MkDocs themes also support it.

You can see a Bootswatch theme support it here:
http://mkdocs.github.io/mkdocs-bootswatch/

> It seems to be a natural extension of the deeper nesting. So instead of a menu just to the page, I can have a submenu to each of the page's main headers.

It doesn't seem like a natural extension to me to be honest, and given this is the first time the idea has been suggested in over 2 years, I don't think it is an expected feature by anyone else.

waylan:
I have to agree that this does not seem like a natural extension either. Although, it does occur to me that there have been a few requests to include external links in the nav. I suppose this could be addressed through whatever solution provides for that--if such a feature is ever supported. I couldn't find the relevant issue just now, but I know I've seen requests for that before.

pfultz2:
> It doesn't seem like a natural extension to me to be honest, and given this is the first time the idea has been suggested in over 2 years, I don't think it is an expected feature by anyone else.

Well, I have been also unaware of the deeper nesting of docs as well. So, I believe more users will start to expect this kind of feature as they become more aware of the deeper nesting.

d0ugal:
Okay, we will wait and see if that happens. I'll close this for now.

pfultz2:
Ok well, I have a need for this feature now.

pfultz2:
Well I do have a need for this feature now, and I would really wouldn't like to use a fork of mkdocs either.

d0ugal:
Sorry! We have a severe lack of maintainer time on the project so we need to be really careful what we accept. This isn't likely to change unless we get some sponsorship for the project.

pfultz2:
What more is needed for maintenance? Would adding tests for this help?

waylan:
What @d0ugal is saying is that this project is maintained by volunteers who work on it in their limited spare time. Without an incentive (monetary or otherwise) to dedicate additional time to the project, we simply cannot add additional features which will likely result in more work (support questions, bugs, etc) down the road.

pfultz2:
I understand that this a volunteer effort. However, I am not asking for additional work. I volunteered and wrote this feature, and would like to know what else I can contribute to get this feature into mkdocs. If you are only looking for monetary contributions then I think a fork of this project would be the best option.

d0ugal:
> However, I am not asking for additional work.

Sorry, I know it doesn't feel like much, but you are asking for extra work. Every new feature needs to be tested, maintained and documented. Every new feature needs to be considered when something else is added, because we can't break existing features. No MkDocs developers have shown interest in this feature, so you are asking us to do this for you.

Contributing code to a project is like giving somebody a puppy, they still need to look after it for the foreseeable future.

> If you are only looking for monetary contributions then I think a fork of this project would be the best option.

No, that is not just what we are looking for. However, company sponsorship is one way that I can see MkDocs progressing and growing faster - it would potentially allow somebody to spend more time on the project in general. Having said that, I don't know if that would even help, I understand all the MkDocs developers are gainfully employed and nobody is looking for new work opportunities that I am aware of.

Sorry for the confusion, I hope that is clearer.

Issue #842, opened 2016-02-25
Images in subpages named index.md have the incorrect path on Windows/cygwin


styfle:
I am using the following markdown in `index.md` and running `mkdocs serve`

``` md
## Diagram
This is the svg
![svg](diagram.svg)
```

The output is correct

``` html
<img alt="svg" src="./diagram.svg" /></p>
```

However I get a broken image as seen below

![image](https://cloud.githubusercontent.com/assets/229881/13328070/b29facb0-dbba-11e5-8629-61b6b2345a8d.png)

It appears that the svg file is served as `application/octet-stream` instead of the expected `image/svg+xml`

d0ugal:
Interesting. Can you try with this option and let me know if that solves the svg issue?

    mkdocs serve --no-livereload

styfle:
@d0ugal I tried and it does not solve the issue.

d0ugal:
Interesting, I tried with a random SVG I found online and the response headers stated `Content-Type:image/svg+xml` but the image was still broken. This happened with and without livereload enabled.

d0ugal:
Ah, I tried again with another SVG file and it worked fine with and without livereload. The content type was correct and the image rendered.

Can you let me know your mkdocs version? Either with `mkdocs --version` or `pip freeze | grep mkdocs`

waylan:
As livereload is a simple wrapper around tordano's static file server, I checked and tordano seems to only ever have had one relevant issue filed regarding SVG (tornadoweb/tornado#1637). And according to that issue, SVG files have always been served as `image/svg+xml`. Not sure why any would be served as `application/octet-stream`.  All that PR did was add SVG to the list of types that get GZip compression and it appears to have not made it into a release yet. And livereload has never had any SVG related issues filed.

@styfle all I can suggest is to make sure you are using the latest release of livereload, tornado, and MkDocs.

d0ugal:
Yeah, my best guess is that it is a really old MkDocs from when we hacked together out own server.

styfle:
@d0ugal @waylan I'm using the latest as far as I can tell.

```sh
$ python --version
Python 2.7.10

$ pip --version
pip 8.0.2 from /usr/lib/python2.7/site-packages (python 2.7)

$ mkdocs --version
mkdocs, version 0.15.3
```

styfle:
I'm running on Windows 10 with Cygwin.
Here is the pip packages which might be useful.
It looks like the latest version of tornadoweb.

```
[...]

PR #852, opened 2016-03-01
Added basic subpages integration test

d0ugal:
See #842 

This test isn't great yet, as it just verifies that the build doesn't error - not that it is correct. However, it is useful to have a premade test of this available.


d0ugal:
I'm going to merge this, at the moment it shouldn't ever really fail - unless there is a Python exception - so it might not be that useful. However, it is handy to have these projects for semi-manual testing (i.e. manually verifying the end of a Tox run). At some point it would be good to have some post-build tests for these that check for specific things in the build output. Checking all the expected files are there would be a start, then checking that links are correct etc. would be even nicer.

PR #857, opened 2016-03-02
Remove font-size in readthedocs/theme_extra.css

styfle:
If you do a side-by-side comparison of [docs.readthedocs.org](https://docs.readthedocs.org/en/latest/builds.html#mkdocs) you can see that mkdocs font is smaller and the colors are different for headers (mkdocs uses blue and readthedocs uses black).

This removes those two discrepancies.


d0ugal:
Yup, that does look a bit closer. Not really sure why those styles were added.

PR #858, opened 2016-03-03
Add a bit more padding to inline `code`

styfle:
This matches readthedocs.org much better since there was previously 0 padding top/bottom for <code>variable</code>.
See this page for a comparison: https://docs.readthedocs.org/en/latest/i18n.html


d0ugal:
Yup, this looks better. Thanks!

Issue #859, opened 2016-03-03
Large search indexes can cause Lunr to freeze the page while it creates the index.


sashadt:
When using readthedocs theme, search shows "no results" for a few seconds while search is being performed, and then results are loaded.  However, few seconds are enough to convince the users nothing is found, and they move on.

For example see:  http://docs.datatorrent.com  Try searching for "test" and you will see the following for a while:

![image](https://cloud.githubusercontent.com/assets/4966983/13509416/2605d938-e140-11e5-9a6b-6ef8446212cc.png)

Perhaps an indicator of search in progress can be added?

waylan:
> Perhaps an indicator of search in progress can be added?

That seems reasonable to me. Presumably the problem occurs due to the large number of pages. On a smaller site, this would not be a problem as the wait time would be much less.

sashadt:
We have 50 files with 18,936 lines of Markdown.  I don't know if that's considered large, but even for smaller projects running on slower servers it would provide a better user experience.  Thanks for looking into it!

waylan:
It appears that the RTD theme includes the default content "Sorry, page not found." on the search results page (see [search.html#L18][1]). However, the search tool will also replace any content with `<p>No results found</p>` (see [search.js#L63][2]) which makes the initial filler redundant. It appears that RTD displays nothing. Should we do the same or display a message to wait for results?

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/themes/readthedocs/search.html#L18
[2]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/assets/search/mkdocs/js/search.js#L63

d0ugal:
This is quite interesting. I assumed the delay was the length of time it takes to download the search index (about 0.5mb). However, there is a long delay after it is downloaded. So is the delay while lunr.js is processing the index? Meaning the slowness is probably more dependant users machines than the server. Looking a the Lunr change log there are a few performance improvements in [`0.5.8`](https://github.com/olivernn/lunr.js/blob/master/CHANGELOG.mdown#058) and later (we ship `0.5.7`). So I think there are two things we should do:

- Add a loading indicator (this could just be static text saying "Searching...") and remove the incorrect text in ReadTheDocs's search.html
- Update Lunr.js to `0.6.0`

d0ugal:
[...]

PR #860, opened 2016-03-04
Update default text when search is running.

waylan:
The search tool will replace the text with "No results found" if that is the case.
No reason to display is here. Also, the default text displays when the search 
is running (especially if the search takes a long time), so it might as well be 
accurate and indicate that the search is running. Addresses #859.

### DIFF
--- a/mkdocs/themes/readthedocs/search.html
+++ b/mkdocs/themes/readthedocs/search.html
@@ -15,7 +15,7 @@
   </form>
 
   <div id="mkdocs-search-results">
-    Sorry, page not found.
+    Searching...
   </div>
 
 {% endblock %}


d0ugal:
(Interesting, it looks like if you create a branch of the MkDocs repo it triggers twice as many CI runs. I always create the branches on my fork.)

Looks good to me!

waylan:
Weird, I thought I created that branch on my fork. Although I did use Github's new web interface for the first time (which now allows you to create branches and such). Guess I was a little careless. Live and learn.

d0ugal:
Not a problem!

sashadt:
This is awesome!  Big help for our users.  Thanks guys!

PR #881, opened 2016-03-31
Support gh-deloy in Python 3 on Windows.

waylan:
In PY3 calls to `subprocess.Popen` must have `universal_newlines=True` so that Unicode strings are returned.

Note: I have not tested this as I have no way to. And the tests don't cover this either (they are only mock tests). We need to hear from a real person who had the problem before this fix is applied who no longer has the problem afterward.

### DIFF
--- a/mkdocs/tests/utils/ghp_import_tests.py
+++ b/mkdocs/tests/utils/ghp_import_tests.py
@@ -9,6 +9,7 @@ import subprocess
 import tempfile
 import unittest
 import shutil
+import sys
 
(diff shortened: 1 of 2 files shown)


d0ugal:
:+1:

waylan:
After finally being able to test this, I confirmed that this does not, in fact, work. I could try to fix the bugs in it, but in the end, the actual fix for the problem this was trying to resolve was pretty simple (see #904). Therefore I'm closing this. It might still make sense to do this in the future, but if it an't broke, don't fix it...

PR #910, opened 2016-04-26
Update 0.16 release notes to include recent changes.

waylan:
This addresses the various fixes I pushed recently. I need to do a better job remembering to include these in each fix.


d0ugal:
Neat, thanks!

Issue #899, opened 2016-04-14
Run a LintChecker against a build of each theme in CI tests.


waylan:
Perhaps we should run a LintChecker against a build of each theme in our CI tests to avoid bugs like #894. The [LintChecker](https://wummel.github.io/linkchecker/) lib looks like it may work (no personal experience, perhaps another is better?). It can run against `http://` and `file://` URLs to ensure no links return 404s (or other errors).

The trick is if we go with `http://` then we need to have the dev server run on the CI server for the duration of the test and then shut it down. Using `file://` should be easier. I expect issues with stricter browser security on `file://` URLs wouldn't apply in this context, but I can't say for sure. May have to explore the possibilities.

anantshri:
Simplest option would be to run simple http server in the directory and let linkchecker do its job. I did exactly that to identify the issue in first place.

d0ugal:
👍  Sounds like a good idea. I assume something like this would work on Travis.

    python -m SimpleHTTP &
    listchecker http://localhost:8000/

If we wanted to get extra clever about it we could use the sitemap to test all the pages.

d0ugal:
I have added this but without `--check-extern` as it was causing problems with HTTPS Google addresses. It was working for some time, but then stopped. I am not sure why. I am happy for anyone to have a go at adding it in, because I think it would be useful.

I am going to merge #913 once CI finishes, but I'll leave this open for `--check-extern`

waylan:
> It was working for some time, but then stopped.

Just a guess, but Google may have a rate limiter that is denying the requests. Which reminds me that LinkChecker has the following option:

```
-PNUMBER, --pause=NUMBER
Pause the given number of seconds between two subsequent connection requests to the same host. Default is no pause between requests.
```

Although, it may not be the time between requests, but the number of requests from the same IP address that is the problem. In either event, this would need to be addressed later.

PR #912, opened 2016-04-27
Add a linkchecker test for each theme.

waylan:
Addresses #899. Not sure if this will work. Will have to wait and see what Travis reports.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,11 +27,8 @@ env:
 - TOXENV=markdown-lint
 - TOXENV=mkdocs-themelint
 - TOXENV=rtd-themelint
-addons:
-  apt:
-    packages:
-      linkchecker
 install:
+- pip install linkchecker
 - pip install tox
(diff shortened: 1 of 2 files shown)


d0ugal:
Interesting, codecov has started adding comments. I don't think we want that added noise? I'll take a look at turning them off later.

waylan:
> Interesting, codecov has started adding comments. I don't think we want that added noise? I'll take a look at turning them off later.

I wondered about that. Thought maybe you changed something. Guess not.

waylan:
Hmm, looks like the commands are not actually running.

d0ugal:
I shuffled things around a little bit in #913, it is now simpler but checks all the builds from the integration tests (so all themes with the mkdocs docs plus a few extra docs sites that test specific things like unicode). It seems to be working now, let me know what you think of this approach.

waylan:
Closing this in favor of #913.

PR #913, opened 2016-04-28
Linkchecker

d0ugal:


d0ugal:
There is only one remaining error, I'm not sure what to do about it. It is there because we have this in the `index.md` and no server running like you would if following the docs.

    <http://127.0.0.1:8000/>

d0ugal:
... hah, as it happens, the version of linkchecker on Travis doesn't have `--check-extern` so I think removing that resolves the outstanding error (for now).

waylan:
Ahh, I completely missed that the integration tests were already doing builds.

.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: false
```
waylan: By using `sudo` we are no longer in the [container based infrastructure][1], which runs much much faster. We switched over in 3c0849b and it made a huge difference in how fast the tests were completed. I would strongly recommend staying with the container based infrastructure. [Here's how][2].

[1]: https://docs.travis-ci.com/user/workers/container-based-infrastructure
[2]: https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Container-Based-Infrastructure

waylan:
Regarding `<http://127.0.0.1:8000/>`, my suggestion would be to removed the angle brackets (`http://127.0.0.1:8000/`), which would make it no longer be a link, just plain text... which I think is correct, as it is not always a valid link anyway. Personally, I've always been annoyed when documentation links me to a not-running local server.

.travis.yml
d0ugal: The reason I switch to apt-get is that the addons stuff you tried just didn't work. I tried it a few different ways with no luck. However, I have another solution, we can just install linkchecker from PyPI!

d0ugal:
I had to change `<http://127.0.0.1:8000/>` to `'http://127.0.0.1:8000/'` (except with backticks, I can't figure out how to include the backticks in an inline example!) otherwise the markdown linter complained - it is hard to keep all the linters happy :wink:

waylan:
The travis config forces the use of Python3.5 as the default python to call pip, tox, etc. This is required to force an install of Python 3.5, otherwise the PY35 test fails with a message that Python 3.5 is not installed. Unless we can force Python 3.5 to install some other way (or Travis has resolved the issue) , we will need to install form tox and force the `linkchecker` env to use Python 2.7 as LinkChecker does not support Python 3.x.

Annoyingly, the tox test will then fail on Windows as LinkChecker contains a C lib and can't be installed via pip on windows (you need the precompiled `.exe` file).

d0ugal:
> Annoyingly, the tox test will then fail on Windows as LinkChecker contains a C lib and can't be installed via pip on windows (you need the precompiled .exe file).

I think everything else is fine, we can handle the Python 3 situation with tox like we do with the flake8 test.

Most users are likely to see an error, I think it is unlikely that most developers (particularly on windows) will have 2.6, 2.7, 3.3, 3.4 and 3.5 installed :smile:. So a failing linter hopefully isn't an issue, I just hope the error isn't too confusing. If it is, we can always remove them from the default envlist later.

waylan:
Looks like you got it working. A couple things:

We may want to add the `--check-extern` option to the `linkchecker` command so that external links are checked. Great for ensuring external links don't go stale.

And while looking at the options, I see that there is support for a `--ignore-url` option. Presumably, we could tell it to ignore `http://127.0.0.1:8000/` if we really wanted a link (I don't, just mentioning it for completeness).

d0ugal:
> We may want to add the --check-extern option to the linkchecker command so that external links are checked. Great for ensuring external links don't go stale.

Yup, just added that back in. I removed it as the version I did get from apt-get didn't have it.

d0ugal:
I just tidied up the git history, it was bugging me. I think everything is done now, so just waiting for CI to catch up. Appveyor is slow and miles behind, but I am in no rush thankfully :smile:

d0ugal:
@waylan It has started to fail now :( https://travis-ci.org/mkdocs/mkdocs/jobs/126401874

I think I'll need to remove the external check.

waylan:
> I think I'll need to remove the external check.

So to would appear. I took another look at the docs and don't see any workaround for failing SSL certificates. Although we could exclude those specific URLs with the `--ignore-url` option. I would think `--ignore-url="^https.*google.*" would do the trick.

d0ugal:
I tried a couple of regexes but kept getting it wrong, I got bored of the slow feedback loop as I don't get the error locally - so I just removed it for now, it would be good to add at some point.

waylan:
I can live with that. We know that all links are valid today. We can always address this again in the future when an external link changes...

PR #914, opened 2016-04-28
Add a config file for CodeCov

d0ugal:
CodeCov now requires that we customise it with a config file (ala Travis and
AppVeyor.). This is just the default config.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    patch: false
+    project:
+      default: {}


d0ugal:
I planned to add something to this config to disable the comments, but it doesn't actually seem to be an option. I think we can just customise what is included in the comment.

waylan:
@d0ugal, it looks like CodeCov has added the ability to disable comments now (see [here][1]).  Not sure it you still want to do that. All the options are listed [here][2].

[1]: https://github.com/codecov/support/issues/162#issuecomment-217634710
[2]: https://github.com/codecov/support/wiki/codecov.yml#variable-descriptions

d0ugal:
Neat, thanks. Since CodeCov only posts one comment and then edits and updates it when a the pull request is updated it is less annoying that I had feared. So, I might look into the comment formatting to see if we can make it more useful - at the moment I don't feel like it adds much value. Failing that, I'll remove it.

Issue #900, opened 2016-04-14
per page customization


anantshri:
Example there is a specific page where i want the ToC to either be not present or collapsible. i am not a ui specialist and hence a simpler solution i was looking at was some sort of options in per md page where i can say something like toc=no and toc doesn't gets added to it.

So is it possible or i will need to spend time in getting a new theme worked out for it.

waylan:
Page meta-data is passed to the template under the `meta` variable. You would just need to provide a template which made use of it. While overriding templates is currently possible, more work needs to be done both to make it easier and to document what is possible. See #607 for the details.

I suspect anything outside of that would need to wait for the as-yet unimplemented Plugin API (see #206).

anantshri:
Thanks for the pointer will check them out.
Also is there a way to remove one of the default plugin's in mkdocs (meta, toc, tables, and fenced_code).

I am most probably trying to find a way with minimal tweaks in core mkdocs stuff or files coz with auto builds that basically means more tweaking and testing of build environment and more chances of failure.

d0ugal:
> Also is there a way to remove one of the default plugin's in mkdocs (meta, toc, tables, and fenced_code).

No, not at the moment. Some of these are required for MkDocs to work, I have not experimented with disabling any. Are they causing a problem?

anantshri:
@d0ugal toc is something i wanted to remove for a specific project of mine. but i think i will end up creating a custom theme for it.

d0ugal:
Yup, makes sense. I think the toc extension is the one we rely on the most :smile:

anantshri:
i think this simple pull request should add this feature in default set and yet doesn't impact much https://github.com/mkdocs/mkdocs/pull/917

Issue #700, opened 2015-07-30
Disabling the ToC on a specific page


pmackay:
Is there any way to disable or hide the ToC on a given page? I cannot see a way to define a custom theme layout for that could be used for just one or a few pages, rather than all pages, which would be another way to do it. Is that possible at all?

robodude666:
You should be able to define a `hide_toc` attribute in the metadata of the markdown file you don't want the ToC generated for. Then, modify the template to not render the ToC if `meta.hide_toc` is true.

d0ugal:
Yup, good idea. That would do it. Otherwise there isn't a way to do this in mkdocs - theme customisation is required.

PR #917, opened 2016-04-29
added support for removing ToC from specific file

anantshri:
This theme update will allow user to define a variable ```no_toc``` to disable ToC from that specific page.

This should be useful for scenario's like https://github.com/mkdocs/mkdocs/issues/700 or https://github.com/mkdocs/mkdocs/issues/900

### DIFF
--- a/mkdocs/themes/mkdocs/base.html
+++ b/mkdocs/themes/mkdocs/base.html
@@ -44,9 +44,13 @@
         {% include "nav.html" %}
 
         <div class="container">
-            {% block content %}
-                <div class="col-md-3">{% include "toc.html" %}</div>
-                <div class="col-md-9" role="main">{% include "content.html" %}</div>
+	{% block content %}
+		{% if meta.no_toc %}
+			<div class="col-md-12" role="main">{% include "content.html" %}</div>
+ 		{% else %}
+			<div class="col-md-3">{% include "toc.html" %}</div>
+			<div class="col-md-9" role="main">{% include "content.html" %}</div>
+		{% endif %}
             {% endblock %}
         </div>
 


waylan:
My suggestion would be to reject this in favor of #607, i.e., improvements should be made to the theme so that a user could override a specific portion of the template to match their desired output/behavior.

anantshri:
@waylan I would have also suggested the same if there was anything happening on that specific issue. as there is no work on that specific issue and i needed and found couple of other folks asking for this 
specific feature hence i added it to the default theme so that 

1) its available in default theme
2) People have a clear idea how to implement it in case they want it in custom themes.

waylan:
To me, this feature to too specialized and not something that will be wanted/needed across a larger user base. I'm suggesting that rather than MkDocs offering this very specialized feature, MkDocs should make it easier for a user to implement this behavior (or various other similar behaviors) however they want.

While it is true that work on #607 is not moving very quickly, the reason is that we are working on MkDocs in our spare time. That time is already spread thin. If we add more specialized features that we don't need/use, that is one more maintenance headache that will sap our time and take even more time away from working on the features that are actually useful to the larger user base. Besides, if we accept this specialized feature, then we should also accept all the other specialized feature requests. In the end, that time would be better spent making it possible for users to implement there own specialized behavior.

d0ugal:
I agree with @waylan. Adding features like this in MkDocs itself isn't sustainable, it means we would need to document them and support them across all themes and encourage all external themes to support them too. Just supporting it in one specific place isn't generally useful.

If you really want to go down this path, you could fork the theme and customise it as you wish.

anantshri:
If anyone was following this i have forked the default mkdocs theme https://github.com/anantshri/mkdocs-reloaded and i am working on adding these features which i wanted to add on theme.

Himura2la:
Seems like I still can't remove ToC without touching the theme, right?

waylan:
@Himura2la that is correct. This will likely be addressed in the next major release, but not before.

mprimi:
@waylan This issue is still top search result for "disabling TOC for a page". Was this feature implemented?
I don't see it mentioned in the documentation.

waylan:
@mprimi theme customization is documented [here][1]. You probably want to use the [custom_dir][2] and [template blocks][3] for this sort of customization.

If you want this change to only affect some pages, then you could create a separate template (rather than replacing `main.html`) and also make use of the `template` [meta-data][4] variable to point to that template from a specific page.

[1]: https://www.mkdocs.org/user-guide/styling-your-docs/#customizing-a-theme
[2]: https://www.mkdocs.org/user-guide/styling-your-docs/#using-the-theme-custom_dir
[3]: https://www.mkdocs.org/user-guide/styling-your-docs/#overriding-template-blocks
[4]: https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data

waylan:
To accomplish this with the `mkdocs` theme, create your `custom_dir` and add a file `no_toc.html` which contains the following:

```
{% extends "base.html" %}

{% block content %}
	<div class="col-md-12" role="main">{% include "content.html" %}</div>
{% endblock %}
```

Then, in your Markdown pages add this front matter:

```markdown
---
template: no_toc.html
---

# Page content here...
```

So long as you have the [theme.custom_dir][1] config option properly set, any page which contains that front matter will not have a TOC, and any page without the `template` variable set will use the default template `main.html` and have a TOC. 

If you want the changed behavior to be the default, then rename the template from `no_toc.html` to `main.html` and remove the `template` variable from the front matter of any pages.

Of course, the same method could be used for other themes. However, the content of the template would need to be adjusted to fit within the structure of that theme.

[1]: https://www.mkdocs.org/user-guide/configuration/#custom_dir

mprimi:
Thank you @waylan, that worked like a charm!

Future gogglers. In case the above explanation is not clear, here's how followed the suggestion to remove the TOC from **ALL** pages: https://github.com/mprimi/notes/commit/e7057d91723bf9e30a7fc1e8ee99040da94f0276

Issue #918, opened 2016-04-29
build docs from setup.py 


williamjmorenor:
Hi 

I was looking at the setup.py file of https://github.com/mfenniak/pg8000 and it uses a setup_command to expose a  python setup.py build_sphinx options to setup.

Looking at sphinx source code for setup_command (0) look like mkdocs can implement a similar oproach to let users builds the documentation with a setup.py options (something like python setup.py mkdocs).

I am not sure if this feature interests to mkdocs core team, but maybe I can take a time to work in a setup_command implementation for mkdocs, even try to get it for the 0.16.0 release.

Regards

(0): https://github.com/sphinx-doc/sphinx/blob/master/sphinx/setup_command.py

waylan:
Whether we add it to MkDocs proper or not, I'm sure it would be helpful to others. I see that Sphinx [registers][1] their `build_sphinx` command with the `distutils.commands` entry point. That makes it easy for people to use it. I would be interested in a similar `build_mkdocs` command if was not already written. 

[1]: https://github.com/sphinx-doc/sphinx/blob/master/setup.py#L209

waylan:
As a reminder, review the comments in #919 when working on this.

Although, I am questioning the value of this. Initially I noted that this could be useful. Presumably, a developer could document their Python project with MkDocs and add a line to their `setup.py` script to add a `build_mkdocs` command for their users to build the documentation to their project. While that sounds like a nice feature (using standard Python tools, etc), the command would fail if the user didn't have the dependencies installed (namely MkDocs). Sure, a nice error message could be issued telling the user to install the dependencies, but that still seems less than ideal to me. Note that the same problem exists with the cited example for building documentation with Sphinx. It only works if the user has the dependencies installed. Sure, a properly configured `setup.py` script will auto-install dependencies for the package, but that would mean the developer would need to list MkDocs as a dependency and the dependencies would need to be installed before the build script is run (I don't recall off-hand which order things are done s that may or may not work).

What I'm sure about is that getting the user to install the dependencies manually is a larger hurdle than getting them to run the `mkdocs build` command directly.

oprypin:
`setup.py` is dead :slightly_smiling_face:  :tada: 
* #2988

PR #919, opened 2016-04-30
Some advance with setup_command

williamjmorenor:
Signed-off-by: William Moreno Reyes <williamjmorenor@gmail.com>

### DIFF
--- a/mkdocs/setup_command.py
+++ b/mkdocs/setup_command.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+"""
+    Setuptools/distutils commands to assist the building of MkDocs
+    documentation from python setup.py
+"""
+
+from distutils.cmd import Command
+
+from mkdocs.__main__ import build_command
+
+
+class BuildDoc(Command):
+    """
+    A custom command to build MkDocs documentation from setup.py
+    """
+
+    description = "Build MkDocs documentation"
+    user_options = []
+    boolean_options = []
+    
+    def initialize_options(self):
+        pass
+    
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        Build = build_command
+        Build()
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,10 @@ if sys.argv[-1] == 'publish':
     print("  git push --tags")
     sys.exit()
 
+try:
+    from mkdocs.setup_command import BuildDoc
+except ImportError:
+    pass
 
 setup(
     name="mkdocs",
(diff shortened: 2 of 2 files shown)


williamjmorenor:
Hi I have been doing some testing with the setup_command (https://github.com/mkdocs/mkdocs/issues/918).

The comand seems to work, I can run:

>>> setup_command.BuildDoc.run()

The build_mkdocs is visible with python setup.py --help-commands

But when I run the command I got:

python3 setup.py build_mkdocs 
running build_mkdocs
Usage: setup.py [OPTIONS]

Error: Got unexpected extra argument (build_mkdocs)

I am doing something wrong with the run() module but I can not find what is.

waylan:
I'm just guessing here, but you are using the Click `build` command (from `__main__.py`). Click gets its arguments from reading argv and argv contains the argument `build_mkdocs` which is unknown to Click.  In my experience,  nesting one cli system inside another will almost always break without some extra steps. I would suggest searching for how to use Click with disutils commands.

waylan:
Actually, if you look at the Spinx example you pointed to in #918, you will see that they re-implement all of the options directly using disutils command tools. It may be that that is the only way it will work.

d0ugal:
I played around with this and tried to make it work. The issue you are hitting is the conflict between different command line frameworks. Click looks at the full input path `python setup.py build_mkdocs` command and spots `build_mkdocs` as an argument, but that isn't one that we have registered in the Click build command.

I think @waylan is correct, the easiest way would be to just allow build be called with no optional arguments. So, just like calling `mkdocs build` but don't support `mkdocs build --theme...` etc. I think that would be the best option.

Otherwise you would need to duplicate all the command line options and we would need to maintain them both. I don't want to do that.

waylan:
Click has the [CLIRunner][1] Class, for [testing][2] your commands. It allows you to pragmatically call a command, passing in options and args isolated from the current process's `argv`. Perhaps the `setup_command` could gather the options and args and pass them on to Click this way (perhaps the `--help` flag could even be passed through?). The trick would be getting `setup_command` to accept any arbitrary options and args without them being redefined. If that's possible, this should be an easy fix. If not, then I wonder if there is even any value in bothering to implement a basic, option-less, command.

[1]: http://click.pocoo.org/5/api/#click.testing.CliRunner
[2]: http://click.pocoo.org/5/testing/#basic-testing

waylan:
I'm closing this as we don't need both the issue and PR open. Someone can always open a new PR.

PR #921, opened 2016-05-01
Refactor page specific template variables.

### LINKED ISSUE #874: Page_Title is "None" on root document
`{{ page_title }}` is displaying as "None" despite my mkdocs.yml configuration as such:

    pages:
    - Usage: 'index.md'
    - Installing: 'Installing.md'

I just updated to version 0.15.3. This wasn't a problem with whatever version I had before. This only happens to the index page.

waylan:
A deprecation warning is issued for all old variables and all new
page specific variables are attributes of the 'page' object.

Global variables are uneffected, except page_description.

See the changes described in the release notes for details.

Fixes #874.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -15,6 +15,46 @@ You can determine your currently installed version using `mkdocs --version`:
 
 ## Version 0.16 (2016-02-??)
 
+### Major Additions to Version 0.16.0
+
+#### Template variables refactored. (#874)
+
+Page specific variable names in the template context have been refactored as
+defined in [Custom Themes](../user-guide/custom-themes/#page). The
+old variable names will issue a warning but continue to work for version 0.16,
+but may be removed in a future version.
+
+No global variables were altered except `page_description`. Previously
+its value was altered programicaly based on whether the current
+page was the homepage. Now it simply maps to `config['site_description']`.
+Use `page.is_homepage` in the template to conditionally change the
+description.
+
+Any of the following old variables should be updated to the new ones in user
+created and third-party templates:
+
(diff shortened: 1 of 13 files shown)


docs/about/release-notes.md
```diff
@@ -15,6 +15,46 @@ You can determine your currently installed version using `mkdocs --version`:
...
+#### Template variables refactored. (#874)
+
+Page specific variable names in the template context have been refactored as
+defined in [Custom Themes](../../user-guide/custom-themes/#page). The
+old variable names will issue a warning but continue to work for version 0.16,
+but may be removed in a future version.
+
+No global variables were altered except `page_description`. Previously
+its value was altered programicaly based on whether the current 
+page was the homepage. Now it simply maps to `config['site_description']`.
```
d0ugal: I think we should do the same thing with the `config`. We pass the config to the templates and a bunch of values from the config. So like you did with the others `copyright` would become `config.copyright` etc. Then we can just deprecate `page_description` and have people use `config.site_description`.

d0ugal:
This looks great, I really like this approach.

docs/about/release-notes.md
waylan: I agree. Just trying to do this one piece at a time.

waylan:
It was late and I was tired when I pushed this last night. I'll address the errors later today when I have some time.

d0ugal:
Sounds good. No rush. I'm happy for the config variables to be handled in a follow up (or here, I don't mind).

d0ugal:
Looks good. I'm happy for this to land, but I'll let you do it as I'm not totally sure you didn't have any extra changes planned.

PR #922, opened 2016-05-01
Limit requests version for LinkChecker

d0ugal:
We seem to get getting some Linkchecker failures in #921


d0ugal:
Aha, we are hitting. https://github.com/wummel/linkchecker/issues/649

PR #927, opened 2016-05-04
Ensure tests use actual package location. Fixes #923.

### LINKED ISSUE #923: tests assume package will be unpacked into "mkdocs" directory
The following tests assume that mkdocs will be unpacked into a directory called "mkdocs". This is often not the case, e.g. the source downloaded from github and PyPI calls it ``mkdocs-0.15.3``.

https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/config/config_options_tests.py#L182
https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/config/config_options_tests.py#L182

waylan:
I think this will resolve the issue. Not able to test conclusively locally as I refuse to use deb packages for Python libs. @brianmay, can you confirm this solves the problem raised in #923?

### DIFF
--- a/mkdocs/tests/config/config_options_tests.py
+++ b/mkdocs/tests/config/config_options_tests.py
@@ -4,6 +4,7 @@ import os
 import tempfile
 import unittest
 
+import mkdocs
 from mkdocs import utils
 from mkdocs.config import config_options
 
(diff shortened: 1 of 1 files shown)


d0ugal:
Thanks! 🎉

PR #934, opened 2016-05-10
Adding PubStorm deployment instructions

ajhit406:
We also added an MkDocs template to [Nitrous](https://www.nitrous.io). Great for those who don't want to install python locally.

### DIFF
--- a/docs/user-guide/deploying-your-docs.md
+++ b/docs/user-guide/deploying-your-docs.md
@@ -94,6 +94,14 @@ There are a few prerequisites for the above to work:
 [docs_dir]: configuration.md#docs_dir
 [site_dir]: configuration.md#site_dir
 
+## PubStorm
+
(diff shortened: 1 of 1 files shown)


d0ugal:
Thanks for writing this. Unfortunately we can't add further instructions that are not used by one of the MkDocs developers as there is no way for us to ensure they are maintained and kept relevant. The three we document are included for specific reasons - we have close integration in the code with github pages, readthedocs is a *very* popular website that integrates mkdocs and PyPI is the other option used commonly in the Python ecosystem.

It might be that we need to add another wiki page with third party resources and link to that from the documentation, like we do with third party themes.

waylan:
+1 for the wiki page

ajhit406:
OK thanks for the response @d0ugal -- I think the MkDocs community could manage to keep the repo current (have seen this with other repos), but I understand. 

We'll look into adding a third party deployment services page. 

All the best,

PR #939, opened 2016-05-16
Support SOURCE_DATE_EPOCH environment variable for "reproducible" builds.

### LINKED ISSUE #938: Please support SOURCE_DATE_EPOCH specification for build time stamps
Copied from http://bugs.debian.org/824266:

> Hi,
> 
> mkdocs integrates build time stamps into the documentation it is
> generating. This makes at least unburden-home-dir no more reproducibly
> building because it now uses mkdocs to generate HTML documentation at
> build time.
> 
> See https://tests.reproducible-builds.org/rb-pkg/unstable/amd64/unburden-home-dir.html and
> https://tests.reproducible-builds.org/dbd/unstable/amd64/unburden-home-dir_0.4.diffoscope.html
> 
> Other cases of this issue are the source package djangorestframework (see
> https://tests.reproducible-builds.org/rb-pkg/testing/i386/djangorestframework.html and
> https://tests.reproducible-builds.org/dbd/testing/i386/djangorestframework_3.3.2-2.diffoscope.html)
> and of course the source package python-mkdocs itself, see
> https://tests.reproducible-builds.org/rb-pkg/unstable/amd64/python-mkdocs.html and
> https://tests.reproducible-builds.org/dbd/unstable/amd64/python-mkdocs_0.15.3-3.diffoscope.html
> 
> At https://reproducible-builds.org/specs/source-date-epoch/ you can find
> a bound to be widely adopted specification for how to solve such kind of
> issues generally: If the environment variable SOURCE_DATE_EPOCH exists
> use the Unix time stamp (seconds since 1st of January 1970, 00:00:00
> UTC) it contains to preseed all build time stamps with that time stamp
> instead of the current time.
> 
> See https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal#Examples
> for some examples on how to implement support for SOURCE_DATE_EPOCH,
> including an example for Python.

waylan:
Fixes #938.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -73,6 +73,8 @@ created and third-party templates:
   pages config. (#728)
 * Update searching to Lunr 0.7, which comes with some performance enhancements
   for larger documents (#859)
+* Bugfix: Support SOURCE_DATE_EPOCH environment variable for "reproducible"
+  builds (#938)
 
 ## Version 0.15.3 (2016-02-18)
 
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -2,6 +2,7 @@
 
 from __future__ import unicode_literals
 from datetime import datetime
+from calendar import timegm
 import io
 import logging
 import os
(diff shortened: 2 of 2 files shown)


d0ugal:
Thanks!

Issue #884, opened 2016-04-05
Better way to perform pre-processing on source md files / Support a custom directive "including" a path-specified file


ronjouch:
Hi.

I'm working to improve the accessibility of a projects high-level doc (as opposed to docstrings-based per-class/method doc). It's already markdown-based, and I picked mkdocs because it does the job well. So as expected, I'm ending up with a `docs` folder at the root of the project, with a dozen `.md` files.

However, as part of *"improving the accessibility of high-level doc"*, we had a goal of bringing the doc as close as possible to the code. So in cases where it makes sense, we want high-level doc to be included directly in the code (c++ in our case) as comments, to minimize eventual doc/code de-synchronization: when the documentation is in your face *in the code*, it's easier to remember updating it when changing code.

To my knowledge this is not possible with mkdocs, so I ended up with a simple hackish pre-processing Python script that reads `mkdocs.yml` and, for each `.md` referenced in the `pages` section, looks for something like

```
!md_include path/to/some.cpp
```

... opens `path/to/some.cpp`, then finds

```
/** md
Awesome markdown-based doc on `some.cpp`.
*/
```

... and replaces the `!md_include` with the contents delimited by the special `/** md` ... `*/` multi-line comment. Pre-processed files are dumped to a temporary folder, from where the script then calls `mkdocs build` and calls it a day. This works well for us, except it's hackish and breaks `mkdocs serve`.

→ Do you see a better way of reaching the same result with the current mkdocs?  
→ If not, is there interest to add support for such custom pre-processing?

waylan:
May I suggest taking a look at #777.

If that doesn't meet your needs, then I suspect this would need to wait for the Plugin API (see #206).

Either way, this would be a duplicate.

ronjouch:
#777 doesn't cover my use case as I need to "extract" markdown content from files, not just include the whole files. Cool to see a plugin API is being thought of. Thanks for the fast answer @waylan :)

facelessuser:
@ronjouch You could adapt the markdown-include extension listed in #777 by modifying it to suite your need.  You could have it include your source file(s) and then have the extension filter out the non relevant text leaving you only with your markdown source comments before inserting it into the body of the markdown file.  If that is still not doable, as @waylan said, you would need to wait for the Plugin API listed in #206.

waylan:
[...]

Issue #641, opened 2015-06-18
How to build API reference docs


twiecki:
Along with documentation, sphinx can auto-generate references by walking through the files and looking at the doc-strings. Is this possible or are there any plans to add this?

d0ugal:
This isn't currently possible and we have no plans to add it to MkDocs itself. I would like to see this happen via the extension API when we add one. See #206.

d0ugal:
FWIW, API docs (and a general plugin API) is something that I really want in some cases. So it is bubbling up in priority for me but I have a few critical things I need to take care of first. So I have no idea when I would have time to take it on.

waylan:
It is interesting to note that some people consider it bad form to generate documentation from code. For example, [this states][1] (emphasis in original):

> If your documentation is generated from source code, I am immediately skeptical. You should use words to communicate with your users, and those words shouldn’t live in your source code. **If you included all of the things needed to document a project in source, your code would be unreadable.**

> So please, use a tool that allows you to write prose documentation outside of your source code. Your users will thank you.

I realize that that is an opinion, which you are entitled to disagree with. And I expect this feature will continue to be requested until it is made available, regardless of how it is implemented (as a plugin, a Markdown extension, or built-in). But it is an opinion that many people share. In fact, that post links to a few supporting resources which, in turn, link to more people/projects who share the same opinion.

[1]: http://ericholscher.com/blog/2014/feb/27/how-i-judge-documentation-quality/#prose

d0ugal:
Yeah, I almost totally agree. They are not a replacement to well written docs. So if all your docs are generated from code, then I think you are doing it wrong :)

However, I do like them as an additional resource to prose documentation. Personally, for example, I find them particularly useful when I know a project well but I just want to look up something I've not used in a while.

waylan:
Can't say I disagree with you @d0ugal. Just thought that is might be good to have that stated somewhere to point people to if/when the need arises. If you read my post carefully, I indicate no opinion of my own.

d0ugal:
Yup, I thought that was the case :)

kernc:
User's manual is one thing, and API reference (docstrings) is another. A minimally usable documentation tool should be able to produce a HTML of the latter. Period.

d0ugal:
> User's manual is one thing, and API reference (docstrings) is another. A minimally usable documentation tool should be able to produce a HTML of the latter. Period.

[...]

Issue #357, opened 2015-02-25
Feature request: specify pre-build script


twiecki:
I have a script that compiles my IPython NBs to markdown which are then picked up by mkdocs. It'd be nice if I could specify this script to be run on every `mkdocs build`. Something like `pre-build-script: convert_nbs.sh`.

d0ugal:
That could be kinda nice, I guess what you really want is the plugin system see #206 (this probably wont be landing particularly soon). However, other than that, I don't see value in specifying this. Why not just add a `build.sh` file with something like:

    #!/usr/bin/env bash
    convert_nbs.sh
    mkdocs build

AdrianSosic:
Is there nowadays a simple off-the-shelf solution available for this that does not require writing custom plugins?

pawamoy:
[Hooks](https://www.mkdocs.org/user-guide/configuration/#hooks) will let you do that without having to create an installable package 🙂

AdrianSosic:
@pawamoy: Thanks, you just made my day 🌻 For some reason, I just couldn't find it myself ...

PR #942, opened 2016-05-23
Adds the ability to include yaml files in each other

aeslaughter:


d0ugal:
This functionality is provided by python-markdown extensions. For example: https://pypi.python.org/pypi/markdown-include

I don't think this belongs in MkDocs itself.

waylan:
No, this allows the YAML config file to "include" sub-YAML files. This has nothing to do with Markdown.

Regardless, it is not clear to me what the use-case would be for this feature. In some projects, there may be multiple configs for development/testing/production, but MkDocs is a pretty simple project and shouldn't need multiple configs. Besides, being a static site generator, there is no distinction between development and production. Therefore, I agree this should be closed, albeit for a different reason than @d0ugal stated.

aeslaughter:
The use case for this is that I have a basic configuration that does not change and will be checked into our project repository (https://github.com/idaholab/moose). However, the pages that that will be displayed will be generated on-the-fly as a part of our continuous integration (i.e., the change will not merge unless the website builds correctly).

mkdocs.yml
```
theme: readthedocs
pages: 
    - Home: index.md
    - Getting Started: getting_started.md
    - Systems: !include Systems.yml # this will be generated with CI
```

waylan:
Similar requests have been made in #641, #357, and #884 (and possibly others - those are the ones I found in a quick search). In each case, the request was denied in favor of a future Plugin API (see #206). I suspect this would fall into the same category.

Additionally, when it was pointed out that YAML already supports defining Python types [here][1], the [reaction][2] was that MkDocs should disable that default behavior of YAML as it is ugly. In this case, you are introducing a similar non-default syntax to YAML. If there is a non-YAML hacking way to do this, that would be the preferred method.

[1]: https://github.com/mkdocs/mkdocs/issues/212#issuecomment-115404162
[2]: https://github.com/mkdocs/mkdocs/issues/212#issuecomment-115573329

aeslaughter:
Thanks for the info, we will probably just use a fork of MkDocs until there is a way to plugin to the MkDocs API.

d0ugal:
Thanks for the clarification @waylan. Luckily my conclusion is the same :smile:

PR #943, opened 2016-05-23
Adds ability to add to the watching directories when using serve

aeslaughter:


mkdocs/commands/serve.py
```diff
@@ -24,12 +24,15 @@ def _livereload(host, port, config, builder, site_dir):
...
+    for d in config['watch_dirs']:
+        server.watch(d, builder)
+
     server.serve(root=site_dir, host=host, port=int(port), restart_delay=0)
 
 
 def _static_server(host, port, site_dir):
 
-    # Importing here to seperate the code paths from the --livereload
+    # Importing here to separate the code paths from the --livereload
```
d0ugal: Good spot!

d0ugal:
Like #942, sorry, but I don't think this belongs in MkDocs. This seems like a very specific use-case. We are trying to reduce the number of configuration options, not add more for the moment.

waylan:
I agree. Perhaps if/when we add a Plugin API, that could be used for something like this. But it doesn't belong in MkDocs proper.

waylan:
A recent [query][1] on the mailing list got me thinking about this. If you want to "include" some markdown files in other files (using the available Markdown extension), where do you store those "included" files? If you put them in the `docs_dir` (or a sub-directory), then those files will be included as pages of their own and in the site navigation, which may not be what is wanted. Of course, the `pages` config option could be populated manually, excluding those "included" pages, but that may be less than ideal in a large site. The natural next step would be to keep the files outside of the `docs_dir` which this feature would have allowed/supported. Of course, users still can, but they lose the server's auto-reload feature.

I get the sense that "including" sub-pages is a pretty popular feature among users and I expect we will continue to bump into this over time. At this time is appears that the policy is that users should place the "included" files in the `docs_dir`, and populate the `pages` config option manually, excluding the "included" pages. @d0ugal if you have a different take on this (or if I'm missing something), any feedback would be appreciated.

[1]: https://groups.google.com/forum/#!topic/mkdocs/hrj00YhKqdA

aeslaughter:
Just for reference here is what I am doing. MOOSE is a framework comprised of a set of c++ objects that do certain items related to solving a set of equations (e.g., Diffusion operator). Next to the header (e.g., include/Diffusion.h) for one of these pieces of code there is a markdown file (incude/Diffusion.md) that will include a detailed description (paragraph to pages depending on the physics).

Then using a custom markdown generator another markdown file is auto generated (see example below) in the 'docs_dir' (e.g., docs/Diffusion.md), this markdown file includes the aforementioned file living in the code with all sorts of other information about the code (input file syntax for using the class, a list of uses within the examples and tests, etc.). 

Therefore, when application developers are creating new c++ objects (something that is very common in our framework) and they want to document the class (a requirement for getting their physics into the repository) they will be creating/editing these markdown files outside of the 'docs_dir'. So, this PR simply enabled the auto update to work when editing these outside files, which is on of the major reasons we selected MkDocs to implement this project.

**Example Auto Generated Markdown**
``` 
Diffusion
The Laplacian operator.

{!/include/Diffusion.md!}

Input Parameters
...

Examples
....

Tests
...

Common Links:
* Github
* Doxygen
```

Issue #946, opened 2016-05-24
Remove prev next from navigation


huegelc:
Hi,

how can I remove prev/next items from navigation? Is there a option in `mkdocs.yml` ?

waylan:
Have a l look at #633 in which the `include_next_prev` config option is being deprecated (which explains why it is not documented). As noted in the last comment on that issue, users should remove the next/prev links via theme customization as discussed in  #607.

gringolalia:
Could you provide a quick step-by-step? I do not understand! Where are the theme files and templates located and how can I edit them? Newbie sorry.

waylan:
@gringolalia you don't "edit" the templates, but you override them either through replacement files or through template inheritance. However, the mechanisms for doing that are not fully implemented in the current release (0.15), but will be in the next release (0.16). That release will also include extensive documentation about how to override various parts of the templates. For a preview, you can view the pending updates to the documentation here: https://github.com/mkdocs/mkdocs/blob/master/docs/user-guide/styling-your-docs.md#customizing-a-theme. Note that "Template Blocks" are completely missing from the version 0.15 (they hadn't been implemented yet) and some of the other stuff has been tweaked a little since that release. In other words, those are the docs for the new upcoming features. Things currently don't work exactly that way (mostly but not quite), but its the best we have. Hopefully the next release will happen soon and everything will work as documented there.

bretonio:
Looking forward to it thx!

oprypin:
The current link is https://www.mkdocs.org/user-guide/customizing-your-theme/#overriding-template-blocks and indeed it works as described: you can add an *empty* `{% block next_prev %}{% endblock %}` to override the existing one, both for `readthedocs` theme and `mkdocs` theme:
https://github.com/mkdocs/mkdocs/blob/e1b77ab66e8305b5f0373b4842b50d2618390a96/mkdocs/themes/readthedocs/footer.html#L2
https://github.com/mkdocs/mkdocs/blob/e1b77ab66e8305b5f0373b4842b50d2618390a96/mkdocs/themes/mkdocs/base.html#L113

oprypin:
OK so... you're doing *something* (we don't know what) and you're getting errors.

If you can't figure those out, perhaps go to https://github.com/mkdocs/mkdocs/discussions and open a new thread **with full context**: MkDocs version, what you did, what you're seeing (you provided only the latter)

oprypin:
[...]

PR #947, opened 2016-05-25
Support theme customization with blocks

waylan:
Partially addresses #607.

Note that we have to rename `base.html` to `main.html` and then create a new `base.html` which is an empty file that "extends" `main.html `. We could reverse the roles of those two files, but that would be backward incompatible with existing custom/third party themes as all themes are loaded from `base.html`. Existing custom themes may not have a `main.html` and switching the role of the two files, despite being more sensible,  could break existing custom and third party themes. Of course, if we want to make the backward incompatible change, now is the time to do it. Let me know and I'll update this PR.

Also note that this provides a way for users to override the behavior of next/prev links when `include_next_prev` is deprecated (see #633, #679,  #681, & #946), use a different analytics service, override search, and various other common requests -- all without any new config settings or Plugin API.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -65,7 +65,7 @@ behavior of the navigation, among other things. See the relevant
 To enable this feature, the primary entry point for page templates has been
 changed from `base.html` to `main.html`. This allows `base.html` to continue to
 exist while allowing users to override `main.html` and extend `base.html`. For
-version 1.16, `base.html` will continue to work if no `main.html` template
+version 0.16, `base.html` will continue to work if no `main.html` template
 exists, but it is deprecated and will raise a warning. In version 1.0, a build
 will fail if no `main.html` template exists. Any custom and third party
 templates should be updated accordingly.
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -53,6 +53,37 @@ created and third-party templates:
 [page.previous_page]: ../user-guide/custom-themes/#pageprevious_page
 [page.next_page]: ../user-guide/custom-themes/#pagenext_page
 
+#### Increased Template Customization. (#607)
+
+The built-in themes have been updated by having each of their many parts wrapped
+in template blocks which allow each individual block to be easily overriden
+using the `theme_dir` config setting. Without any new settings, you can use a
+differant analytics service, replace the default search function, or alter the
+behavior of the navigation, among other things. See the relevant
+[documentation][blocks] for more details.
+
+To enable this feature, the primary entry point for page templates has been
+changed from `base.html` to `main.html`. This allows `base.html` to continue to
+exist while allowing users to override `main.html` and extend `base.html`. For
+version 1.16, `base.html` will continue to work if no `main.html` template
+exists, but it is deprecated and will raise a warning. In version 1.0, a build
+will fail if no `main.html` template exists. Any custom and third party
+templates should be updated accordingly.
+
+The easiest way for a third party theme to be updated would be to simply add a
+`main.html` file which only contains the following line:
+
+```django
+{% extends "base.html" %}
+```
+
+That way, the theme contains the `main.html` entry point, and also supports
+overriding blocks in the same manner as the built-in themes. Third party themes
+are encouraged to wrap the various pieces of their templates in blocks in order
+to support such customization.
+
+[blocks]: ../user-guide/styling-your-docs/#overriding-template-blocks
+
 ### Other Changes and Additions to Version 0.16.0
 
 * Bugfix: Support `gh-deploy` command on Windows with Python 3 (#722)
--- a/docs/user-guide/styling-your-docs.md
+++ b/docs/user-guide/styling-your-docs.md
@@ -153,6 +153,52 @@ Your directory structure should now look like this:
     `theme`. If you want to remove files, or build a theme from scratch, then
     you should review the documentation for building a [custom theme].
 
+#### Overriding Template Blocks
+
+The built-in themes implement many of their parts inside template blocks which
+can be individually overridden in the `main.html` template. Simply create a
+`main.html` template file in your `theme_dir` and define replacement blocks
+within that file. Just make sure that the `main.html` extends `base.html`. For
+example, to alter the title of the MkDocs theme, your replacement `main.html`
+template would contain the following:
+
+```django
+{% extends "base.html" %}
+
+{% block title %}
+<title>Custom title goes here</title>
+{% endblock %}
+```
+
+In the above example, the title block defined in your custom `main.html` file
+will be used in place of the default title block defined in the parent theme.
+You may re-define as many blocks as you desire, as long as those blocks are
+defined in the parent. For example, you could replace the Google Analytics
+script with one for a different service or replace the search feature with your
+own. You will need to consult the parent theme you are using to determine what
+blocks are available to override. The MkDocs and ReadTheDocs themes provide the
+following blocks:
+
+* `site_meta`: Contains meta tags in the document head.
+* `htmltitle`: Contains the page title in the document head.
+* `styles`: Contains the link tags for stylesheets.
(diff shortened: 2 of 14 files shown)


waylan:
Also, I should note that I did not address the readthedocs theme. That could be addressed separately though as the only edit would be to wrap a few pieces of the template in blocks.

waylan:
I was able to work out a solution for the `base.html` vs. `main.html` issue. Now the user overrides the `main.html` template which "extends" `base.html` which makes more sense. And if a custom/third party theme does not contain a `main.html` template, MkDocs falls back to using `base.html`, but a warning is raised suggesting that the theme should be updated prior to version 1.0 when the fallback will be removed, which provides a clean upgrade path for custom and third party templates. And if third party templates follow our lead and use the `main.html` and `base.html` templates like we are, their updated templates should continue to work with older versions of MkDocs (which will load `base.html` directly).

If this backward incompatible change (with a clean upgrade path) is acceptable, I'll clean up the commit history so that the files aren't juggled back and forth (and so blame will still work on the lengthy `base.html` template files). If this change is not acceptable, I'll just back out the last commit. Any input @d0ugal?

d0ugal:
I've only had time to scan this very quickly - but it looks cool. I'll try and review properly today/tomorrow.

docs/about/release-notes.md
```diff
@@ -53,6 +53,37 @@ created and third-party templates:
...
+in template blocks which allow each individual block to be easily overriden
+using the `theme_dir` config setting. Without any new settings, you can use a
+differant analytics service, replace the default search function, or alter the
+behavior of the navigation, among other things. See the relevant
+[documentation][blocks] for more details.
+
+To enable this feature, the primary entry point for page templates has been
+changed from `base.html` to `main.html`. This allows `base.html` to continue to
+exist while allowing users to override `main.html` and extend `base.html`. For
+version 1.16, `base.html` will continue to work if no `main.html` template
```
d0ugal: Should be `version 0.16`

docs/user-guide/styling-your-docs.md
```diff
@@ -153,6 +153,52 @@ Your directory structure should now look like this:
     `theme`. If you want to remove files, or build a theme from scratch, then
     you should review the documentation for building a [custom theme].
 
+#### Overriding Template Blocks
```
d0ugal: It would be nice to reference this section from the `custom-themes.md` file. I would like it if we could encourage external themes to use the same block names. This may happen anyway, but would be good to call out.

mkdocs/themes/mkdocs/main.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+
+{#
+The entry point for the MkDocs Theme.
+ 
+Any custom themes which inherit from this theme should override this file to redefine
```
d0ugal: How about changing to "Any theme customisations should override this file..."

There are two subtly different use cases. External custom themes and theme customisations, I want to avoid the two being confused.

docs/user-guide/styling-your-docs.md
```diff
@@ -153,6 +153,52 @@ Your directory structure should now look like this:
...
+blocks are available to override. The MkDocs and ReadTheDocs themes provide the
+following blocks:
+
+* `site_meta`: Contains meta tags in the document head.
+* `htmltitle`: Contains the page title in the document head.
+* `styles`: Contains the link tags for stylesheets.
+* `scripts`: Contains the JavaScript libraries included in the page.
+* `analytics`: Contains the analytics script.
+* `extrahead`: An empty block in the `<head>` to insert custom tags/scripts/etc.
+* `search_modal`: Contains the search pop-up modal.
```
d0ugal: I think the search_modal is the only anomaly here. Since it makes sense in the MkDocs theme, but not the readthedocs theme. Maybe it should be limited more to the search results within the modal (to allow people to replace any text and customise that without revealing the specific details of how the theme is implemented). 

However, I am not sure that is better approach since it still work work well with the read the docs theme, so maybe we shouldn't include it at all. I'm really not sure.

d0ugal:
I generally like this approach, I think it should work well and help a bunch of users. The only bit I don't like is the config setting `theme_dir` - I realise that you only used this as it already exists. However, I think it is ugly and doesn't really make it obvious how we expect it to be used (for customisations, not full theme replacements). I don't have a good suggestion at the moment, I am going to give it some thought. However, I am open to renaming this if we can come up with a better name (maybe we can even do a convention based approach like we do with docs dir and set a default).

I see two use cases for theme_dir at the moment.

- Full theme replacements, for this, we should allow the "theme" setting to be a path and overload that. Then `theme: my_custom_theme` would be valid (we might want an explicit syntax like `file://my_custom_theme` or something simple like `local:my_custom_theme`).
- Theme customisations, which is what this PR is all about. These should be smaller tweaks for most users. We could add a new setting, `theme_customisations` would be obvious (unless you are american and expect a `z`!) but it is really long so I am not a big fan. 

Perhaps we could make the theme setting more flexible. Something like this might work:

```
theme:
    base: mkdocs
    custom: project_tweaks
```

Some of this could be done as a follow up, as I don't want to block this work but I think it is relevant to the discussion.

waylan:
> I see two use cases for theme_dir at the moment.

Right. As is stands now, if `theme` is set to a theme (MkDcos by default), then `theme_dir` is used to override the `theme` as addressed here. However, if `theme` is set to `None`, then `theme_dir` is a full theme. Documentation for that was added in 4bb0c1cb. The documentation here builds on that. I think with the two together, everything should be clear.

I'll address the other issues you raise when I have some time.

waylan:
I'm removing the `search_modal` block and in the `mkdocs` theme moving the search modal to a separate file (`search.html`) which is included in `base.html`. If a user wants to override search, they can just replace that file in their `theme_dir`. The `readthedocs` theme already has  a separate `search.html` file, so both follow a similar pattern now. The only difference is that in `readthedocs`, search is a separate page, whereas in `mkdocs`, it is included in the base page.

waylan:
I just pushed another commit with should address your concerns @d0ugal. I added some more to the docs which should help clear things up. I left the cleanup as a separate commit so you can see what I changed. With your okay, I'll rebase and squash all of the commits before merging.

waylan:
> Perhaps we could make the theme setting more flexible. Something like this might work (but again, I am not a huge fan of this idea):

>     theme:
>         base: mkdocs
>         custom: project_tweaks

Actually, I've thought for some time that  something like this would be useful. The `theme` setting could have a string value, in which case it would work as it does now. Or, the `theme` setting could have a set of nested values, in which case those settings would alter the behavior of the theme. Third party themes could potentially take advantage of this to offer features we don't want to support out-of-the-box.

> Some of this could be done as a follow up, as I don't want to block this work but I think it is relevant to the discussion.

I agree. Elsewhere we discussed creating a theme class or somesuch. I expect this would be part of that.

d0ugal:
Great, thanks! I'll try and give it another quick look tonight before merging, but if I don't manage it I am away until Monday.

d0ugal:
By the way, we can ignore the codecov failure. I think it is just complaining that out of the changed lines, the coverage decreased in `mkdocs/commands/build.py` - I'll try and find an option to turn that failure off, I don't find it particularly useful.

waylan:
I just cleaned up the history into two commits. All previous work was squashed into one commit and the [second commit][1] is my response to review. @d0ugal  when you get a change to review and if it looks okay, you could merge is as-is, or if you prefer, let me know and I'll squash the two commits. Either is fine with me.

[1]: https://github.com/mkdocs/mkdocs/pull/947/commits/37a542cdb2db60a10d7de6c5304be98d7d2683b7

docs/user-guide/styling-your-docs.md
waylan: Done.

d0ugal:
Looks good. Thanks! I think this will be a great feature for many users.

PR #948, opened 2016-05-25
Follow links when copying media files.

### LINKED ISSUE #869: allow links to be followed in copy_media_files
I want to set up a small documentation server that checks out all our projects from a local
gitlab repo and link the *docs* directories of each of them as sub folders of a separate *docs*
folder with its own index.md file and a page for each project in the mkdoc.yml like in this example:

```yaml
site_name: Our Documentation Collection
pages:
  - Home: index.md
  - project1: project1/index.md
  - project2: project2/index.md  
```

Because *project1* and *project2* are logical links to the git clone's docs directories, media files
are not copied to *site_dir* and thus e.g. images inside the projects not visible in the generated
html. 

To fix it I propose to add  ", followlinks=True" to os.walk in copy_media_files

```python
def copy_media_files(from_dir, to_dir):
    """
    Recursively copy all files except markdown and HTML into another directory.
    """
    for (source_dir, dirnames, filenames) in os.walk(from_dir, followlinks=True):
```

waylan:
Fixes #869.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -75,6 +75,7 @@ created and third-party templates:
   for larger documents (#859)
 * Bugfix: Support SOURCE_DATE_EPOCH environment variable for "reproducible"
   builds (#938)
+* Follow links when copying media files (#869).
 
 ## Version 0.15.3 (2016-02-18)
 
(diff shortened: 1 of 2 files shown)


d0ugal:
:+1:

PR #949, opened 2016-05-25
Ensure tests use actual location. Fixes #923 for real this time.

### LINKED ISSUE #923: tests assume package will be unpacked into "mkdocs" directory

waylan:


d0ugal:
Great!

PR #955, opened 2016-06-02
Avoid logging duplicate messages.

### LINKED ISSUE #954: Eliminate duplicate warnings
A few recent updates have added warnings for deprecated behavior (#921 & #947). However, the same warnings are issued once for every single page. The more pages in a project, the more warnings are issued. Each warning only needs to be printed to the console once per build. 

[This] may be a potential solution. Will explore when I have some time unless someone beats me to it.

[This]: http://stackoverflow.com/a/31953563/866026

waylan:
Fixes #954

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -16,7 +16,20 @@ from mkdocs.utils import filters
 from mkdocs.relative_path_ext import RelativePathExtension
 import mkdocs
 
+
+class DuplicateFilter(object):
+    ''' Avoid logging duplicate messages. '''
+    def __init__(self):
+        self.msgs = set()
+
+    def filter(self, record):
+        rv = record.msg not in self.msgs
+        self.msgs.add(record.msg)
+        return rv
+
+
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


d0ugal:
My only concern is that I think running 'mkdocs serve' will gradually take up more memory as it runs and the set grows.

waylan:
> My only concern is that I think running 'mkdocs serve' will gradually take up more memory as it runs and the set grows.

Yes, I understand the concern. However, the messages are stored in a set so we only store one copy of each message. Of course, there are other possible messages (besides the two) which don't repeat and those will be cached as well. But, give each message is only stored once, the total number of messages should stay relatively short. Once the site is built, the cache should not grow any unless an edit introduces a new error/warning that had not previously been raised.

Also note that this only applies to `build.py`. It does not apply to anything outside of that file or anything   logged by the server directly.

To me, given the relatively small impact, the benefits outweigh the concern. Of course, if you disagree, we can explore other options or leave it as-is.

d0ugal:
Cool. I missed that it only happens on build.py. Thanks!

Issue #932, opened 2016-05-05
Document different installation methods


d0ugal:
Currently we tell everyone to do a sudo pip install which isn't ideal for all users. This could be documented with yum, dnf, apt-get, using a virtualenv or some combination of these.

waylan:
So then you would like to add something like the following:

> Or install MkDocs using the package management system of your choice (apt-get, yum, dnf, etc). Please see the documentation for your specific system for details.

Not sure what else we should add. Providing an example command for each tool means we know what the package is called in that system's repository. Unless you are creating and submitting the files to those repositories yourself, you don't have control over that. And then some offer support for multiple different Python versions and have naming conventions to designate which version works with which. It can get complicated fast.

And finally, regarding virutalenv. If someone is using virtualenv, then it seems to me that virtualenv should be providing documentation for how to install in an environment (activate and don't use sudo, etc). Generally, I expect virtualenv users to already know this as they are more likely to be advanced users.

d0ugal:
Yup, that's a fair summary of the situation. Just referencing/linking out to the alternative sources would be fine with me.

I raised this issue after I was asked about it on IRC.

d0ugal:
The problems when installing in Windows came up recently in #957. It would be good to take that into account when revamping the install guide.

waylan:
The windows issues have been addressed in #1028. 

I'll leave this open to remind us to add something for system specific package managers.

PR #957, opened 2016-06-04
Add mkdocs command troubleshooting help on Windows

advancedrei:
Searched everywhere to figure out how to fix this, finally found it here: https://www.sitepoint.com/building-product-documentation-mkdocs/. Adding to the docs to save the next poor soul from wasting 2 hours. :/

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -66,6 +66,20 @@ $ mkdocs --version
 mkdocs, version 0.15.3
 ```
 
+### Troubleshooting Installation
+
+If you are on Windows, and the `mkdocs` command is not working, you may need to add the Python Scripts folder to your `PATH` Environment variable.
+
+If using Chocolatey:
+```
+C:\ProgramData\chocolatey\lib\python3\tools\Scripts
+```
+
+If you installed Python from their Windows installer:
+```
+C:\[path-to-python-folder]\Scripts
+```
+
 ---
 
 ## Getting started
--- a/docs/index.md
+++ b/docs/index.md
@@ -68,17 +68,14 @@ mkdocs, version 0.15.3
 
 ### Troubleshooting Installation
 
-If you are on Windows, and the `mkdocs` command is not working, you may need to add the Python Scripts folder to your `PATH` Environment variable.
+If you are on Windows, and the `mkdocs` command is not working, you may need to 
+add the Python Scripts folder to your `PATH` Environment variable.
 
 If using Chocolatey:
-```
-C:\ProgramData\chocolatey\lib\python3\tools\Scripts
-```
+`C:\ProgramData\chocolatey\lib\python3\tools\Scripts`
 
 If you installed Python from their Windows installer:
-```
-C:\[path-to-python-folder]\Scripts
-```
+`C:\[path-to-python-folder]\Scripts`
 
 ---
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -68,7 +68,7 @@ mkdocs, version 0.15.3
 
 ### Troubleshooting Installation
 
-If you are on Windows, and the `mkdocs` command is not working, you may need to 
+If you are on Windows, and the `mkdocs` command is not working, you may need to
 add the Python Scripts folder to your `PATH` Environment variable.
 
 If using Chocolatey:


waylan:
You have a few [Markdown Lint errors][1] that will need to be corrected before this is accepted, if it is. To me this addition seems unnecessary, but then I've been using Python for years and know this without even thinking about it (I'd be surprised if it worked without doing that). I realize that may not be the case for new users. I just wonder if this could be worked into the Windows install instructions in a more subtle way rather than creating a new sub-section in the docs.

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/135322113#L194

advancedrei:
@waylan I have been a .NET developer for 15 years, but was new to Python until today. The Chocolatey manager, which this project recommends, does not appear to update the PATH variable when installing Python. So I wanted to make sure users get the info they need.

The reason I created a separate section is because the docs just assume your installation will be successful. They do not provide any assistance if it fails, which based on a review of the Issues and Google Groups, can happen often. This was my attempt to start building out a troubleshooting section of common issues for all platforms, so new users can get productive more quickly.

waylan:
> does not appear to update the PATH variable when installing Python.

That has been normal behavior for as long as I can remember, regardless of installation method. IIRC, some newer Python versions do update the PATH variable when installed in certain ways, but that is only for the Python runtime (`python.exe`), not the Python "bin" dir (annoyingly named "Scripts" on Windows). The docs should be written with that assumption in mind. Incorporating a comment into the basic installation instructions rather than a separate "troubleshooting" section would accomplish that.

As an aside, I always assumed Chocolatey was being recommended because it handled things like updating the PATH variable (I've never used Chocolatey personally). Perhaps the problem is that none of the core developers use Windows as their primary/development system.

advancedrei:
Yes, that seems like a pretty big problem. Another one is that the documentation assumes a lot about the skill level and familiarity of the system. Now that .NET is OSS and leveraging projects like this, that's probably not a good thing.

I really don't care what you guys do, I doubt I'll have much else to contribute to the project (at least in the way of code). The right thing to do is to accept the PR so I get contribution credit, and then modify the text as you choose. However, I will point out that your docs assume success and not failure, and the Google Group and Issues section have several different ways installation can fail. To save you support headaches (and keep your users from wasting time) you should probably document those things. Then again, you guys develop with Python, so maybe you like headaches 😀 🙊 😶 😉

waylan:
@d0ugal this is your call.

d0ugal:
> Perhaps the problem is that none of the core developers use Windows as their primary/development system.

Indeed, this is the issue. I don't know what is best to recommend here.

d0ugal:
> Yes, that seems like a pretty big problem. Another one is that the documentation assumes a lot about the skill level and familiarity of the system. Now that .NET is OSS and leveraging projects like this, that's probably not a good thing.

@advancedrei Which projects are using MkDocs?

d0ugal:
> I really don't care what you guys do, I doubt I'll have much else to contribute to the project (at least in the way of code). The right thing to do is to accept the PR so I get contribution credit, and then modify the text as you choose. However, I will point out that your docs assume success and not failure, and the Google Group and Issues section have several different ways installation can fail. To save you support headaches (and keep your users from wasting time) you should probably document those things. Then again, you guys develop with Python, so maybe you like headaches 😀 🙊 😶 😉 

The tone here isn't helpful, we are all volunteers and trying to help make things better. :sparkles: :heart:

d0ugal:
I don't recall doing any  PATH changes when I installed with Chocolately and I am fairly sure I copied everything down that I done. However, this was probably over a year ago and in a windows VM that I don't have now. I'm not sure if anything has changed since then.

Windows support is a "best effort" because we don't have any volunteers on the MkDocs team that are able and willing to help out. Therefore I am going to close this as it stands. I would be very happy for somebody to drive this forward and improve the docs. I had considered the different install options lately #932, perhaps as part of that work we should improve it for Windows users.

advancedrei:
@d0ugal ASP.NET WebAPI and RESTier are likely to start using it at my recommendation... though I'm now likely to see if the Sphinx experience is more user-friendly.

d0ugal:
Okay, I think the issues you had are all Python-on-Windows related so Sphinx is likely to be similar (but requires more understanding of Python IMO). However, It may be better documented as a far more mature project. Anyway, good luck and do report any specific MkDocs issues if you run into them!

PR #967, opened 2016-06-17
Error on failure to git push as part of gh-deploy

### LINKED ISSUE #966: gh-deploy command fails silently
If the `git push` command fails for some reason (e.g., creds access), we don't check the result and then fail the deploy.

PR with a fix coming momentarily.

pjbull:
Ran into this thinking my CI server had deployed properly, but it hadn't because of an authentication issue. However, the process returned a success code and the logs indicated success.

Fixes #966 

One note is that this doesn't raise on error. Happy to update if you'd like to have this raise an exception in addition to logging the error.

### DIFF
--- a/mkdocs/utils/ghp_import.py
+++ b/mkdocs/utils/ghp_import.py
@@ -169,5 +169,8 @@ def ghp_import(directory, message, remote='origin', branch='gh-pages'):
 
     proc = sp.Popen(['git', 'push', remote, branch],
                     stdout=sp.PIPE, stderr=sp.PIPE)
-    proc.communicate()
-    return proc.wait() == 0
+
+    out, err = proc.communicate()
+    result = proc.wait() == 0
+
+    return result, dec(err)
--- a/mkdocs/commands/gh_deploy.py
+++ b/mkdocs/commands/gh_deploy.py
@@ -65,29 +65,32 @@ def gh_deploy(config, message=None):
     log.info("Copying '%s' to '%s' branch and pushing to GitHub.",
              config['site_dir'], config['remote_branch'])
 
-    ghp_import.ghp_import(config['site_dir'], message, remote_name,
-                          remote_branch)
-
-    cname_file = os.path.join(config['site_dir'], 'CNAME')
-    # Does this repository have a CNAME set for GitHub pages?
-    if os.path.isfile(cname_file):
-        # This GitHub pages repository has a CNAME configured.
-        with(open(cname_file, 'r')) as f:
-            cname_host = f.read().strip()
-        log.info('Based on your CNAME file, your documentation should be '
-                 'available shortly at: http://%s', cname_host)
-        log.info('NOTE: Your DNS records must be configured appropriately for '
-                 'your CNAME URL to work.')
-        return
-
-    host, path = _get_remote_url(remote_name)
-
-    if host is None:
-        # This could be a GitHub Enterprise deployment.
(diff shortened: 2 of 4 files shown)


pjbull:
Alright, actually decided to raise `SystemExit` on error as happens with some of the other commands. This will work better with CI servers that check exit codes to determine success/failure.

I added the `--force` flag to the `git push` call after seeing some pushes fail. This shouldn't be problematic since we already warn users that `gh-pages` will [get destroyed on deploy](http://www.mkdocs.org/user-guide/deploying-your-docs/). Looks like the [ghp-import](https://github.com/davisp/ghp-import/blob/master/ghp_import.py#L201) project already added an option to do the same thing.

waylan:
Thanks for taking the time to work on this. At a quick glance this looks good. I'm out of town through the weekend and will take a closer look when I get home. However, could you point to the recent change to ghp-import where they added the `--force` option? I just want to verify that we aren't veering to far from that project. Maybe we need to update our vendored copy?

Also, not a big deal, but it would be nice if you could squash your commits for a cleaner commit history.  I would make the force flag one commit and everything else another.

mkdocs/utils/ghp_import.py
```diff
@@ -167,7 +167,10 @@ def ghp_import(directory, message, remote='origin', branch='gh-pages'):
 
     run_import(directory, branch, message, nojekyll)
 
-    proc = sp.Popen(['git', 'push', remote, branch],
+    proc = sp.Popen(['git', 'push', remote, branch, '--force'],
```
waylan: ghp_import leaves this as an option for the user. If we are going to add it, so should we. I understand the argument that your history gets hoses anyway, but that same argument could be used for ghp_import (we just repeat their warning) and they felt the need to make it optional. I'm inclined to follow their lead.

Maybe we should just update our vendored copy to a more recent version of ghp_import.

waylan:
@pjbull lets remove the addition of the `--force` option. We can always add that later if/when we update our local vendored copy of the ghp_import script. Looks good otherwise. Would be even better if you squashed all your commits into one for a clean history.

d0ugal:
+1 to what @waylan said.

pjbull:
Awesome. I'll squash the commits.

It looked like I needed force to get the deploy from my CI (codeship) to work. I'll see if that is actually the case or if we can swing something else. Will handle that separately from this PR.

d0ugal:
I just looked through this a bit more. In theory, force should never be required as the code attempts to rebase the local GitHub pages branch on the remote one. So I would be very interested in a failing case for this, ideally it would be better to try and fix it elsewhere, rather than using force.

It looks like `--force` was added in [February last year](https://github.com/davisp/ghp-import/pull/32), which is actually *after* I imported the code to MkDocs. I may have removed it for some reason, I did have to do various tweaks to make the code work well as a library rather than a CLI tool. We essentially have a fork now, which is perhaps a mistake. So maybe we should try and go back to a direct copy of ghp-import - however, we do have test coverage and a few other things that they don't have, so I'm not sure about that either.

pjbull:
Thanks @d0ugal, I added #973 to track the failed push the `gh-pages`. Will probably get to that and cleaning this PR up this weekend.

pjbull:
Removed the `--force` flag (see [this line](https://github.com/mkdocs/mkdocs/pull/967/files#diff-51e7c336b6e4056433a917f8d0577ecdL170)) and squashed commits.

Should be good to go if CI checks pass.

:+1:

PR #970, opened 2016-06-22
Pass loader option up call chain

aeslaughter:
The yaml_load function has the ability to pass a custom loader. However, this option was lacking from the items calling this function.

This seems like a reasonable fix that allows for users to do whatever they want with their yaml files. Fo example, the ability to include yaml files within each other, as discussed here #942.

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 import logging
 import shutil
 import tempfile
+import yaml
 
 from mkdocs.commands.build import build
 from mkdocs.config import load_config
@@ -51,7 +52,7 @@ def _static_server(host, port, site_dir):
 
 
 def serve(config_file=None, dev_addr=None, strict=None, theme=None,
-          theme_dir=None, livereload=True):
+          theme_dir=None, livereload=True, loader=yaml.Loader):
     """
     Start the MkDocs development server
 
@@ -70,6 +71,7 @@ def serve(config_file=None, dev_addr=None, strict=None, theme=None,
             strict=strict,
             theme=theme,
             theme_dir=theme_dir,
+            loader=loader
         )
         config['site_dir'] = tempdir
         build(config, live_server=True, clean_site_dir=True)
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 import logging
 import os
+import yaml
 
 from mkdocs import exceptions
 from mkdocs import utils
@@ -96,8 +97,8 @@ class Config(utils.UserDict):
         self.user_configs.append(patch)
         self.data.update(patch)
 
-    def load_file(self, config_file):
-        return self.load_dict(utils.yaml_load(config_file))
+    def load_file(self, config_file, loader):
+        return self.load_dict(utils.yaml_load(config_file, loader))
 
 
 def _open_config_file(config_file):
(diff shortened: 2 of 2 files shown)


waylan:
As with #942, I'm not convicted this belongs in MkDocs. I would expect whatever use case this could potentially solve could also be addressed by the proposed Plugin API (see #206). Seems to me like this should wait to be a part of that, if it is accepted at all.

aeslaughter:
@waylan I am guessing you probably are not interested in a similar ability to pass a custom livereload.Watcher object either, which allows me to create custom dependencies for the live server reload.

I am working on a large project that will have 10,000's of pages, so having the ability to customize the yaml loader is critical. If I can get these two features (custom yaml.Loader and Watcher) I won't have to maintain a fork, which is not appealing.

I will stop trying if this PR gets rejected, thanks for the consideration.

waylan:
@aeslaughter the issue is that we have a planned way for people to alter the behavior of MkDocs: the Plugin API. We have no interest in adding a similar feature separate from the Plugin API. When the Plugin API is added, then we would need to maintain two ways to accomplish the same thing, which is not reasonable. In other words, if you want to do the work to provide a way to alter MkDocs behavior, your time would best be spent on the Plugin API. For example, we need a clearly defined set of insertion points/events that users can connect to. Perhaps start a wiki page defining those events and if we decide that something does not belong in the Plugin API, then we can consider offering a different way for users to accomplish that. But until we have a clearly defined Plugin API, the assumption will be that all customization not already implemented will be deferred to the Plugin API.

aeslaughter:
Thanks, I will add some comments to #206.

waylan:
For the sake of argument, lets assume this is outside of the scope of the Plugin API (see [this comment][1]). I still have a few concerns about this PR. 

The entry point for changing the YAML loader is in the call to `mkdocs.serve.serve` (a `loader` argument has been added). But how would a user tell MkDocs to pass in a different loader? A command-line flag could be added to the CLI (in `__main__.py`). But that would pas in a string, not a Python object (`yaml.Loader` subclass). I suppose the CLI flag could be configured to try to import the user provided string, and upon success, pass that to the `mkdocs.serve.serve` function, or error on failure.

But that is clearly missing from this proposed patch, which suggests to me that you are not intending to use it that way. The only other way this would be useful is if you were using MkDocs as a library by importing `mkdocs.serve.serve` and calling it in your own code. While this is technically possible, we do not consider MkDocs to be a library, but a end-user tool. In other words, we offer to guarantee of any internal API stability and do not officially support such a use. For example, the next release could completely remove the `mkdocs.serve.serve` function, or move it, or change the arguments completely, etc, and we would not consider that a backward incompatible change and would not warn users in the release notes as long as the public `serve` command continues to work from CLI as documented. As this PR is adding a feature which would only be accessible in a way we do not officially support, I can't in good conscience accept this.

[1]: https://github.com/mkdocs/mkdocs/issues/206#issuecomment-229170258

aeslaughter:
> The only other way this would be useful is if you were using MkDocs as a library by importing mkdocs.serve.serve and calling it in your own code.

This is what I am doing for now, I will be happy to re-factor this as need to make it more general. But, as I learn more about the mkdocs project I see where this methodology is a problem.

The reason for this method is that I am essentially generating the `pages:` portion of the config file in multiple other files (each generated page contains thousands of pages itself). So, in my `mkdocs.yml` I am doing the following:

```
pages:
    - Kernels: !include generated/kernels/pages.yml
    - BoundaryConditions: !include generated/bcs/pages.yml
```

So, all I really need is a way to inject the pages variable into the Config object, which can probably be done using the API stuff we are discussing during a "config" event, inside this custom function I can build up this dict() and set the 'pages' directly.

P.S. Thanks for indulging me on all of this stuff, hopefully it will pay off for you and I can help hammer out this plugin API.

PR #971, opened 2016-06-22
Adds cache of pages when live reloading is enabled

aeslaughter:
Based on my past success trying to contribute enhancements, this probably won't go well 😄 , but I will try anyway.

This caches `Page` objects enabling livereload to much more efficient, which is nice when you have a large project.

This is in response to: https://groups.google.com/forum/#!topic/mkdocs/UaukffzOZMQ

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -247,11 +247,15 @@ def build_extra_templates(extra_templates, config, site_navigation=None):
         utils.write_file(output_content.encode('utf-8'), output_path)
 
 
-def build_pages(config, dump_json=False):
+def build_pages(config, dump_json=False, cache=None):
     """
     Builds all the pages and writes them into the build directory.
     """
-    site_navigation = nav.SiteNavigation(config['pages'], config['use_directory_urls'])
+    if cache:
+        site_navigation = cache
+        site_navigation.update (config['pages'], config['use_directory_urls'])
+    else:
+        site_navigation = nav.SiteNavigation(config['pages'], config['use_directory_urls'])
     loader = jinja2.FileSystemLoader(config['theme_dir'] + [config['mkdocs_templates'], ])
     env = jinja2.Environment(loader=loader)
 
@@ -299,6 +303,14 @@ def build_pages(config, dump_json=False):
     for page in site_navigation.walk_pages():
 
         try:
+            # Do nothing if the page hasn't changed
+            if (os.path.getmtime(page.input_path) == page.modified_time):
+                continue
+
+            # Set the modified time if it has not been set
+            if page.modified_time == None:
+                page.modified_time = os.path.getmtime(page.input_path)
+
             log.debug("Building page %s", page.input_path)
             build_result = _build_page(page, config, site_navigation, env,
                                        dump_json)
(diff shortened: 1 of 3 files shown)


waylan:
This is interesting, but I do have a few questions/concerns. The `build` command already has the option to build `--clean` or not. In fact the default is not a clean build. In other words, the default build offers the same end result as this "cache" feature, albeit implemented in a different way. I realize that the `serve` command forces "clean" with no option to override. I guess it seems to me the way to get this behavior would be to just enable `serve` to disable forcing  "clean".

Regardless, any time there is any change to any global template variables (including the nav), then all pages need to be rebuilt. For example, one page has its title changed. That title is displayed in the nav on every page, so every page needs to be rebuilt. So are we willing to forego consistency across pages for slightly quicker builds when using the dev server? I expect different people will give different answers to that question. I suppose leaving to the user to decide makes sense. Although, I think the default should stay as-is.

Still, I doubt we want to maintain 2 different code paths for accomplishing the same thing. If this gives us something better than not disabling  `--clean` for the `serve` command, then maybe this should replace that feature/code. However, if there is no clear advantage, then just add a flag to the `serve command` to disable the `clean` argument passed to `build`.

d0ugal:
@aeslaughter Did you intend to close this?

I support making MkDocs faster, that would be a great thing. I think generally it is fast enough for most users, but there will always be some larger projects that have trouble.

> In other words, the default build offers the same end result as this "cache" feature, albeit implemented in a different way.

The default approach is the worst implementation of a cache, as it only caches stale artefacts. Every page is rebuild and overwrites the existing pages. So there is no speed boost. Honestly, the fact that `--clean` isn't the default is a bug and we should change that for `1.0` and perhaps add a new flag that means `--dont-clean`, but with a better name.

I would like to see a smarter cache be explored, perhaps there is a way we can cover enough of the edge cases to make it okay.

waylan:
> The default approach is the worst implementation of a cache

I suspected as much (never really dug into that part of the code). And that is why I added my last paragraph above. We may actually be able to use this with some modifications.

aeslaughter:
@d0ugal I did close this on purpose, it failed all sorts of tests. So, as I expected it is far more involved than what I need at the moment. It is something I may come back to in the future, but for now this works for my basic use case. It is also not a critical component to what I am trying to do.

As far as the use goes, I understand that all links and such will not be correct, but for what I (and those who will be using our tool) need this quick reload is useful. People will be writing snippets of markdown that explain physics equations (lots of latex math), these snippets are then automatically placed into a larger markdown file that contains auto generated syntax for a framework. It is really nice to have the markdown you are writing show up live, using the same parser as it will be deployed with, but waiting for all the pages to be created.

The purpose of all this is to have a single source documentation channel, where the markdown may be converted to a website (mkdocs), slides (Reveal.js), or pdf (to-be-determined).
Anyway, I will be rolling out a demo of my project soon, I will be sure to pass a link along so you all can see what type of crazy stuff I am doing with your tool. Which, by-the-way, is a very well written bit of code that I am enjoying working with.

aeslaughter:
On more comment, perhaps fixing up something similar to what I have here and  calling it `--develop` or `--quick` would be useful, just be sure to include documentation regarding that all pages will not be correct.

waylan:
For anyone following this discussion, it has been picked up and continued in the open issue at #990.

PR #975, opened 2016-06-27
Direct "Edit on ..." links to individual pages

### LINKED ISSUE #269: 'Edit on Github' should link to the page, not the project

lorengordon:
This patch adds support for a configuration option `edit_uri` that is used to generate a link directly to an individual page in the source repository.

Fixes #269

### DIFF
--- a/docs/user-guide/custom-themes.md
+++ b/docs/user-guide/custom-themes.md
@@ -259,6 +259,11 @@ documentation page.
 The full, canonical URL to the current page. This includes the `site_url` from
 the configuration.
 
+##### page.edit_url
+
+The full URL to the input page in the source repository. Typically used to
+provide a link to edit the source page.
+
 ##### page.url
 
 The URL to the current page not including the `site_url` from the configuration.
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -116,6 +116,9 @@ def get_page_context(page, content, toc, meta, config):
     if config['site_url']:
         page.set_canonical_url(config['site_url'])
 
+    if config['repo_url']:
+        page.set_edit_url(config['repo_url'], config['edit_uri'])
+
     page.content = content
     page.toc = toc
     page.meta = meta
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -202,6 +202,12 @@ class RepoURL(URL):
             else:
                 config['repo_name'] = repo_host.split('.')[0].title()
 
+        if config['repo_url'] is not None and config.get('edit_uri') is None:
+            if config['repo_name'].lower() == 'github':
+                config['edit_uri'] = 'blob/master/docs/'
+            elif config['repo_name'].lower() == 'bitbucket':
+                config['edit_uri'] = 'src/default/docs/'
+
 
 class Dir(Type):
     """
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -151,6 +151,7 @@ class Page(object):
         # Placeholders to be filled in later in the build
         # process when we have access to the config.
         self.canonical_url = None
+        self.edit_url = None
         self.content = None
         self.meta = None
         self.toc = None
@@ -185,6 +186,18 @@ class Page(object):
         if not base.endswith('/'):
             base += '/'
         self.canonical_url = utils.urljoin(base, self.abs_url.lstrip('/'))
+
+    def set_edit_url(self, repo_url, edit_uri):
+        if not repo_url.endswith('/'):
+            repo_url += '/'
+        if not edit_uri:
+            self.edit_url = repo_url
+        else:
+            if not edit_uri.endswith('/'):
+                edit_uri += '/'
+            self.edit_url = utils.urljoin(
+                repo_url + edit_uri,
+                self.input_path)
 
 
 class Header(object):
--- a/mkdocs/themes/readthedocs/breadcrumbs.html
+++ b/mkdocs/themes/readthedocs/breadcrumbs.html
@@ -13,12 +13,13 @@
     {% if page %}<li>{{ page.title }}</li>{% endif %}
     <li class="wy-breadcrumbs-aside">
       {%- block repo %}
-      {% if repo_url %}
-        {% if repo_name == 'GitHub' %}
-          <a href="{{ repo_url }}" class="icon icon-github"> Edit on GitHub</a>
-        {% elif repo_name == 'Bitbucket' %}
-          <a href="{{ repo_url }}" class="icon icon-bitbucket"> Edit on BitBucket</a>
-        {% endif %}
+      {% if page and page.edit_url %}
+        <a href="{{ page.edit_url }}"
+        {%- if repo_name|lower == 'github' %}
+          class="icon icon-github"
+        {%- elif repo_name|lower == 'bitbucket' %}
+          class="icon icon-bitbucket"
+        {% endif %}> Edit on {{ repo_name }}</a>
       {% endif %}
       {%- endblock %}
     </li>
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -50,6 +50,37 @@ When set, provides a link to your GitHub or Bitbucket repository on each page.
 **default**: `'GitHub'` or `'Bitbucket'` if the `repo_url` matches those
 domains, otherwise `null`
 
+### edit_uri
+
+Path from the base `repo_url` to the docs directory when directly viewing a
+page, accounting for specifics of the repository host (e.g. GitHub, Bitbucket,
+etc), the branch, and the docs directory itself. Mkdocs concatenates `repo_url`
+and `edit_uri`, and appends the input path of the page.
+
+When set, provides a link directly to the page in your source repository. This
+makes it easier to find and edit the source for the page. If `repo_url` is not
+set, this option is ignored.
+
+For example, for a GitHub-hosted repository, the `edit_uri` would be as follows.
+(Note the `blob` path and `master` branch...)
+
+```yaml
+edit_uri: blob/master/docs/
+```
+
+For a Bitbucket-hosted repository, the equivalent `edit_uri` would be as
+follows. (Note the `src` path and `default` branch...)
+
+```yaml
+edit_uri: src/default/docs/
+```
+
+For other repository hosts, `edit_uri` works the same way. Simply specify the
+relative path to the docs directory.
+
+**default**: `blob/master/docs/` or `src/default/docs/` for GitHub or Bitbucket
+repos, respectively, if `repo_url` matches those domains, otherwise `null`
+
 ### site_description
 
 Set the site description. This will add a meta tag to the generated HTML header.
--- a/mkdocs/config/defaults.py
+++ b/mkdocs/config/defaults.py
@@ -75,6 +75,12 @@ DEFAULT_SCHEMA = (
     # "GitHub" or "Bitbucket" for known url or Hostname for unknown urls.
     ('repo_name', config_options.Type(utils.string_types)),
 
+    # Specify a URI to the docs dir in the project source repo, relative to the
+    # repo_url. When set, a link directly to the page in the source repo will
+    # be added to the generated HTML. If repo_url is not set also, this option
+    # is ignored.
+    ('edit_uri', config_options.Type(utils.string_types)),
+
     # Specify which css or javascript files from the docs directory should be
     # additionally included in the site. Default, List of all .css and .js
     # files in the docs dir.
--- a/mkdocs/tests/config/config_options_tests.py
+++ b/mkdocs/tests/config/config_options_tests.py
@@ -125,6 +125,27 @@ class RepoURLTest(unittest.TestCase):
         self.assertEqual(config['repo_url'], config['repo_url'])
         self.assertEqual(config['repo_name'], "Launchpad")
 
+    def test_edit_uri_github(self):
+
+        option = config_options.RepoURL()
+        config = {'repo_url': "https://github.com/mkdocs/mkdocs"}
+        option.post_validation(config, 'repo_url')
+        self.assertEqual(config['edit_uri'], 'blob/master/docs/')
+
+    def test_edit_uri_bitbucket(self):
+
+        option = config_options.RepoURL()
+        config = {'repo_url': "https://bitbucket.org/gutworth/six/"}
+        option.post_validation(config, 'repo_url')
+        self.assertEqual(config['edit_uri'], 'src/default/docs/')
+
+    def test_edit_uri_custom(self):
+
+        option = config_options.RepoURL()
+        config = {'repo_url': "https://launchpad.net/python-tuskarclient"}
+        option.post_validation(config, 'repo_url')
+        self.assertEqual(config.get('edit_uri'), None)
+
 
 class DirTest(unittest.TestCase):
 
--- a/mkdocs/themes/mkdocs/nav.html
+++ b/mkdocs/themes/mkdocs/nav.html
@@ -69,9 +69,9 @@
               {%- endblock %}
 
               {%- block repo %}
-                {%- if repo_url %}
+                {%- if page and page.edit_url %}
                     <li>
-                        <a href="{{ repo_url }}">
+                        <a href="{{ page.edit_url }}">
                             {%- if repo_name == 'GitHub' %}
                                 <i class="fa fa-github"></i>
                             {%- elif repo_name == 'Bitbucket' -%}


lorengordon:
I had a need for this feature and didn't see further progress on it since the earlier PRs were closed, so I took a stab at it. I did study https://github.com/mkdocs/mkdocs/pull/752 and https://github.com/mkdocs/mkdocs/pull/802, and read through the original issue https://github.com/mkdocs/mkdocs/issues/269. Hopefully I addressed the concerns and this aligns with the desired approach. Let me know if you'd like me to adjust anything.

mkdocs/commands/build.py
```diff
@@ -143,6 +146,7 @@ def get_page_context(page, content, toc, meta, config):
         'meta': meta,
 
         'canonical_url': page.canonical_url,
+        'edit_url': page.edit_url,
```
waylan: This is deprecated and will be removed in the next release, so lets not even add it here. As the `edit_url` is page specific, then it should only ever be accessible from the `page` object: `page.edit_url`.

mkdocs/themes/readthedocs/breadcrumbs.html
```diff
@@ -13,11 +13,13 @@
     {% if page %}<li>{{ page.title }}</li>{% endif %}
     <li class="wy-breadcrumbs-aside">
       {%- block repo %}
-      {% if repo_url %}
-        {% if repo_name == 'GitHub' %}
-          <a href="{{ repo_url }}" class="icon icon-github"> Edit on GitHub</a>
+      {% if edit_url %}
```
waylan: This should be `{% if page and page.edit_url %}`. As mentioned in another comment, `edit_url` is page specific. I'm okay with no "edit " link when not on a (markdown generated) "page" as there is nothing to edit anyway.

waylan:
@lorengordon overall this looks pretty good. I made a few inline comments. Also, I don't see any changes to the `mkdocs` theme. Make those changes and I think this will be good.

waylan:
Oh, I almost forgot, the new template context  variable needs to be documented [here][1]. Just add an entry for `page.edit_url`.

[1]: https://github.com/mkdocs/mkdocs/blob/master/docs/user-guide/custom-themes.md#page

lorengordon:
@waylan, I'm on it. This is my first time contributing to mkdocs... Any preference for amending the prior commit vs adding a second? I didn't see any particular guidance in the contributing doc.

waylan:
We prefer a clean commit history, so if you can amend the previous commit and do a `--force` push, that would be preferred,  but if not we can work with that as well.

lorengordon:
:+1: Yep, my preference as well. It's not everyone's, so I like to check first.

lorengordon:
@waylan, ok, that should address the comments. Anything else?

docs/user-guide/custom-themes.md
```diff
@@ -118,6 +118,7 @@ favicon           | [site_favicon]        |
 page_description  | [site_description]    |
 repo_url          | [repo_url]            |
 repo_name         | [repo_name]           |
+edit_uri          | [edit_uri]            |
```
waylan: Let's remove this. We don't actually have a global `edit_uri` variable. You have to access it through the config as `config.edit_uri` and I don't think we need or want to advertise that.

mkdocs/themes/readthedocs/breadcrumbs.html
```diff
@@ -13,11 +13,13 @@
...
-        {% if repo_name == 'GitHub' %}
-          <a href="{{ repo_url }}" class="icon icon-github"> Edit on GitHub</a>
+      {% if page and page.edit_url %}
+        {% if repo_name == GitHub %}
+          <a href="{{ page.edit_url }}" class="icon icon-github"> Edit on {{ repo_name }}</a>
         {% elif repo_name == 'Bitbucket' %}
-          <a href="{{ repo_url }}" class="icon icon-bitbucket"> Edit on BitBucket</a>
+          <a href="{{ page.edit_url }}" class="icon icon-bitbucket"> Edit on {{ repo_name }}</a>
+        {% else %}
+          <a href="{{ page.edit_url }}"> Edit on {{ repo_name }}</a>
```
waylan: I feel like this template could be simplified a little. The link is repeated three times with the same variables. The only difference is the use of class to set the icon for the link. Maybe:

```
<a href="{{ page.edit_url }}"
    {%- if repo_name == GitHub %}
        class="icon icon-github" 
    {%- elif repo_name == 'Bitbucket' %}
        class="icon icon-bitbucket"
    {% endif %}> Edit on {{ repo_name }}</a>
```

docs/user-guide/custom-themes.md
lorengordon: Oh my bad, I misunderstood what that table was for.

mkdocs/themes/readthedocs/breadcrumbs.html
lorengordon: I saw that, but was keeping the template changes to a minimum. Let me fiddle with it a bit...

docs/user-guide/custom-themes.md
waylan: Not a problem. That will be getting a refactor soon, which should clear things up a bit.

mkdocs/themes/readthedocs/breadcrumbs.html
waylan: > keeping the template changes to a minimum

I figured as much. Not a big deal, but with the current changes, if makes sense to clean it up now.

mkdocs/nav.py
```diff
@@ -186,6 +187,18 @@ def set_canonical_url(self, base):
...
+            repo_url += '/'
+        if not edit_uri:
+            self.edit_url = repo_url
+        else:
+            if not edit_uri.endswith('/'):
+                edit_uri += '/'
+            self.edit_url = utils.urljoin(
+                repo_url+edit_uri,
+                self.input_path.lstrip('/'))
+
```
waylan: Just took a closer look as this method and I have  few concerns:

1. You are using `utils.urljoin` to join the parts, so you shouldn't need to also be adding `/` manually.

2. I don't see were the page specific path is being added here. Shouldn't `self.input_path` be appended to the end of each url (see this [comment][1])?

[1]: https://github.com/mkdocs/mkdocs/pull/752#issuecomment-173363860

lorengordon:
@waylan, further comments addressed.

mkdocs/nav.py
lorengordon: 1. I was cribbing from `set_canonical_url()`. I can remove those bits if that is not necessary. It gets a little tricky because `utils.urljoin()` only joins a pair of strings, not a list. i.e. this no worky:
```
self.edit_url = utils.urljoin(repo_url, edit_uri, self.input_path.lstrip('/'))
```
So I have to call `urljoin` twice, or make sure that `repo_url` and `edit_url` are properly constructed so I can concatenate them.

2. Line 200 adds `self.input_path`... I am actually building my docs locally to prove to myself that the patch works...

lorengordon:
Alright, think the last line comment is now covered.

waylan:
Thanks. This looks good to me.

The only concern is that this does add a new config setting. I think its necessary to support the feature, but I would like an okay from @d0ugal before merging it. If he's not okay with that, then we would have to hardcode the `edit_uri` value in for each supported `repo`. Of course, the current patch gives us support for most any hosting service out-of-the-box, whereas hardcoding the `edit_uri` would limit support only to the hosting services we code for.

Although, it occurs to me that we could hardcode some defaults for the new setting (if `repo_name` is `github`, default to this or if `repo_name` is `bitbucket` default to that). That way, users would only need to set the value if they want to use a non-standard URL (a different branch than master) on a known service, or an unknown (to us) service (such as GitLab).

For completeness, I should also note that now that we have #947, its possible for users to easily override the template and put anything they want in the "edit this page link" via the `repo` block. While template customization is certainly more difficult for users than altering a setting, it  doesn't require us to offer support for any specific feature and gives the users much more power to do whatever they want. I would think this is the strongest argument against adding a new setting for this feature.

However, as we specifically advertise this project as a tool for documenting your coding projects, explicitly offering this feature with a config setting is a reasonable expectation for users to have.

lorengordon:
Yes, one goal of mine was to support arbitrary repo hosts and branches. I personally like to use a `develop` branch, and would prefer to send users there rather than whatever branch the `repo` host likes to "default" to.

I did take note of the comment on https://github.com/mkdocs/mkdocs/pull/752 about minimizing the number of config options, which is why I combined everything into one setting. Plus, I think it is more flexible this way, anyway.

It appears the config_option classes are able to lookup the config dictionary, so it looks trivial to add a new class, `EditUri()`, similar to the approach for `RepoURL()`, that sets defaults for `GitHub` and `Bitbucket`.

Or, hmm, don't want to run into ordering issues with the config file. Perhaps just extending `RepoURL()` would work? I'll twizzle something out...

lorengordon:
Also, `urljoin()` doesn't appear to work that way. Without the trailing slash on the `base` param, the last fragment gets dropped. So this:

`urljoin('https://example.com/foo', 'bar.md')`

returns this:

`'https://example.com/bar.md'`.

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

lorengordon:
Anyway, I added [this bit](https://github.com/mkdocs/mkdocs/pull/975/files#diff-21a6a972ad3c1097b825235b2e9b4848R205) to set default values for GitHub and Bitbucket. Does that seem reasonable?

lorengordon:
Going that approach, would it be valuable to add a few unit tests to the [RepoURLTest](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/config/config_options_tests.py#L102-L126) class?

```
    def test_edit_uri_github(self):

        option = config_options.RepoURL()
        config = {'repo_url': "https://github.com/mkdocs/mkdocs"}
        option.post_validation(config, 'repo_url')
        self.assertEqual(config['edit_uri'], 'blob/master/docs/')

    def test_edit_uri_bitbucket(self):

        option = config_options.RepoURL()
        config = {'repo_url': "https://bitbucket.org/gutworth/six/"}
        option.post_validation(config, 'repo_url')
        self.assertEqual(config['edit_uri'], 'src/default/docs/')

    def test_edit_uri_custom(self):

        option = config_options.RepoURL()
        config = {'repo_url': "https://launchpad.net/python-tuskarclient"}
        option.post_validation(config, 'repo_url')
        self.assertEqual(config.get('edit_uri'), None)
```

mkdocs/config/config_options.py
```diff
@@ -202,6 +202,12 @@ def post_validation(self, config, key_name):
             else:
                 config['repo_name'] = repo_host.split('.')[0].title()
 
+        if config['repo_url'] is not None and config.get('edit_uri') is None:
+            if config['repo_name'].lower() == 'github':
+                config['edit_uri'] = 'blob/master/docs/'
+            elif config['repo_name'].lower() == 'bitbucket':
+                config['edit_uri'] = 'src/default/docs/'
+
```
waylan: This is exactly what I had in mind.

waylan:
Yes, some more tests would be a good idea. Actually, we should probably add a few tests of the new `Page.set_edit_url` method as well. Just create a few dummy `Page` objects and confirm the method returns the correct result with various input.

lorengordon:
Can you give me a few more pointers for the `Page` tests? I looked at [nav_tests.py](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/nav_tests.py) and poked around a few others, but it wasn't clear how that part of the code was being tested.

waylan:
Sigh. 

It appears that `Page` objects are not being tested directly anywhere. Of course, they used to only be simple objects as part of the nav. However, today they are the container for each actual page and are a major part of how things work. Looks like we need to add some tests. If you want to leave that to be addressed separately, that's fine.

lorengordon:
Yeah, if you don't mind, I'm not yet confident enough in the mkdocs code and data structures to know how to setup the tests without an example to guide me. I'm happy to add them later once the basic constructs are in place though! Just ping me whenever.

waylan:
@d0ugal as far am I'm concerned work on this is done. If your okay with the new setting say the word and I'll merge (or you can merge it yourself).

d0ugal:
Sorry for being so quiet until now. You have both done a wonderful job - I was reading along and this was a fantastic example of how to collaborate in a code review. I just tried it out locally and it works well. 

I think we are just missing an addition in the release nodes. We can add that later.

Thanks both!

viktorbenei:
Sorry if this is not the right channel, just have a related question: is there a reason why you prefer the `/blob/` URL (GitHub) instead of the `/edit` URL? E.g. https://github.com/mkdocs/mkdocs/blob/master/setup.py vs https://github.com/mkdocs/mkdocs/edit/master/setup.py

Based on our experience, users who are not familiar with GitHub (but might want to fix a typo etc.) can have a hard time finding the "edit" option on GitHub, so linking the `/edit/` URL is easier for them.

marcelstoer:
@lorengordon @waylan fantastic, thank you so much!

> so linking the `/edit/` URL is easier

I second that, particularly because the link label says "edit" rather than "read"/"view"/etc.

> We prefer a clean commit history, so if you can amend the previous commit and do a `--force` push

@waylan I fully agree but in my own projects I stopped pestering PR authors with that ever since GitHub gave us the option to squash (and edit commit message) on merge.

d0ugal:
> Sorry if this is not the right channel, just have a related question

@viktorbenei I think this is fine, the only risk is that it get's lost in the noise. So if the discussion lasts a while then we should go back to the [original issue](https://github.com/mkdocs/mkdocs/issues/269).

I had wondered about using the edit url, my only concern is that if you are not logged in, you can't even view the page. I know, if you are going to edit on GitHub you'd expect users to be logged in, but we are talking about less experienced users, so I think it could be jarring for them.

ReadTheDocs uses the blob url, this can be seen in the [pip docs](https://pip.readthedocs.io/en/latest/). I wonder if they have had a discussion we can learn from, or it might just be they added it before GitHub had that feature.

I could be persuaded either way, so let's wait for @lorengordon and @waylan's input.

marcelstoer:
>  if you are not logged in, you can't even view the page

Riiight...haven't thought of that because I'm never logged out 😉

d0ugal:
Me neither, I had to open a page in incognito! :smile:

viktorbenei:
I didn't know about the non logged in issue either.

The login IMO wouldn't be an issue (you clicked on an edit button after all), but I just found out that if you click the "edit/pencil" icon on GitHub it auto creates a fork for you in the background and shows the edit page automatically, while if you link directly to the URL and you don't have a fork of the repo / haven't clicked the edit/pencil icon before on that repo, GitHub won't show the edit page right away. Instead it'll present an info text and button to create the fork like this:

screen shot 2016-06-28 at 12 05 34

(This prompt is only shown if you don't have a fork of the project yet, if you clicked the edit/pencil button at least once then you'll get the edit page directly)

Although I still think that an "edit" link should link to the edit page (in general), I'm less sure about it now. WDYT?

viktorbenei:
That said, this might still be a more guided approach for the user which will eventually lead to the page where they can edit the content directly, instead of knowing that they can click the little pencil icon on the page.

I saw the confusion with developers too, that they don't know about the edit option on GitHub's web UI, and that they think they have to fork & `git clone` the repository, because the pencil icon is too hidden on the UI.

d0ugal:
Given that this can be changed in your mkdocs.yml (To switch to edit, you would add: `edit_uri: edit/master/docs/`), I don't think it really matters. We should just pick one and note the alternative in the documentation.

I'm still not sure what the most sensible default is, but I think I am moving towards changing it to the edit page.

viktorbenei:
I agree, as long as there's an option to select the one which fits the project better I'm fine with either as a default :)

lorengordon:
I'm fine with changing the GitHub default to `edit`. I didn't know about that previously. Doesn't look like Bitbucket has a similar url...appears the edit function is handled in javascript or something? Anyone know otherwise? Here's an example... https://bitbucket.org/jespern/django-piston/src/default/AUTHORS.txt

lorengordon:
See PR #978 for the change to use the `edit` URL.

waylan:
> Given that this can be changed in your mkdocs.yml (To switch to edit, you would add: `edit_uri: edit/master/docs/`), I don't think it really matters. We should just pick one and note the alternative in the documentation.

This was my thought exactly. I'm indifferent as which is the default.

PR #978, opened 2016-06-28
Use the edit url for the 'Edit on...' links

lorengordon:
Rather than the view-based `blob` URL, default to the `edit` URL for GitHub-hosted repositories. Also update the release notes.

See #975 for previous work and the discussion leading to this change.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -107,6 +107,9 @@ to support such customization.
 * Bugfix: Support SOURCE_DATE_EPOCH environment variable for "reproducible"
   builds (#938)
 * Follow links when copying media files (#869).
+* Change "Edit on..." links to point directly to the file in the source
+  repository, rather than to the root of the repository (#975), configurable
+  via the new [`edit_uri`](../user-guide/configuration.md#edit_uri) setting.
 
 ## Version 0.15.3 (2016-02-18)
 
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -62,10 +62,10 @@ makes it easier to find and edit the source for the page. If `repo_url` is not
 set, this option is ignored.
 
 For example, for a GitHub-hosted repository, the `edit_uri` would be as follows.
-(Note the `blob` path and `master` branch...)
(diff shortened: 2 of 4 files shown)


lorengordon:
If anyone knows the equivalent URL for Bitbucket repos, I can change that as well.

docs/about/release-notes.md
```diff
@@ -107,6 +107,8 @@ to support such customization.
 * Bugfix: Support SOURCE_DATE_EPOCH environment variable for "reproducible"
   builds (#938)
 * Follow links when copying media files (#869).
+* Change "Edit on..." links to point directly to the file in the source
+  repository, rather than to the root of the repository (#975).
```
waylan: This is fine. However, it might be good to point out that a new config setting has been added to make this possible. Maybe add:

    ... configurable via the new [`edit_uri`](../user-guide/configuration.md#edit_uri) setting.

waylan:
I just checked my (mostly unused) Bitbucket account and they don't use a separate URL for edits. When you click the `[edit]` button, the internal content of the page is updated to a textarea via JavaScript. In fact, when I disable JavaScript, the button no longer works and I can't edit the page.

docs/about/release-notes.md
lorengordon: Good idea. Done!

docs/user-guide/configuration.md
```diff
@@ -78,9 +78,16 @@ edit_uri: src/default/docs/
 For other repository hosts, `edit_uri` works the same way. Simply specify the
 relative path to the docs directory.
 
-**default**: `blob/master/docs/` or `src/default/docs/` for GitHub or Bitbucket
+**default**: `edit/master/docs/` or `src/default/docs/` for GitHub or Bitbucket
 repos, respectively, if `repo_url` matches those domains, otherwise `null`
 
+!!! note "Note:"
```
waylan: Just a nitpick, but you don't need the second `"Note:"` here.

lorengordon: I think it maintains stylistic consistency with the other admonitions on the same page...but if you want me to remove it, I will.

PR #979, opened 2016-06-28
Add a note about the MkDocs team members.

### LINKED ISSUE #950: Add a page, or section about the MkDocs team
I'd like to cover who the current and previously active MkDocs team members are. That is myself, Tom and Waylan. I think this can probably be included in the Release Notes page above or below the Upgrading section.

Something like how Django REST Framework does it would be a good guide to follow. http://www.django-rest-framework.org/topics/project-management/#maintenance-team

d0ugal:
We never explicitly listed people before, this should make it
clearer who is behind the project.

Fixes #950


d0ugal:
Went for something simple, but I think this covers what I wanted.

PR #980, opened 2016-06-28
Adopt the PyPA Code of Conduct

d0ugal:


d0ugal:
While I have been expecting this standard from people already, I have been meaning to explicitly state this for a while. The PyPA CoC seems like a good choice, and is used by other popular Python projects like cookiecutter.

PR #981, opened 2016-06-28
Add a note about 404 pages on the deploying page

### LINKED ISSUE #732: Documentation on the 404 feature
It seems like there's a feature in MkDocs, making it possible to specify a 404.html file in a custom template and let MkDocs show that on page not found. I cannot make it work and there's no documentation telling how to configure that.

d0ugal:
Fixes #732

### DIFF
--- a/docs/user-guide/deploying-your-docs.md
+++ b/docs/user-guide/deploying-your-docs.md
@@ -129,3 +129,12 @@ your hosts' file system.
 See your host's documentation for specifics. You will likely want to search
 their documentation for "ftp" or "uploading site".
 
+## 404 Pages
+
+When MkDocs builds the documentation it will include a 404.html file in the
+[build directory][site_dir]. This file will be automatically used when
+deploying to [GitHub](#github-pages) but only on a custom domain. Other web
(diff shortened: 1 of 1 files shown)


docs/user-guide/deploying-your-docs.md
```diff
@@ -129,3 +129,12 @@ your hosts' file system.
 See your host's documentation for specifics. You will likely want to search
 their documentation for "ftp" or "uploading site".
 
+## 404 Pages
+
+When MkDocs builds the documentation it will include a 404.html file in the
+[build directory][site_dir]. This file will be automatically used when
+deploying to [GitHub](#github-pages) but only on a custom domain. Other
+webservers such can be configured to use it but the feature wont always be
```
waylan: You have a few typos in the last sentence. Try:

    Other web servers may be configured to use it but the feature won't always be available.

Also, I wonder if we should add:

    See the documentation for your server of choice for more information.

d0ugal: Updated, thanks!

PR #982, opened 2016-06-28
Add white-space: pre-wrap to code in the readthedocs theme

### LINKED ISSUE #709: `break-word` on inline code blocks is a bit too eager
Example:
image

I understand why it was added in response to issue #313, but I can't help but feel there's a better solution?

d0ugal:
I think this offers the best compromise. This is based on my
testing of the suggestion @facelessuser made.

Fixes #709

### DIFF
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -56,6 +56,7 @@ code {
  */
 p code {
     word-wrap: break-word;
+    white-space: pre-wrap;
 }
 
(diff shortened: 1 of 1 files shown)


d0ugal:
I done a bunch of testing in FireFox and chrome with this and I think it is the best default. However, if anyone else wants to try that would be great.

facelessuser:
I've been using it since I started using Mkdocs in my custom CSS.  So I'm kind of partial to it :).

d0ugal:
Okay, we can try this out.

PR #985, opened 2016-06-28
Remove extra theme_dir alias

d0ugal:
theme_dir is listed twice at the bottom links, this removed the
one that relates to this page. All usages make sense when linking
directly to the config page.


d0ugal:
Spotting the duplicated alias would be a useful linter check.

waylan:
> Spotting the duplicated alias would be a useful linter check.

Agreed. Unfortunately, this doesn't seem to be in the Markdown linter [rules]. I checked for feature requests and only came up with mivok/markdownlint#95 and mivok/markdownlint#136, which are related but don't address this directly. So I made one at mivok/markdownlint#137.

[rules]: https://github.com/mivok/markdownlint/blob/master/docs/RULES.md

Issue #986, opened 2016-06-30
Static resources (especially CSS/JS!) shouldn't be automatically included


jimporter:
(Pulling this out from #972.)

Currently, MkDocs automatically includes all static resources from the docs dir in a build. This can cause issues if some files aren't meant to be included but are located in the docs dir. For example, this could happen with files that get preprocessed by another tool into `.md` files for MkDocs (which was how I planned on doing auto-generation of docs from code for MkDocs if I ever get around to it).

However, there's a more-significant issue: all CSS and JS files are automatically included in the generated HTML itself (except when hosting on RTD, apparently). This means that any CSS/JS used for static HTML files will get erroneously included in the MkDocs-generated HTML by default. The inconsistency with RTD adds another wrinkle to this. I think the behavior should be the same no matter what, and as Python says, "Explicit is better than implicit."

At minimum, I think CSS and JS shouldn't be automatically added to the generated HTML, but I'd prefer to see MkDocs only include files that are explicitly listed in `mkdocs.yml`. Doing so would make things more flexible when people want to do unusual things with MkDocs, and with globbing or the ability to include whole subdirectories of static resources, the configuration wouldn't be much more complex than it is now.

waylan:
I have to admit that I have always found the the `extra_css` and `extra_javascript` settings to be odd (same for `extra_templates`, although that works slightly differently). We provide a way for users to insert their own CSS and JavaScript from the `theme_dir`, but we also allow the same thing through these "extra" settings from the `docs_dir`. Why are we offering users two ways to accomplish the same thing?

I'm guessing the answer is historical reasons. My assumption is that the "extra" settings were added first and the `theme_dir` came later (I haven't actually checked the commit history). But if the "extra" settings are there only for historical reasons, perhaps they should be deprecated altogether. Especially now that the `theme_dir` is being more fully documented and its use and purpose more fully defined (see #911 & #947 to be available in version 0.16). And, hey, it meets our goal of reducing the number of settings.

[...]

PR #988, opened 2016-06-30
Start the deprecation of auto-populating extra_css and extra_javascript

d0ugal:
See #986 

This is perhaps a bit rough, and could probably do with a test addition. I had to make an addition to the validation workflow so that warnings can be added in the post validation stage. At the same time I added support for this in the pre-validation stage too, since the logic is the same.

Outstanding bits:

- [ ] Add test
- [ ] Update the documentation to note this pending change
- [ ] Possibly change the wording of the warning

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -382,6 +382,16 @@ class Extras(OptionallyRequired):
 
         config[key_name] = extras
 
+        if not extras:
+            return
+
+        self.warnings.append((
+            'The following files have been automatically included in the '
+            'documentation build and will be added to the HTML: {}. In '
+            'MkDocs they will need to be explicitly included with the '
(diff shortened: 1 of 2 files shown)


mkdocs/config/config_options.py
```diff
@@ -382,6 +382,16 @@ def post_validation(self, config, key_name):
 
         config[key_name] = extras
 
+        if not extras:
+            return
+
+        self.warnings.append((
+            'The following files have been automatically included in the '
+            'documentation build and will be added to the HTML: {}. In '
+            'MkDocs they will need to be explicitly included with the '
```
waylan: Shouldn't that be: " In MkDocs version 1.0 they will need to be explicitly included ..."?

Notice that I added "version 1.0".

PR #997, opened 2016-07-20
A simple method for providing a quick/dirty reload option (refs #990)

### LINKED ISSUE #990: Support partial builds with serve
Is there any way to speed up the rendering when 'mkdocs serve' is running? 

Currently I have approximately 750 markdown documents in the docs folder and when I make a small change to a file it takes approximately 3-5 minutes to update. 

It's as if everything that sits in the docs file needs to be refactored, not the single document itself.

Anyone know how to overcome this?

aeslaughter:
This provides a --dirty-reload option to the serve command. It simply operates by comparing the modified time of the markdown and html. This is a much simpler solution to what was proposed in 
#971 and is meeting my needs well.

Note, that there was some discussion of --clean/--no-clean with the serve command. However, passing a --clean/--no-clean option to the serve command doesn't do anything since each time it runs is creates a different tmp directory (at least for me). So, effectively it is always clean. 

(closes #990)

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -92,6 +92,26 @@ to support such customization.
 
 [blocks]: ../user-guide/styling-your-docs/#overriding-template-blocks
 
+#### Support for dirty builds. (#990)
+
+For large sites the build time required to create the pages can become problematic,
+thus a "dirty" build mode was created. This mode simply compares the modified time
+of the generated html and source markdown. If the markdown has changed since the
+html then the page is re-constructed. Otherwise, the page remains as is. This mode
+may be invoked in both the `mkdocs serve` and `mkdocs build` commands:
+
+```text
+mkdocs serve --dirtyreload
+```
+
+```text
+mkdocs build --dirty
+```
+
+It is important to note that this method for building the pages is for development
+of content only, since the navigation and other links do not get updated on other
+pages.
+
 ### Other Changes and Additions to Version 0.16.0
 
 * Bugfix: Support `gh-deploy` command on Windows with Python 3 (#722)
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -77,7 +77,9 @@ theme_dir_help = "The theme directory to use when building your documentation."
 theme_help = "The theme to use when building your documentation."
 theme_choices = utils.get_theme_names()
 site_dir_help = "The directory to output the result of the documentation build."
-reload_help = "Enable and disable the live reloading in the development server."
+reload_help = "Enable the live reloading in the development server (this is the default)"
+no_reload_help = "Disable the live reloading in the development server."
+dirty_reload_help = "Enable the live reloading in the development server, but only re-build files that have changed"
 commit_message_help = ("A commit message to use when commiting to the "
                        "Github Pages remote branch")
 remote_branch_help = ("The remote branch to commit to for Github Pages. This "
@@ -101,7 +103,9 @@ def cli():
 @click.option('-s', '--strict', is_flag=True, help=strict_help)
 @click.option('-t', '--theme', type=click.Choice(theme_choices), help=theme_help)
 @click.option('-e', '--theme-dir', type=click.Path(), help=theme_dir_help)
-@click.option('--livereload/--no-livereload', default=True, help=reload_help)
+@click.option('--livereload', 'livereload', flag_value='livereload', help=reload_help)
+@click.option('--no-livereload', 'livereload', flag_value='no-livereload', help=no_reload_help)
+@click.option('-d', '--dirtyreload', 'livereload', flag_value='dirty', help=dirty_reload_help)
 @common_options
 def serve_command(dev_addr, config_file, strict, theme, theme_dir, livereload):
     """Run the builtin development server"""
@@ -115,7 +119,7 @@ def serve_command(dev_addr, config_file, strict, theme, theme_dir, livereload):
             strict=strict,
             theme=theme,
             theme_dir=theme_dir,
-            livereload=livereload,
+            livereload=livereload
         )
     except (exceptions.ConfigurationError, socket.error) as e:
         # Avoid ugly, unhelpful traceback
@@ -123,7 +127,7 @@ def serve_command(dev_addr, config_file, strict, theme, theme_dir, livereload):
 
 
 @cli.command(name="build")
-@click.option('-c', '--clean', is_flag=True, help=clean_help)
+@click.option('-c', '--clean/--dirty', is_flag=True, help=clean_help)
 @click.option('-f', '--config-file', type=click.File('rb'), help=config_help)
 @click.option('-s', '--strict', is_flag=True, help=strict_help)
 @click.option('-t', '--theme', type=click.Choice(theme_choices), help=theme_help)
@@ -139,7 +143,7 @@ def build_command(clean, config_file, strict, theme, theme_dir, site_dir):
             theme=theme,
             theme_dir=theme_dir,
             site_dir=site_dir
-        ), clean_site_dir=clean)
+        ), dirty=not clean)
     except exceptions.ConfigurationError as e:
         # Avoid ugly, unhelpful traceback
         raise SystemExit('\n' + str(e))
@@ -169,7 +173,7 @@ def json_command(clean, config_file, strict, site_dir):
             config_file=config_file,
             strict=strict,
             site_dir=site_dir
-        ), dump_json=True, clean_site_dir=clean)
+        ), dump_json=True, dirty=not clean)
     except exceptions.ConfigurationError as e:
         # Avoid ugly, unhelpful traceback
         raise SystemExit('\n' + str(e))
@@ -190,7 +194,7 @@ def gh_deploy_command(config_file, clean, message, remote_branch, remote_name):
             remote_branch=remote_branch,
             remote_name=remote_name
         )
-        build.build(config, clean_site_dir=clean)
+        build.build(config, dirty=not clean)
         gh_deploy.gh_deploy(config, message=message)
     except exceptions.ConfigurationError as e:
         # Avoid ugly, unhelpful traceback
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -32,6 +32,15 @@ log = logging.getLogger(__name__)
 log.addFilter(DuplicateFilter())
 
 
+def get_complete_paths(config, page):
+    """
+    Return the complete input/output paths for the supplied page.
+    """
+    input_path = os.path.join(config['docs_dir'], page.input_path)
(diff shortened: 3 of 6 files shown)


waylan:
I agree, this is much simpler than a cache system and most likely all we need (comparing modified times using the existing features of the file system). I think I like it. I'll need to look at the specific implementation more closely later (I'm preparing to leave for a long weekend today). But, a few quick things I noticed:

Could we also add a `--dirty` flag to the `build` command which does the same thing? As I stated in #990:

> This would actually be pretty easy. Boolean Flags can support two flags to enable/disable a feature. We can just add the second flag and change the default to be `--clean`. We don't even need to deprecate anything. 

Also, would it make sense to have `--dirty-reload` be a third option with `--livereload/--no-livereload`? (we'd need to change it from a Boolean Flag to a [Feature Switch][1] in order to support three options)  It seems kind of silly to have both `--livereload --dirty-reload` in the same command. Obviously `--dirty-reload` implies `--livereload`, so why make it necessarily redundant. And yes, I realize that because `--livereload` is the default, users don't actually need to use both flags. I'm just trying to keep the interface clean. As an added benefit, using `--no-livereload --dirty-reload`  in the same command would (and should) generate an error, which would be less confusing to users.

Finally, I'm curious; how much does this affect performance on a large site? How noticeable of a difference is there? Actually, if we add the `--dirty` flag to the `build` command, I imagine we could benchmark the builds (`--dirty` vs. `--clean`) relatively easily.

[1]: http://click.pocoo.org/5/options/#feature-switches

aeslaughter:
@waylan I will make the changes you suggest.

> Finally, I'm curious; how much does this affect performance on a large site? How noticeable of a difference is there? Actually, if we add the --dirty flag to the build command, I imagine we could benchmark the builds (--dirty vs. --clean) relatively easily.

My initial build of the site takes about 3 min. and re-building a single file takes almost nothing, which makes live editing possible.

waylan:
> My initial build of the site takes about 3 min. and re-building a single file takes almost nothing

Wow, that seems significant. As I have only ever used MkDocs for smaller projects, sometimes I forget about how build times can affect workflow for larger projects. A 3 min wait for every save would be untenable... But the real reason I asked was to check that these changes were actually shortening the build time as the current exclude-the-clean-flag approach does not actually shorten the build time (as you know).

As I mentioned earlier, I'll be out of town this weekend (leaving in a hour or two), so I'll review more closely next week, unless @d0ugal beats me to it.

aeslaughter:
The latest push includes the command-line argument improvements, including a --dirty flag for the build command.

Initial build:
```
time mkdocs build
INFO    -  Building documentation to directory: /Users/slauae/projects/moose-doc/site 

real	2m39.405s
user	2m15.570s
sys	0m23.849s
```

Dirty build (changing single file):
```
time mkdocs build --dirty
INFO    -  Building documentation to directory: /Users/slauae/projects/moose-doc/site 
INFO    -  The directory contains stale files. Use --clean to remove them. 

real	0m2.405s
user	0m1.481s
sys	0m0.901s
```

I am have a long weekend, so no-worries about getting this finished up next week. Thanks.

mkdocs/__main__.py
```diff
@@ -68,6 +68,7 @@ def common_options(f):
 
 
 clean_help = "Remove old files from the site_dir before building"
+dirty_reload_help = "Only re-build files that have changed when livereload is enabled"
```
waylan: This is redefined later. This line should be removed.

mkdocs/commands/build.py
```diff
@@ -32,6 +32,26 @@ def filter(self, record):
...
+    Return the modified time of the supplied file. If the file does not exists zero is returned.
+    see build_pages for use.
+    """
+    if os.path.exists(file_path):
+        return os.path.getmtime(file_path)
+    else:
+        return 0.0
+
+
+def get_complete_paths(config, page):
```
waylan: It might make more sense to put this (and `modified_time`) in `utils.py`

waylan:
This is looking good. Although, I'm wondering if maybe we should issue a warning when `--dirty` is used. With `--clean`  being the default, it should not be much of a problem, but it wouldn't hurt to issue the warning on `--dirty` anyway. Some users may not understand the tradeoffs and a warning that the current state may contain errors in navigation/etc would be helpful.

It just occurred to me that this ignores media files (in the `site_dir`) and static assets (in the `theme`). Those are still being re-copied each time. I suppose if a user has large images/videos/etc, it would be useful to be able to skip those on reloads. We might want to pass `dirty` on to [`utils.copy_media_files`][1] and/or [`utils.copy_file`][2]. We could address static assets as well, but those are more likely to be rather minimal in comparison.

Also, we need the release notes updated. This should be noted in the major changes section as a default behavior is being changed.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L132
[2]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L91

aeslaughter:
The latest push adds a warning, dirty flags for the copy methods, and moves the modified time function to utils.

One think I noticed, is the `copy_files` doesn't seem to be used anywhere.

mkdocs/commands/build.py
```diff
@@ -339,14 +359,14 @@ def build(config, live_server=False, dump_json=False, clean_site_dir=False):
     for theme_dir in reversed(config['theme_dir']):
         log.debug("Copying static assets from theme: %s", theme_dir)
         utils.copy_media_files(
-            theme_dir, config['site_dir'], exclude=['*.py', '*.pyc', '*.html']
+            theme_dir, config['site_dir'], exclude=['*.py', '*.pyc', '*.html'], dirty=dirty
         )
 
     log.debug("Copying static assets from the docs dir.")
     utils.copy_media_files(config['docs_dir'], config['site_dir'])
```
waylan: Don't we want to pass the value of `dirty` on here? Same applies to any other calls to this function or `utils.copy_file`.

mkdocs/utils/__init__.py
```diff
@@ -155,6 +171,11 @@ def copy_media_files(from_dir, to_dir, exclude=None):
             if not is_markdown_file(filename):
                 source_path = os.path.join(source_dir, filename)
                 output_path = os.path.join(output_dir, filename)
+
+                # Do not copy when using --dirty if the file has not been modified
+                if dirty and (modified_time(source_path) < modified_time(output_path)):
+                    continue
+
                 copy_file(source_path, output_path)
```
waylan: Why not just pass `dirty` on to `copy_file` and implement the comparison once?

waylan:
> One thing I noticed, is the `copy_files` function doesn't seem to be used anywhere.

I assume you mean `copy_file` (no 's').  In any event, you're right. Lets just merge that into the `utils.copy_media_files` function.

aeslaughter:
> I assume you mean copy_file (no 's'). In any event, you're right. Lets just merge that into the utils.copy_media_files function.

So, `copy_file` is called in `copy_media_files` (I searched for `copy_files`), so doing the check here alone should be enough, I think this is what you were wanting.

mkdocs/commands/build.py
```diff
@@ -324,13 +340,17 @@ def build(config, live_server=False, dump_json=False, clean_site_dir=False):
...
         log.info("Cleaning site directory")
         utils.clean_directory(config['site_dir'])
+
     if not live_server:
         log.info("Building documentation to directory: %s", config['site_dir'])
         if not clean_site_dir and site_directory_contains_stale_files(config['site_dir']):
             log.info("The directory contains stale files. Use --clean to remove them.")
 
+    # Set the status of a dirtyreload feature, a build is considered 'dirty' if clean_site_dir = False
+    dirty = not clean_site_dir
```
waylan: Sorry I missed this before, but let's just change the `build` function's argument from `clean_site_dir=False` to `dirty=False`. After all, we are changing the default, and the consistent API will be less confusing later.

mkdocs/commands/build.py
```diff
@@ -324,13 +340,17 @@ def build(config, live_server=False, dump_json=False, clean_site_dir=False):
     if clean_site_dir:
         log.info("Cleaning site directory")
         utils.clean_directory(config['site_dir'])
+
     if not live_server:
         log.info("Building documentation to directory: %s", config['site_dir'])
         if not clean_site_dir and site_directory_contains_stale_files(config['site_dir']):
             log.info("The directory contains stale files. Use --clean to remove them.")
```
waylan: Perhaps the warning should be issued here rather than in the serve command. That way it will be issued for any build. In fact, it is actually more important here, as a user could push the `site_dir` to production after running a "dirty" build. Whereas the temp site_dir of the `serve` command is always deleted when the server is stopped.

aeslaughter: Done and much better as well.

mkdocs/commands/build.py
aeslaughter: This is now changed and much better.

waylan:
Looks good. If you could add a note to the release notes, I'll merge it.

I should note that the `--clean` flag is mentioned in `docs/user-guide/deploying-your-docs.md`. The purpose is not explained, but it appears in multiple examples. If doesn't really hurt anything to leave it there, but given the default has changed, its not necessary. If you are so inclined, updating the documentation would be appreciated as well.

aeslaughter:
@waylan I added some the requested documentation and release note updates. Give it a read to see if I made some typos. I put this in separate commit, let me know if you want me to squash it all together or not.

docs/about/release-notes.md
```diff
@@ -92,6 +92,25 @@ to support such customization.
...
+```text
+mkdocs serve --dirtyreload
+```
+
+```text
+mkdocs build --dirty
+```
+
+It is important to note that this method for building the pages is for development
+of content only, since the navigation and other links do not get updated.
```
waylan: Can we add "on other pages" to the end of this sentence (or something similar)? I just want to make it clear that just because the nav is correct on the "updated" page, that does not mean it is on others.

docs/about/release-notes.md
```diff
@@ -119,6 +138,7 @@ to support such customization.
   repository, rather than to the root of the repository (#975), configurable
   via the new [`edit_uri`](../user-guide/configuration.md#edit_uri) setting.
 
+
```
waylan: This extra line appears to be generating this [markdown-lint error](https://travis-ci.org/mkdocs/mkdocs/jobs/147590121#L198).

waylan:
I prefer the documentation regarding a change to be with the associated changes, so if you could squash, that would be preferred. That said, I would have taken it as-is if it wasn't for the Markdown-lint error.

aeslaughter:
@waylan Should be good now, thanks for helping me get this in.

waylan:
@aeslaughter thanks for the work on this and your patience with my endless nitpicks.

PR #1009, opened 2016-08-03
Don't override config value for strict mode if not specified on CLI.

### LINKED ISSUE #738: Configuration strict option not checking broken links
I was testing mkdocs earlier this year and was using "strict: true" in my mkdocs.yml. If I recall correctly, it was working then and building would break if there was a broken link in the .md.

However I am now using v0.14.0 and this config option seems to have no effect on the build - I get no warnings or errors when building regardless of the value of strict.

waylan:
Fixes #738 

This probably still needs a test or two. I just ran out of time. I'll get to it later.


waylan:
Also needs a comment in the release notes.

waylan:
I added a not to the release notes and a test. But the test is failing and its not clear to me why. I'm not an expert on Mock, so maybe I'm misunderstanding something here. But I don't know how else to test this.

PR #1013, opened 2016-08-07
Allow someone to just set the navigation

wpsmith:
I have a need to build all the pages (some pages have internal links to other pages) but only expose a subset of those pages in the navigation. This PR allows for that.


wpsmithtwc:
Has issues, needs closing with no merging.

wpsmith:
Has issues, needs closing with no merging.

waylan:
Yeah, it would be significantly more complicated to accomplish than that. I suspect what you are asking for will not be supported unless it can be done via a plugin (see #206).

PR #1014, opened 2016-08-07
Make livereload the default as stated by the docs.

memeplex:


waylan:
When I was working on the tests for #1009, I noticed that the default was `None`, which I found odd. But I didn't have time to investigate at the time. Of course, now the test I added in #1009 is wrong and causing your fix to fail. The test needs to be updated by setting `livereload='livereload'` [here][1]. Add that and I'll merge.

[1]: https://github.com/waylan/mkdocs/blob/bc35869d34ac61e11baa6314d787fa8cb6110fbc/mkdocs/tests/cli_tests.py#L31

waylan:
Turns out I'm working on #1016 right now. I'll add the test myself and merge this in from there.

waylan:
Closed in favor of #1018.

PR #1023, opened 2016-08-11
Add --force flag to gh-deploy command.

### LINKED ISSUE #973: Sometimes deploy to gh-pages fails
Encountered while fixing #966.  Originally added `--force` in #967, but it makes more sense to do some more research on the root cause.

Possibly:
 - GH token permissions
 - Same GH token used on multiple machines(?)
 - Some git call is failing silently

Will investigate repro steps and update here.

waylan:
This replicates the recent addition to the ghp_import.py tool (see
https://github.com/davisp/ghp-import/commit/49cfe6e6).
The default is to not force, but if needed (perhaps when deploying
from a CI server), the --force flag can be included.

Fixes #973.


waylan:
For some reason one of the appveyor builds timed out. I'm assuming everything is okay and moving forward.

PR #1025, opened 2016-08-13
Remove chocolately recommendation 

dennisroche:
Suggest removing the recommendation to install python and pip using chocolately. Python now comes bundled with pip as of Python 2 >=2.7.9 or Python 3 >=3.4. 

Also, as I found out, the chocolately package has a dependency on `easy_install` which isn't resolved by the script.

I think the new edits are now xplat friendly.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -50,8 +50,6 @@ pip 1.5.2
 
 MkDocs supports Python versions 2.6, 2.7, 3.3, 3.4 and 3.5.
 
-On Windows it is recommended that you install Python and pip with [Chocolatey].
-
 Install the `mkdocs` package using pip:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -66,6 +64,18 @@ $ mkdocs --version
 mkdocs, version 0.15.3
 ```
 
+#### Need to install Python or Pip?
+
+Install [Python] by downloading the installer from [https://www.python.org/downloads/](https://www.python.org/downloads/).
+
+> [pip] is already installed if you're using Python 2 >=2.7.9 or Python 3 >=3.4 downloaded from python.org, but you'll need to upgrade pip.
```
waylan: This is not a quote, so let's not put it in a blockquote. We have the [Admonition Extension][1] enabled, so you can use a "note" or "warning" to make some text stand out as we do at `docs/user-guide/custom-themes.md` among other places.

[1]: https://pythonhosted.org/Markdown/extensions/admonition.html

docs/index.md
```diff
@@ -66,6 +64,18 @@ $ mkdocs --version
...
+
+If you using an older version of [Python] that you can't update and need to install [pip], recommend using the `get-pip.py` script.
```
waylan: This is not a complete sentence. Let's try:

> If you using an older version of [Python] and need to install [pip], you can download and run the `get-pip.py` script.

Then let's point to the script and tell people how to download it.

waylan:
Thank you. Overall, this is a good change. We should have done this a while ago.

In addition to a couple inline comments I made, you have some formatting issues as pointed out the [failing][1] `markdownlint` test. Fix those things up and I'll merge this.

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/152016824

dennisroche:
@waylan thanks for the review. i'll make the changes you suggested.

waylan:
I'm closing this in favor of #1028 where I completely refactored the installation instructions.

@dennisroche I squashed your changes in with mine so you still get credit. Thanks again.

dennisroche:
:+1: no problem.

Issue #725, opened 2015-09-11
Remove site_favicon


carlwgeorge:
The documentation states that the default for `site_favicon` is ["null"](https://github.com/mkdocs/mkdocs/blame/74031a084d776f45ad43568e334da2b89133986b/docs/user-guide/configuration.md#L66-L75).

```md
**default**: `null`
```

However, if `site_favicon` is not specified, the favicon is [set](https://github.com/mkdocs/mkdocs/blob/74031a084d776f45ad43568e334da2b89133986b/mkdocs/themes/mkdocs/base.html#L10-L11) to a [book icon](https://github.com/mkdocs/mkdocs/blob/74031a084d776f45ad43568e334da2b89133986b/mkdocs/themes/mkdocs/img/favicon.ico).

```html
{% if favicon %}<link rel="shortcut icon" href="{{ base_url }}/{{ favicon }}">
{% else %}<link rel="shortcut icon" href="{{ base_url }}/img/favicon.ico">{% endif %}
```
![book](https://raw.githubusercontent.com/mkdocs/mkdocs/74031a084d776f45ad43568e334da2b89133986b/mkdocs/themes/mkdocs/img/favicon.ico)

What is the intention?  Should the book icon be used if `site_favicon` is not set?  Or should it be completely disabled if `site_favicon` is not set?  I think either the behavior should be modified to match the documentation, or the documentation should be updated to match the behavior.

d0ugal:
Good question! I think we should update the theme to match the documentation.

Ultimately it is up to the theme how it wants to handle this, but we should do this for the mkdocs themes at least.

d0ugal:
Good question! I think we should update the theme to match the documentation.

Ultimately it is up to the theme how it wants to handle this, but we should do this for the mkdocs themes at least.

d0ugal:
(Sorry, accidental close)

waylan:
I find it curious that this setting exists at all. While it certainly makes sense for users to be able to use their own custom favicon, it seems like that should be part of theme customization, not as a config setting. 

In fact, if the template was reduced to: 

    <link rel="shortcut icon" href="{{ base_url }}/img/favicon.ico">

Then, as a theme customization, a user could simply include there own favicon at `docs/img/favicon.ico` and their custom favicon will be used rather than the theme provided favicon (at least I'm assuming that it works that way based on how it works for other things). If course, if using a custom theme, the favicon could be included in the theme dir.

Not sure what value there is in also having a setting for it.

[...]

PR #1026, opened 2016-08-15
Global Template vars refactor

waylan:
This is a continuation of the work started in #921. That PR addressed the page context. This addresses the global context (see #874 for more info).

Still a lot of work to do on this. I started by rebasing #681 and gracefully deprecating `include_nav` and `include_next_previous` rather than removing them. I've left the individual commits for now so you can compare my changes to the original.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -27,19 +27,15 @@ The current and past members of the MkDocs team.
 
 #### Template variables refactored. (#874)
 
+##### Page Context
+
 Page specific variable names in the template context have been refactored as
 defined in [Custom Themes](../user-guide/custom-themes/#page). The
 old variable names will issue a warning but continue to work for version 0.16,
 but may be removed in a future version.
 
-No global variables were altered except `page_description`. Previously
-its value was altered programicaly based on whether the current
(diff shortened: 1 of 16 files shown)


waylan:
I've cleaned up the commit history and finished the global template variable refactor. Assuming all the tests pass, this should be ready to go.

The one concern I have is that three variables still exist both in the config and as global variables: `favicon`, `extra_css`, and `extra_javascript`. In [build.py] we are doing the following:

```python
site_favicon = nav.url_context.make_relative('/' + config['site_favicon'])
extra_javascript = utils.create_media_urls(nav, config['extra_javascript'])
extra_css = utils.create_media_urls(nav, config['extra_css'])
```

So the values differ between the config settings and the template variables. I have highlighted this difference in the docs, but wanted to point it out here in case we should be doing something different. It occurs to me that we could create some sort of object (similar to the `page` object) which contained `local_path` and `url` attributes for each object. That way, the config variable would hold both pieces of data side-by-side. However, that feels a little too complex for this already big change. Perhaps that could be done another time.

waylan:
In light of #725 (which I had forgotten about), I just added a commit which deprecates `favicon`, leaving `extra_css` and `extra_javascript` as the only two outliers.

PR #1027, opened 2016-08-16
Account for wide tables and wrap inline code spans.

### LINKED ISSUE #834: Markdown tables expand beyond window with no scrollbar
I created the following test table in a `.md` file:

```
|Test1|Test2|
|-------------|-------------|
|Hello|`really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really long code`|
```

And when loaded via a ReadTheDocs theme the resulting table looks like this:

screen shot 2016-02-21 at 6 39 10 pm

The contents of a row that expand beyond the current window are not viewable. There is also no way to scroll horizontally on the table to view the contents at the end of the row. I am assuming this is happening because the text with the `<code>` tag is not splittable.

* My `mkdocs.yml` file contains:

```
theme: readthedocs

pages:
    - Home: index.md

markdown_extensions:
  - toc:
      permalink: True
  - tables
```

* Viewing the docs using Chrome. 

Any ideas how to get around this?

waylan:
The code span wrapping  eliminates the need to account for the long table as
reported, but other things might cause a table to be wider than the page,
so we still address it.

Fixes #834.

Note that I applied this to both the `readthedocs` and `mkdocs` themes. The same problem existed in both. The only difference with `mkdocs` is the entire page can scroll allowing the table to be viewed. This is better, I think.

### DIFF
--- a/mkdocs/themes/mkdocs/css/base.css
+++ b/mkdocs/themes/mkdocs/css/base.css
@@ -318,3 +318,20 @@ h1:hover .headerlink, h2:hover .headerlink, h3:hover .headerlink, h4:hover .head
     -moz-border-radius: 6px 0 6px 6px;
     border-radius: 6px 0 6px 6px;
 }
+
+/*
+ * Account for wide tables which go off the side.
+ * Also account for long inline code in tables
+ * 
+ * https://github.com/mkdocs/mkdocs/issues/834
+ */
+table {
+  width: 100%;
+  overflow: auto;
+  display: block;
+}
+
+table code {
+  word-wrap: break-word;
+  white-space: pre-wrap;
+}
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -153,3 +153,20 @@ code.cs, code.c {
   border: 1px solid rgba(0, 0, 0, 0.2);
   background: rgba(255, 255, 255, 0.7);
 }
+
+/*
+ * Account for wide tables which go off the side.
+ * Also account for long inline code in tables
+ * 
+ * https://github.com/mkdocs/mkdocs/issues/834
+ */
+.rst-content table.docutils {
+  width: 100%;
+  overflow: auto;
+  display: block;
+}
+
+table code {
+  word-wrap: break-word;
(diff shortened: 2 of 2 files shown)


styfle:
I'm not sure the width is working properly.
Here are screenshots of my readthedocs theme after replacing `mkdocs/themes/readthedocs/css/theme_extra.css`

### Before
image

### After

image

waylan:
Hmm, apparently setting the width to 100% stretches the border for tables which are narrower than the parent's width. Problem is, we need the 100% for tables wider than the parent. I'm sure there's a workaround, I'm just not an expert on CSS.

styfle:
If you are looking to fix #834 then you won't want to change the `table` css. You want to change the `code` css. What about removing the `code {  white-space: pre;  }` and `code { white-space: nowrap; }`

The code should wrap but the code block (aka pre > code) should not which is already handled. So those could be removed.

waylan:
If you notice the diff, I actually did both (wrap the code and scroll the table). Perhaps I should just do the code wrap and leave the table scrolling issue out.

waylan:
On your wide tables, are they still too wide for the page with this change? If not, then we can just remove the table scrolling part and leave the code wrapping.

styfle:
The table is too wide and should not be 100% width. Additionally, there is no reason for a table to scroll.
If anything, maybe the code block should scroll which would be the `<pre>`.

waylan:
Markdown does not allow a code block inside a table, only an inline code span (and altering Markdown's syntax/behavior is not up for consideration). So we are limited in what we can do here with the code. We can wrap it on work breaks, but sometimes code doesn't have word breaks for quit a long stretch of characters. Therefore, it is conceivable that a wrapped code span could still push the left side of a table out past the edge of the page. The only way to make the left edge of that table visible is via overflow/scrolling in that situation.

Thus my questions If you can verify that the very long code spans in the table you pointed to in #834 is being wrapped sufficiently that scrolling is not necessary, then I'll rip the scrolling part of this fix out. If not, I'll try to find a better way to implement the scrolling.

styfle:
Do you want me to make submit a PR with my idea? I think I can fix it very easily.

waylan:
I'm closing this in favor of #1037.

PR #1029, opened 2016-08-17
Update github.io URL

KrauseFx:
- Add missing trailing / when printing out GitHub URL. Without the trailing `/` a GitHub error message is shown
- Switched from `http` to `https`, since it's now enabled by default by GitHub


KrauseFx:
That was fast, thanks for merging 🚀

PR #1033, opened 2016-08-24
Add linters for JS and CSS.

### LINKED ISSUE #929: Add a linter for JavaScript
We check our Python our Markdown and our HTML! We should also check the JavaScript - it is all currently totally untested.

waylan:
Third-party CSS and JS libs have been excluded.

Fixes #929.


waylan:
I've parsed out all the warnings vs. errors and defined then in the .csslintrc` file. However, some of those warnings should probably be errors. We need to go through them one-by-one and alter the rules to eliminate the warnings were possible.  Then, where it makes sense, move some rules to the ignore list.

Currently, the `order-alphabetical` rule is in the ignore list for a few reasons: (1) It occurs a lot---ignoring it makes parsing the list of warnings easier; and (2) updating the rules to alphabetical order makes `git blame` less useful as we work through the other warnings. Perhaps in the end we'll alphabetize everything, but I doubt it.

PR #1034, opened 2016-08-24
Fix #834 with `code` word-wrap

### LINKED ISSUE #834: Markdown tables expand beyond window with no scrollbar

styfle:
This fixes Issue #834 and is alternative to PR #1027.

image

/cc @waylan

### DIFF
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -39,12 +39,13 @@
 }
 
 /*
- * Fix wrapping in the code highlighting
+ * Fix wrapping in the code
  *
  * https://github.com/mkdocs/mkdocs/issues/233
+ * https://github.com/mkdocs/mkdocs/issues/834
  */
 code {
-    white-space: pre;
+    white-space: normal;
     padding: 2px 5px;
 }
 
@@ -64,8 +65,11 @@ p code {
  * font size and padding.
  *
  * https://github.com/mkdocs/mkdocs/issues/855
+ * https://github.com/mkdocs/mkdocs/issues/834
+ * https://github.com/mkdocs/mkdocs/issues/233
  */
 pre code {
+  white-space: pre;
   display: block;
   padding: 12px;
   font-size: 12px;


waylan:
I don't think this is correct. We want

```
word-wrap: break-word;
white-space: pre-wrap;
```
for code spans (as I used in #1027). This is also what the `mkdocs` theme already uses for code spans.

If you don't do this, you will find many situations were inline code will be wrapped in places you do not want it to be (consider punctuation characters used in code, etc).

waylan:
To demonstrate, given the following HTML:

```html
<html>
<head>
<style>

p {
    width: 25em;
    border: 1px red solid;
}

.default code {
    white-space: pre;
}

.normal code {
    white-space: normal;
}

.prewrap code {
    word-wrap: break-word;
    white-space: pre-wrap;
}

</style>
<body>

<p class=default><code>a_very_long_variable_name=some_very_long_namepsace.with_a_very_long_function_name(and_a_very_long_keyword="some very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very long value")</code></p>

<p class=normal><code>a_very_long_variable_name=some_very_long_namepsace.with_a_very_long_function_name(and_a_very_long_keyword="some very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very long value")</code></p>

<p class=prewrap><code>a_very_long_variable_name=some_very_long_namepsace.with_a_very_long_function_name(and_a_very_long_keyword="some very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very long value")</code></p>

</body>
</html>
```

We get the following rendering:

screencapture

Clearly the third example is the one we want. Your example uses a long string of space separated words. As can be seen in my example, that part wraps fine in both your solution and mine, and therefore makes for a poor test. The real test is when we have code without spaces (which is more common in the real world). In fact, the example pointed to in [this comment][1] on #834 consists of a bunch of long strings without spaces.

As an aside, I realize I wrapped the code in a `p` tag rather than a `table` in my example, but I wanted the most basic minimum possible. I added the red border to each paragraph to demonstrate when the code span is staying within the confines of its parent and when it is not.

[1]: https://github.com/mkdocs/mkdocs/issues/834#issuecomment-187518884

waylan:
Btw, one thing I do like about this fix is that the inline style is applied to all `code`  tags except for `pre code`. In the `mkdocs` theme, the inline style only applies to `code` tags inside `p` elements, which means it doesn't get applied inside list items, table cells, and various other non-p elements.

styfle:
Do you have an example input using markdown syntax that breaks using my change?

waylan:
With your fix, this table:

```markdown
|Test1|Test2|
|-------------|-------------|
|Hello|`a_very_very_very_long_variable_name=some_very_very_very_long_namepsace.with_a_very_very_very_long_function_name(and_a_very_very_very_long_keyword="some very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very very long value")`|
```

Renders as:

screenshot

Which is clearly still not correct.

waylan:
I have a working solution, but its king of hacky.

For code tags I have:

```css
code {
    white-space: pre-wrap;
    word-wrap: break-word;
    padding: 2px 5px;
}

pre code {
  white-space: pre;
  word-wrap: normal;
  display: block;
  padding: 12px;
  font-size: 12px;
}
```

And for the table I have:

table {
    table-layout: fixed;
}

td {
    overflow: hidden;
    max-width: 43.5em;
}
```

Which renders like this:

![screencapture](https://cloud.githubusercontent.com/assets/78846/17982408/aef8ca42-6ad5-11e6-8f42-5926beee80e3.png)

 And a narrow table was included to show we don't get any weirdness there. 

I also included a paragraph with a long code span in there to demonstrate that that works also. Weirdly, the code span in the paragraph works fine in multiple different configurations. The table is extra stubborn though. 

Without the `max-width: 43.5em` (which appears to be the maximum width the page content gets) on the `td` tag, the code does not get wrapped, no matter what styles we set on the `code` element.  The problem is, if the browser window is narrowed, that fails to shrink with the page and the problem of the code running long remains. I tried using percentages, but that seemed to have no effect. So, its a rather hacky solution.

**Comment by waylan** [comment]
Given the finickiness of the above, my inclination is to go with this for the table:

```css
table {
    width: 100%;
    overflow: auto;
    display: block;
}
```

Which just adds a scrollbar if the table is too wide. The problem then is the narrow tables:

scroll

If we could get rid of that weirdness with the border, everything would be fine (By the way, the table border is a non-problem in the `mkdocs` theme as it has no border).

I tried this:

```css
table.docutils {
    width: 100%;
    overflow: auto;
    display: block;
    border: none;
}

thead, tbody {
    border: 1px solid #e1e4e5;
}
```

Which almost works:

border

The problem there is the right side of the scrolled table is missing a border unless you scroll all the way to the right, but then its missing the left border. We need the border set on the table element.

waylan:
My test page is in this [gist][1]. I noticed that GitHub also has a missing right border on scrolled tables (and missing left border when scrolled), so I'm okay with it too.

I'm closing this in favor of #1037.

[1]: https://gist.github.com/waylan/f5503feb9177ffc3aa790561d8220598

styfle:
@waylan Very well documented, thanks.
I think following GitHub's rendering is a good choice.

lcherone:
`display: block;` breaks/prevents 100% width so width is kinda useless in both githubs and this templates tables..

PR #1037, opened 2016-08-25
Account for wide tables and wrap inline code spans.

waylan:
A better fix for #834. Also see discussions on PRs #1027 and #1034, of which this is a replacement.

This was tested with [this gist][1] as a sample test page.

[1]: https://gist.github.com/waylan/f5503feb9177ffc3aa790561d8220598

That page renders like this in the `readthedocs` theme:

rtd

And like this in the `mkdocs` theme:

mkdocs

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -166,6 +166,10 @@ better conform with the documented [layout].
   (#738).
 * Add a `--force` flag to the `gh-deploy` command to force the push to the
   repository (#973).
+* Improve alignment for current selected menu item in readthedocs theme (#888).
+* `http://user.github.io/repo` => `https://user.github.io/repo/` (#1029).
+* Improve installation instructions (#1028).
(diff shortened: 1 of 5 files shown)


waylan:
I updated this to remove the scrollbars from the `mkdocs` theme, which doesn't need them as the entire page scrolls. Besides, to get scroll bars on the table, a narrow table won't stretch across the page, which looks weird with the table styling used in the `mkdocs` theme (no borders, etc.). In contrast, the `readthedocs` table didn't stretch before this change was applied, so it doesn't need to now either (thus it gets scroll bars for oversize tables).

screenshot from 2016-08-25 21 17 23

waylan:
Oh, and the failing test is unrelated to this issue, so I'm ignoring it here. It's getting fixed in #1033.

styfle:
LGTM :+1:

Issue #699, opened 2015-07-27
Building hidden pages


jhamman:
Is it possible to tell mkdocs to build pages but not have them show up in the side bar?  If it is possible, an example of how to do this in the documentation would be great.

I asked this question here as well: https://groups.google.com/forum/#!topic/mkdocs/ILW9-SIdERs

keelii:
I have the same question 
mkdocs only build specified .md files in yml. Sometimes i use relative path to reference md file, but it not working for me .

if there is a `glob pattern` to  set, It would be better ^!^ : 

    mdFiles: docs/\*/\*.md

waylan:
No this is not possible as this time. I am not aware that it is on the road-map either, although I am not the one to make such a decision. I would suggest that if such a feature was to be implemented, it would use the page meta-data to define a page as "hidden" (perhaps by adding a `hidden: true` key-value pair to the meta-data for the page). Then, when building the nav, it would simply be skipped.

jhamman:
Thanks @waylan - Good to know.  

Any thoughts on how difficult this feature would be to add? The `hidden: true` api is exactly what I was envisioning.

robodude666:
I've ran across this same exact issue a few weeks ago, and adopted the same method that [Docker](https://docs.docker.com/) used for their docs (they recently adopted Hugo):

- Set the Page Title in the [mkdocs.yml](https://github.com/eolamey/docker/blob/212dfb45dec64b470307aec22623f2d80be6695f/docs/mkdocs.yml) file to a special value (they used \*\*HIDDEN\*\*).
- Update the [Template](https://github.com/docker/docs-base/blob/641e94642573c34fea7894441e0f3b5a2b7ba477/theme/mkdocs/nav.html) to not include a pages with the above special value.

It feels very hacky, but it's working well enough for me right now. The docs base I'm maintaining is far smaller than Docker's so it's OK for now. I can imagine if you have a larger documentation base it might get annoying to maintain.

@waylan From trying to figure out how to get around this limitation, and reading older issues people opened on this topic it doesn't seem to be something that plans to be supported by mkdocs; and I would agree. mkdocs is intended to be used for simple documentation. Having dozens of hidden pages to be used for links only is a step above simple. Personally, I'd rather see this handled via a extension once API support is added.

waylan:
According to a recent edit to the first comment on #689 this is addressed by the work there. That said, it is not clear to me how or when that work will be merged in. 

[...]

PR #1044, opened 2016-09-05
parse pages not including in pages@mkdocs.yaml

outmana:
As the pages not listed in pages@mkdocs.yaml but contained in anothor pages  will invoke a warning:

```
WARNING -  The page "hyperledger/index.md" contained a hyperlink to "hyperledger/chaincode.md" which is not listed in the "pages" configuration.
```

I create a patch for parsing the pages not including in pages@mkdocs.yaml.

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -259,6 +259,32 @@ def build_extra_templates(extra_templates, config, site_navigation=None):
         utils.write_file(output_content.encode('utf-8'), output_path)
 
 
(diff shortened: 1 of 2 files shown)


waylan:
As per the discussion in #699, this sort of feature has been punted to the Plugin API to be implemented as a plugin (see #206). Additionally, if appears to add pages to the nav, which it should not do. See [this comment][1] for the sorts of implementations we would be looking for (as a plugin).

[1]: https://github.com/mkdocs/mkdocs/issues/699#issuecomment-200857096

PR #1057, opened 2016-09-30
fixed link for community docs and theme options

sfphoenix:
Removed indent on initial note about where to find MkDocs community themes and added bullet points to lay out theme building choices.

### DIFF
--- a/docs/user-guide/custom-themes.md
+++ b/docs/user-guide/custom-themes.md
@@ -6,12 +6,15 @@ A guide to creating and distributing custom themes.
 
 !!! Note
 
-    If you are looking for third party themes, they are listed in the MkDocs
-    [community wiki](https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes). If
-    you want to share a theme you create, you should list it on the Wiki.
+If you are looking for third party themes, they are listed in the
+[MkDocs community wiki](https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes). If 
+you want to share a theme you create, you should list it on the Wiki.
 
-When creating a new theme, you can either follow the steps in this guide to
-create one from scratch or you can download the `mkdocs-basic-theme` as a
+When creating a new theme, you have a few options:
+- You can follow the steps in this guide to
+below to create one from scratch
+
+- You can download the `mkdocs-basic-theme` as a
 basic, yet complete, theme with all the boilerplate required. **You can find
 this base theme on [GitHub](https://github.com/mkdocs/mkdocs-basic-theme)**.
 It contains detailed comments in the code to describe the different features


docs/user-guide/custom-themes.md
```diff
@@ -6,12 +6,15 @@ A guide to creating and distributing custom themes.
 
 !!! Note
 
-    If you are looking for third party themes, they are listed in the MkDocs
```
waylan: The indentation is intentional here. If you notice, the previous line contains the three exclamation points followed by an Admonition type (`!!! Note`). This is using the [Admonition Extension](https://pythonhosted.org/Markdown/extensions/admonition.html) syntax and the indented paragraph is the body of the note. Your change breaks it. If it helps, the Admonition Extension is enabled in our [config file](https://github.com/mkdocs/mkdocs/blob/master/mkdocs.yml#L27).

docs/user-guide/custom-themes.md
```diff
@@ -6,12 +6,15 @@ A guide to creating and distributing custom themes.
...
-    [community wiki](https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes). If
-    you want to share a theme you create, you should list it on the Wiki.
+If you are looking for third party themes, they are listed in the
+[MkDocs community wiki](https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes). If
```
waylan: The trailing white-space here is causing the [MarkdownLint](https://travis-ci.org/mkdocs/mkdocs/jobs/163943352#L205) test to fail, as does the [long line](https://travis-ci.org/mkdocs/mkdocs/jobs/163943352#L206).

docs/user-guide/custom-themes.md
```diff
@@ -6,12 +6,15 @@ A guide to creating and distributing custom themes.
...
-    If you are looking for third party themes, they are listed in the MkDocs
-    [community wiki](https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes). If
-    you want to share a theme you create, you should list it on the Wiki.
+If you are looking for third party themes, they are listed in the
+[MkDocs community wiki](https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes). If 
+you want to share a theme you create, you should list it on the Wiki.
 
-When creating a new theme, you can either follow the steps in this guide to
-create one from scratch or you can download the `mkdocs-basic-theme` as a
+When creating a new theme, you have a few options:
```
waylan: I'm not sure I see the value in making this a list. I think it was fine as-is, but a list works as well. However, your list's formatting is a [mess](https://travis-ci.org/mkdocs/mkdocs/jobs/163943352#L197). No blank line before the first item, poorly hard wrapped lines (let's keep line length consistent),...

waylan:
@sfphoenix, thanks for your contribution. However, as per my inline comments, I don't see any value in the changes here. In fact, it seems to make things worse (and breaks things). The only possible value was moving "MkDocs" to be part of the link label for the link to the wiki, but I'm "meh" on that change. I'm inclined to close this. But maybe an update could change my mind.

docs/user-guide/custom-themes.md
sfphoenix: Yeah, I appreciate the break down. Gonna use a linter and test for next time!

PR #1061, opened 2016-10-03
Optimize search by building index locally.

waylan:
This builds an actual lunr.js search index via Node.js in addition to a JSON
array of page objects previously used to build the index client side. Note that
both are needed as the array of page objects is used to display search
results.

If Node fails (not installed or errors out), then fallback to old behavior
by creating an empty index. If the client (browser) receives an empty index,
then it builds the index from the array of pages as previously, which is
fine for most (small) sites. Large sites will want to make sure Node is
installed to avoid the browser from hanging during index creation.

Partially addresses #859.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -190,6 +190,22 @@ better conform with the documented [layout].
 
 [layout]: ../user-guide/writing-your-docs/#file-layout
 
+#### Support for Pre-Built Search Index
+
+Previously, on each page load, the search index would need to be rebuilt from a
+JSON file of all the data by the client. While this was okay for small sites, it
+created a real burden on large sites (with hundreds of pages) and has been known
+to cause the entire browser to hang while the index is being built. MkDocs will
+now attempt to pre-build the index as a seperate JSON file, which is loaded by
+the client. Note that this new behavior only works if [Node.js] is installed on
+the build machine. If the build fails for any reason, MkDocs falls back to the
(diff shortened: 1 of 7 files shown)


waylan:
I've decided that there is no reason to not commit this part of the solution outlined in #859 (see [this comment][1]). The rest can come when we do the refactor later (webworkers will require a refactor anyway, so we might as well do that with the move to a plugin). For most users (who won't have Node.js installed) nothing will change. But for those with have (or need) this feature, they can start using/testing it now and it can even be better in the next release.

[1]: https://github.com/mkdocs/mkdocs/issues/859#issuecomment-251176245

waylan:
Oh, I forgot PyExecJS doesn't work on Python 2.6. Unless we want to drop Py26 now, this will have to wait till the next release.

micahhausler:
Any update on this? This affects us and we'd love to see faster search

waylan:
Work on search was delayed until the Plugin API was put in place. The Plugin API is now available in the 1.0.dev branch and the existing search feature has been broken out into a plugin . We intend to refactor search in that (or a new) plugin in the near future. Of course, as it is now a plugin, third parties could create their own plugin which offers a different set of defaults. The 1.0.dev branch will be available in the next major release; however feedback on the new API now would help move move development along.

waylan:
This is being closed in favor of the simpler solution in #1418. That is, the prebuilt index solution is simpler. In other respects #1418 is a complete refactor of search so we also get web worker support etc.

PR #1063, opened 2016-10-04
Use official SPDX license text

balupton:
Some of the formatting of the current license does not match the SPDX standard. I've updated the license text to match the SPDX standard, as used by tooling such as GitHub's license API and https://www.npmjs.com/package/spdx

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,26 +1,9 @@
-Copyright © 2014, Tom Christie. All rights reserved.
+BSD 2-clause "Simplified" License
+Copyright (c) 2014 Tom Christie. All rights reserved.
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
 
-Redistribution and use in source and binary forms, with or
-without modification, are permitted provided that the following
-conditions are met:
+1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
 
-Redistributions of source code must retain the above copyright
-notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


waylan:
Why did you remove the hard wrapped lines? Is that part of the spec? If so, then the spec is severely broken. I never ever allow anything without hard line breaks and never have for many many years. I can't accept this as-is. 

Of course, this is my personal opinion and does not represent the view of the project or the other devs. They may not feel so strongly about this, or even care.

balupton:
Perhaps take it up with SPDX? Here is the license according to them: https://spdx.org/licenses/BSD-2-Clause.html#licenseText

waylan:
But does the spec **require** no hard line breaks? I'm guessing not and don't care enough to bother looking it up.

balupton:
As you wish, happy for you to close. The reason for filing was just to make it easier for people in the future to identify exactly which BSD license it was, as the formatting currently is strange as the conditions read like paragraphs rather than independent clauses. Honestly, the amount of the time this pull request saves is negated by the amount of time discussing this pull request takes. So whatever people want to do.

waylan:
So I got curious and checked the [spec][1]. Every change made here is either ignored (license name) or treated as interchangeable with the original (whitespace = space, (c) = &copy; etc). As the formatting of the original is much better, I'm closing this.

As an aside, Github is correctly parsing the file as-is and listing the licence correctly, so the changes are clearly not needed.

[1]: https://spdx.org/spdx-license-list/matching-guidelines

balupton:
@waylan cool, thanks for looking into this further

Issue #1065, opened 2016-10-06
Support for Image Lightboxes


martnst:
Hi there, 
we are currently using [MDwiki](http://dynalon.github.io/mdwiki/#!index.md) which looks very very similar to MKDocs, but would like to change to MKDocs as it seems to be a more active project and seems to have more features. 

Unfortunately it seems that MKDocs does not have a lightboxes built in as MDwiki does, see [here](http://dynalon.github.io/mdwiki/#!quickstart.md#Images). As we have screenshots in our docs it is crucial to us that the reader can bringt them fullscreen to see all details while still having the images aligned nicely in the text per default. 

**Is there any theme that supports lightboxes or plan to include it in a future version?**

waylan:
I haven;t checked, but perhaps a third party theme offers such support. Check the [wiki][4] for a list.

I see no reason why we would include that specifically in the default theme. However, themes can certainly be modified. In fact, in the next release (0.16) there will be much more support for theme customization (see a [preview][0]). Until then, there is always the [exta_css][1] and [extra_javascript][2] config settings for you to provide your own styling and JavaScript for your lightboxes. You just need to write your HTML then. Of course, Markdown accepts and passes through raw HTML, or you could create (or find an existing) [Markdown Extension][3] to generate the HTML from some simple Markdown markup. And later, when the Plugin API becomes available (see #206), then this could possibly be all wrapped up in a plugin.

[0]: https://github.com/mkdocs/mkdocs/blob/master/docs/user-guide/styling-your-docs.md#customizing-a-theme
[1]: http://www.mkdocs.org/user-guide/configuration/#extra_css
[2]: http://www.mkdocs.org/user-guide/configuration/#extra_javascript
[3]: http://www.mkdocs.org/user-guide/configuration/#markdown_extensions
[4]: https://github.com/mkdocs/mkdocs/wiki/MkDocs-Themes

waylan:
I just occurred to me that there are a collection of bootstrap themes in  [mkdocs/bootstrap][1]. And with a quick search, a [Bootsrap Lightbox plugin][2] exists. You might be able to get that working with less effort.

[1]: https://github.com/mkdocs/mkdocs-bootstrap
[2]: https://ashleydw.github.io/lightbox/

waylan:
Also, the default theme (`mkdocs`) and all the themes in [mkdocs/mkdocs-bootswatch] are built with bootswatch, which is a bootstrap wrapper.

[...]

PR #1067, opened 2016-10-06
Document using theme_dir and template blocks togeather

waylan:
The first commit adds a section to the docs about customizing an existing theme which ties together a JS lib added in the `theme_dir` and a template block to link to the lib from HTML. This is necessary to get a lib to work, so we need to document it.

The second commit adds a link to a more complete example in a standalone repo. Note that the link assumes that the repo currently at [waylan/mkdocs-theme_dir-example](https://github.com/waylan/mkdocs-theme_dir-example) is assumed to have been moved to `mkdocs/mkdocs-theme_dir-example`. I don't have permission to add repos, so someone else will need to do that. 

If it is decided that we don't want to host that example internally, then the second commit should be removed before merging this.

This is related to the discussion in #1065.


waylan:
I've updated this PR by removing the link to the example repo and added that link in the wiki. This documentation needs to get in and it doesn't make sense to keep waiting. After all, that is what the wiki is for.

PR #1075, opened 2016-10-17
Consistently use "dirty" to indicate dirtyreload server.

### LINKED ISSUE #1074: Can't get option --dirtyreload to work?
In reference to issue #990 I am trying to get a dirty reload to work. Since I have a site with almost 300 pages a full rebuild takes about 4 minutes.
Testscenario:
Use mkdocs 1.16
1. test with `mkdocs serve`, on change of a page the full site gets rebuilt
2. test with `mkdocs serve --dirtyreload`, on change of page nothing happens?
3. test with `mkdocs serve --dirtyreload --livereload`:
the output correctly states that mkdocs is `watching changes`. However on a change the full site is rebuilt and not only the changed file.

What I find strange is that when I only use `--dirtyreload` that I don't see a message taht mkdocs is `watching changes`

Is this an issue with how mkdocs watches for changes when doing `--dirtyreload` or am I doing something wrong. I have to add that I'm running using my own docker container, that's why I explicitly tried whether reload does indeed work without `--dirtyreload`

regards,
Rick

waylan:
Also removed the short option `-d` from the `--dirtyreload` flag to match
the other options. Fixes #1074

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -106,7 +106,7 @@ def cli():
 @click.option('-e', '--theme-dir', type=click.Path(), help=theme_dir_help)
 @click.option('--livereload', 'livereload', flag_value='livereload', help=reload_help, default=True)
 @click.option('--no-livereload', 'livereload', flag_value='no-livereload', help=no_reload_help)
-@click.option('-d', '--dirtyreload', 'livereload', flag_value='dirty', help=dirty_reload_help)
+@click.option('--dirtyreload', 'livereload', flag_value='dirty', help=dirty_reload_help)
 @common_options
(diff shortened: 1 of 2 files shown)


rickpeters:
Hi Waylan,

I pulled the new code and tested. I assume I now have to use `--dirty`? I still get the message `option --dirty unknown`...

However when I use `--dirtyreload` I am very happy :-) to see that it now works and the changed page gets rebuilt in no time at all. Thanks very much!

regards,
Rick

waylan:
@rickpeters the use of the string "dirty" is used internally. The code was expecting "dirty" in some locations and "dirtyreload" in others, but only passing in the string "dirty". However that is simply an implementation detail. The command line flag is still called `--dirtyreload`.

Issue #814, opened 2016-01-30
Update mkdocs.org automatically on a release


d0ugal:
Travis now [releases to PyPI](70adae9ad884631598e6718c115a38b9f6f18915) for us, it would be great if it could then also run mkdocs gh-deploy for us too.

Somebody recently figured out how to do this, we can maybe lear from it: http://dragplus.com/post/id/33416269 (it looks somewhat complicated, it would be great if we could make this easier for people).

waylan:
Your first link appears to be bad.

In any event, I don't get this. Why does Travis need to deploy? Why can't you run a couple commands locally to deploy? I realize it is sometimes a problem when you forget which options you need each time. But a simple script which provides those options as defaults works fine. That's why I created a [makefile][1] for Python-Markdown. The two commands `make deploy` and `make docs` takes care of everything (granted, my script does not currently upload docs - but that is for an unrelated reason). And if I forget the subcommands, `make help` reminds me. Just by adding this, deploying a new release is no longer a chore and I'm more inclined to do it more often. And it is a lot easier to develop that the hoops you have to jump through to get Travis to deploy.

[1]: https://github.com/waylan/Python-Markdown/blob/md3/makefile

waylan:
I should note that if some additions were made to MkDocs which made it easier to deploy from a CI Server, that would be beneficial. In that case, it might make sense for MkDocs to eat its own dog food. But I doubt we would want to include a Travis specific script. Perhaps a `deploy` script which only deploys if an environment variable is set (perhaps `MKDOCS_DEPLOY`). After all, different users will have different criteria to determine when a deploy should happen and may be using different CI services. Users could then set up whatever then wanted as criteria and set the environment variable to `True` or `False`. Then, when the deploy command is called, it will act on that variable. Perhaps adding a flag to the current deploy command would be sufficient:

    mkdocs gh-deploy --if MKDOCS_DEPLOY

d0ugal:
> In any event, I don't get this. Why does Travis need to deploy?

Simply, I want to automate as much as possible to remove dependency on me. It isn't much, but the release process is knowldge in my head and nowhere else, if that is all handled by travis and I (or possibly somebody at some point) just needs to push a git tag, then great!

Also, I would also like to  see if we could have the docs for master be built and available at mkdocs.org/master/ or similar. So that would be updated on each commit, then a tag would update the main docs.

d0ugal:
[...]

PR #1082, opened 2016-10-22
Added deploy-docs script to build and push to gh-pages on release

laf:
Fixes: #814

Sorry it took so long.

For this to work you need to set an environment variable in Travis: https://travis-ci.org/mkdocs/mkdocs/settings called GH_TOKEN with a personal access token. This is then used within the script from the travis build environment but they hide it from output.

I hope this is what you had in mind, if not, let me know and I'll make the necessary changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ script:
 - tox
 before_install: pip install codecov
 after_success: codecov
+after_deploy: test $TRAVIS_BRANCH == "master" && bash ./deploy-docs.sh
 deploy:
   provider: pypi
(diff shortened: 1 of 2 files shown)


laf:
No reason at all  other than I've never personally used it.

If it's as simple as running that over git push then I can update the script if you like. I can test it on my own travis and see.

waylan:
I expect that you can replace your entire custom script with one line in the `travis.yml` file. No need for a separate script at all.

waylan:
Closing this as we are no longer using Travis. See #2223.

PR #1088, opened 2016-11-08
Remove link added twice

pascalberger:
Remove a link which was added twice to the markdown file


waylan:
Good catch. Thanks. This is another issue that mivok/markdownlint#137 would have catch if implemented.

PR #1100, opened 2016-11-23
Allow pypy3 failures for now

d0ugal:
It seems that the pypy3 version is old and pip no longer supports it. We can maybe remove this entirely, but for now lets keep it but turn the build green again.


waylan:
This makes more sense than removing the tests completely. This way if the issue ever gets fixed we'll be more likely to notice.

d0ugal:
Agreed. Thanks.

PR #1105, opened 2016-12-01
Only "load" a theme when it is explicitly requested by the user.

### LINKED ISSUE #1104: Broken installed themes should not break MkDocs
As highlighted in [this thread][1], a broken (not importable) installed theme will break MkDocs even if the theme is not being used by the project.

In [the CLI][2] we call [`utils.get_theme_names`][0] and assign the result to a global variable to display in the CLI help. That function wraps [`utils.get_themes`][3], which iterates over the entry points and loads each one. If any one of the modules pointed to by an entry point fails to import, the entire thing crashes. We should fail more gracefully, perhaps with an warning that a theme has a problem and name the specific theme if possible. I suggest a warning unless the theme is the same theme actually being used by MkDocs, in which case an error should be raised, but that would be handled elsewhere anyway.

[0]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L406
[1]: https://groups.google.com/forum/#!topic/mkdocs/qEYkyHxYGAw
[2]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/__main__.py#L78
[3]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L379

waylan:
Previously, all installed themes were "loaded" each time the CLI
was run. Now, only the names are obtained unless the user explicitly
requests a theme by name (in user config for via command line option).
A broken third party theme can no longer "break" a project in which
is is not being used.

Fixes #1104.

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -285,7 +285,7 @@ class ThemeDir(Dir):
 
         package_dir = os.path.abspath(
             os.path.join(os.path.dirname(__file__), '..'))
-        theme_dir = [utils.get_themes()[config['theme']], ]
+        theme_dir = [utils.get_theme_dir(config['theme']), ]
         config['mkdocs_templates'] = os.path.join(package_dir, 'templates')
 
         if config['theme_dir'] is not None:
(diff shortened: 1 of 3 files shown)


waylan:
I think maybe we need to add some better validation (and graceful errors) for the theme requested by the user. I'm going to hold off merging this until I can add that.

waylan:
I've decided the way to improve validation is via a refactor of themes (which had been discussed elsewhere, although I can't find it right now) where we create a `Theme` object for each theme. therefore, I'll merge this as-is.

Issue #826, opened 2016-02-16
utf-8 search support


coffeetalkh:
Hey, we are working on a Persian version of a mkdocs theme, everything works fine but we have a problem on search contents with Persian contents, when i search the expression nothing found, i also test it with Arabic, Chinese and Hebrew and all of theme have problem with this case. any solution?

d0ugal:
Unfortunately search wont work well in non-latin character sets at the moment. We use [lunr.js](http://lunrjs.com/) to provide the search functionality, I believe it can work with them but I am not sure what is involved in making that happen.

It is possible for themes to overwrite and completely replace the search implementation. To do this, themes would need to at least replace [this file](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/assets/search/mkdocs/js/search.js) (or modify the theme so that file isn't loaded) by including a javascript file in the theme with the path `mkdocs/js/search.js`. However, as I said at the top, I have no experience in making search work for different character sets, so I don't know how difficult it will be.

We could also consider adding a setting to disable search. While this isn't ideal, it would be better to turn search off than have it in a broken state.

I don't think there is a good, easy solution for this at the moment. :disappointed:

waylan:
There is also some related discussion in #859.  Apparently, non-English search indexes are extremely large as lunr.js only "excludes" insignificant English words from the index by default. [MihaiValentin/lunr-languages][1] appears to be an add-on to lunr.js which adds support for various other languages including Japanese and Russian (both use non-latin alphabets). Unfortunately, at this time, Persian,  Arabic, Chinese and Hebrew are not on the list of supported languages. Although there are currently outstanding requests for [Arabic][2] and [Chinese][3] to be added. If you would like to have support added for a certain language, I would suggest starting there. If/when, mkdocs adds language support, then your language will work if its supported by lunr.js.

[1]: /MihaiValentin/lunr-languages
[2]: /MihaiValentin/lunr-languages/issues/9
[3]: /MihaiValentin/lunr-languages/issues/17

netroby:
https://github.com/mkdocs/mkdocs/blob/master/mkdocs/commands/build.py#L223

json dump to utf-8?

waylan:
@netroby, that is some old legacy (now deprecated) code which will be removed in the next major release. It has nothing to do with the current search implementation or any future updates to search.

waylan:
[...]

PR #1109, opened 2016-12-13
Remove dependency `lunr.js`

kurubot:
As we all know, `lunr.js` makes search indexes from given data. However, it only supports English. I just modified `search.js` to search without `lunr.js`, so it can support many other languages (Chinese, Japanese, etc).


waylan:
So we loose stemming and a stop word filter, among other features not included in a basic JavaScript text `search`. This feels like one step forward, five steps back. As noted in #826, lunr.js can be made to support other languages.

 In fact, I intend to add that support. It hasn't happened yet because search will be moved to a plugin (see [this comment][1]). When that happens, lunr.js will be configurable, or... you can completely replace it with whatever you want without modifying the core.

[1]: https://github.com/mkdocs/mkdocs/issues/859#issuecomment-251176245

waylan:
If you're concerned that you don't want to wait for all of that to happen, the plan is to have that done by the next major release. In other words, this wouldn't make it to a release any sooner that that. And if you really want it now, then use [theme customization][1] to get it now.

Given the above, I'm closing this.
[1]: http://www.mkdocs.org/user-guide/styling-your-docs/#using-the-theme_dir

waylan:
I understand your concern, but we are not going to introduce new features or backward incompatible changes before the next major release. As per the current plan, there will be no interim release this could be included in. Yes, we may do a "bug-fix release" for two, but this would not be something we would classify a "bug fix." In fact, for most of our current users, it would be a breaking regression.

PR #1124, opened 2017-01-11
Respect order of config extensions

ecarreras:
In some particular case we need to reorder of builtin extensions,
and all the markdown extensions are putted at the `_end`. And the
builtin extension are execute before our extensions.

With this change if a builtin extension is defined in the config
then is removed from builtin and added in the configuration order

In particular we need the `toc` extension after our [i18n extension](https://github.com/gisce/markdown-i18n)
to get the menu translated.

Then I don't know if have sense to execute `utils.reduce_list(self.builtins + extensions)` if they are removed from `self.builtins` if come from config.

If there is another way to accomplish that let me know.

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -525,6 +525,8 @@ class MarkdownExtensions(OptionallyRequired):
                 if len(item) > 1:
                     raise ValidationError('Invalid Markdown Extensions configuration')
                 ext, cfg = item.popitem()
+                if ext in self.builtins:
+                    self.builtins.remove(ext)
                 extensions.append(ext)
                 if cfg is None:
                     continue
@@ -533,6 +535,8 @@ class MarkdownExtensions(OptionallyRequired):
                                           "Extension '{0}'.".format(ext))
                 self.configdata[ext] = cfg
             elif isinstance(item, utils.string_types):
+                if item in self.builtins:
+                    self.builtins.remove(item)
                 extensions.append(item)
             else:
                 raise ValidationError('Invalid Markdown Extensions configuration')


facelessuser:
There are ways to hack it.

You register your extension in extendMarkdown so the reset hook will get triggered.  It always triggers before the pass.  Track with a flag whether you've modified things so you only do it once.  And perform hack by moving your entry.  I can't say its pretty or proper, but I've done it myself.  If I've made a mistake, you'll have to debug.

I know Python Markdown 3 plans to have a better way to manage such things, but it will probably be a while before that releases.

```
    def __init__(self, *args, **kwargs):
        self.configured = False
        super(MyExtension, self).__init__(*args, **kwargs)
    
    def extendMarkdown(self, md, md_globals):
        md.treeprocessors.add("my-extension", MyTreeProcessor(md), "_end")
        md.registerExtension(self)

    def reset(self):
        if not self.configured and 'toc' in self.md.treeprocessors.keys():
            self.configured = True
            self.md.treeprocessors.link("my-extension", "<toc")
```

facelessuser:
You can see if everyone did this, you'd be right back where you started.  I would use this only as a last resort.

facelessuser:
It also might be possible to always attempt to insert before toc, and assuming it throws an exception, catch it and then go for `_end`.  I don't recall if it throws an error, but I would be surprised if it didn't.  I think that would be cleaner.

ecarreras:
Thank you for the tips @facelessuser I'll follow your advice

ecarreras:
@facelessuser It worked like a charm :tada: https://github.com/gisce/markdown-i18n/pull/31

facelessuser:
@ecarreras , glad I could help.

waylan:
For the record, I would not have accepted this PR. An extension should *never* need rely on the order it was added to work. The solution is to always account for this in the extension itself. That said, the current system is less than ideal for this, which is why we are making changes in version 3.0

facelessuser:
> An extension should never need rely on the order it was added to work. 

@waylan  in general I understand the desire for order not to be important, but as someone who can roll their own extensions.

It is more difficult for the average user as they may not be the author of either conflicting extension, and may not have the skills to roll an extension to sort out the conflicts.  Being able to sort your extensions to move on is generally an easy short term solution for user facing these issues until the extension developers can sort out the conflict.

So I can understand how in the future a *user*, might be frustrated and request this again.  I'm neither for or against this as it doesn't really affect me much , just highlighting the difficulties that a *user* may face.  In either case, 3.0 will make it harder to hit these initial conditions, but they could still occur, but they will be much easier for a developer to rectify without the hackery above.

Issue #1127, opened 2017-01-20
Run search only for query having 3 and more characters


davidhrbac:
We have quite a large search_index.json - 2.2MB. The search is very slow. How can we limit the search to start only for queries having 3 and more characters?

waylan:
Is search slow all the time or only the first time after a new page is loaded?

If it is the former, then this is a duplicate of #859. If the latter, then this will likely be addressed by that issue anyway as I intend to do a complete refactor of search allowing users to customize the features available in the underlying library. The holdup is that search will be moved to a Plugin and we need to finish the Plugin API first (see #206). Either way, I'll leave this open to remind me to possibly include this as one of the config options.

davidhrbac:
@waylan the first is definitely slow. Anyway there's no point to let JS search the large base for single or two characters. It might return very large set (hundreds, even more). Thanks.

PR #1128, opened 2017-01-20
Compress search_index.json

davidhrbac:
This is a quick solution to compress search_index.json. It replaces duplicated characters. u00a0 is Unicode nowrap space. Also indentation is disable. I could shrink file from 2248293 Bytes to 1752850 Bytes. Might help with search download and start-up. Links to #1127.

### DIFF
--- a/mkdocs/search.py
+++ b/mkdocs/search.py
@@ -35,6 +35,16 @@ class SearchIndex(object):
         A simple wrapper to add an entry and ensure the contents
         is UTF8 encoded.
         """
+        # Sanity check to compress JSON
+        text=text.replace(u'\u00a0', ' ')
+        text=text.replace('  ', ' ')
+        text=text.replace('\n\n', ' ')
+        text=text.replace('  ', ' ')
+        text=text.replace('\n ', ' ')
+        text=text.replace('  ', ' ')
+        text=text.replace(' \n', ' ')
+        text=text.replace('  ', ' ')
+        text=text.replace('  ', ' ')
         self._entries.append({
             'title': title,
             'text': utils.text_type(text.strip().encode('utf-8'), encoding='utf-8'),
@@ -90,7 +100,7 @@ class SearchIndex(object):
         page_dicts = {
             'docs': self._entries,
         }
-        return json.dumps(page_dicts, sort_keys=True, indent=4)
+        return json.dumps(page_dicts, sort_keys=True)
 
     def strip_tags(self, html):
         """strip html tags from data"""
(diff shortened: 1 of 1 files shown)


waylan:
This looks like a good change. Thank you. However, we are not planning on releasing any new version (except perhaps a bug-fix release) before 1.0 which will have the changes planned in #859. Therefore, I'm inclined to hold off on this now and incorporate these changes into that refactor.

davidhrbac:
OK, no rush, I can have it within our CI made with sed like this:

```bash
#!/bin/bash
sed 's/\\u00a0/ /g' -i "$@"
sed 's,\\n\\n, ,g' -i "$@"
sed 's,\\n , ,g' -i "$@"
sed 's, \\n, ,g' -i "$@"
sed -e 's/^[ \t]*//' -i "$@"
```

mkdocs/search.py
```diff
@@ -90,7 +100,7 @@ def generate_search_index(self):
         page_dicts = {
             'docs': self._entries,
         }
-        return json.dumps(page_dicts, sort_keys=True, indent=4)
+        return json.dumps(page_dicts, sort_keys=True)
```
d0ugal: See the compact encoding in the docs, we should consider doing that:

    json.dumps([1,2,3,{'4': 5, '6': 7}], separators=(',',':'))

From https://docs.python.org/2/library/json.html

I don't really know if it has any drawbacks or limitations.

facelessuser: It shouldn't have any drawbacks as it will still be valid JSON.

d0ugal: Yup, I just read the Python docs. Seems good!

waylan:
As a reminder, we may also want to explore creating a gzipped copy of the index. Both files should exist side-by-side and properly configured servers will serve the gzipped version.

coliff:
This PR is a great help. I was able to decrease the size of my search_index.json from 1250 KB to just 980 KB.

coliff:
I merged the changes made by @davidhrbac on my local installation to search_index.py (https://github.com/mkdocs/mkdocs/blob/master/mkdocs/contrib/legacy_search/search_index.py) and it works well. Could this PR be updated? (If you'd like me to open a new PR with the change I can).

waylan:
@coliff, the entire search function is scheduled for a complete rewrite. Its likely that the rewrite will actually start from scratch. Therefore any changes to the existing code will likely be scrapped. This PR remains open as a means to remind us to include a similar feature in the rewrite.

As an aside, now that search is contained within a plugin, anyone is free to fork it (within the confines of the license) and make any changes they desire. For that matter, someone could build their own third-party search plugin which is  better than and removes the need for us to maintain the one we have now.

Issue #1040, opened 2016-08-28
The dev server doesn't serve template provided error pages.


waylan:
While working on #1039 I noticed that the dev server returns its own (unstyled) error page, rather than the error page provided by the template. It would be nice if that worked out-of-the-box.

As an aside, I realize that in various environments, the user will need to configure their production server and point it to the error page. However, users they might not even know a theme comes with an error page if they don't ever see it during development. The dev server should be serving a full, working site.

PR #1141, opened 2017-02-23
[ISSUE 1040] The dev server doesn't serve template provided error pages

akhilputhiry:
Fix for https://github.com/mkdocs/mkdocs/issues/1040

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -1,12 +1,33 @@
 from __future__ import unicode_literals
+
 import logging
 import shutil
 import tempfile
 
+from os.path import isfile, join
 from mkdocs.commands.build import build
 from mkdocs.config import load_config
 
 log = logging.getLogger(__name__)
+
+
+def _get_handler(site_dir):
+
+    from tornado.web import StaticFileHandler
+    from tornado.template import Loader
+
+    class WebHandler(StaticFileHandler):
+
+        def write_error(self, status_code, **kwargs):
+
+            if status_code in (404, 500):
+                error_page = '{}.html'.format(status_code)
+                if isfile(join(site_dir, error_page)):
+                    self.write(Loader(site_dir).load(error_page).generate())
+                else:
+                    super(WebHandler, self).write_error(status_code, **kwargs)
+
+    return WebHandler
 
 
 def _livereload(host, port, config, builder, site_dir):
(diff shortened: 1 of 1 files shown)


akhilputhiry:
@waylan Your thoughts..

waylan:
> But I don't think we can fix it for `_livereload` since the handler being used resides in the livereload package.

And this is why a fix hasn't been committed yet. I'm not comfortable with this only working for the less-feature-full server. It should work for both or not at all. I have not yet taken a close look at what would be required to make livereload work.

akhilputhiry:
@waylan I have found a way for livereload also. please check.

waylan:
Awesome! One question: Does this break on older versions of LiveReload? If so, do we need to update the minimum versions we support? See [sepup.py#L63][1], [requirements/project.txt#L3][2], and [requirements/project-min.txt#L3][3].

[1]: https://github.com/mkdocs/mkdocs/blob/master/setup.py#L63
[2]: https://github.com/mkdocs/mkdocs/blob/master/requirements/project.txt#L3
[3]: https://github.com/mkdocs/mkdocs/blob/master/requirements/project-min.txt#L3

mkdocs/commands/serve.py
```diff
@@ -1,21 +1,49 @@
...
     # this fails, the --no-livereload alternative should still work.
     from livereload import Server
 
-    server = Server()
+    class LiveReloadServer(Server):
+
+        def get_web_handlers(self, script):
+            handlers = super(LiveReloadServer, self).get_web_handlers(script)
+            handlers = handlers.pop()
+            return [(handlers[0], _get_handler(site_dir), handlers[2],)]
```
waylan: This reads peculiarly. You assign the last item (via pop) in a list of handlers to the `handlers` variable, then you reference items in that last item via index. So `handlers` starts out as a list of lists? 

I'm not questioning that this is correct. But months/years later when someone comes along to work on this code, it may not be immediately obvious whats going on here. Maybe we should use two separate variables (rather than `handlers` for both), or include a comment explaining what is contained in `handlers` on line 42 and how that differs from the new `handlers` on line 43.

waylan: And on closer inspection I realize that the method returns a list of one tuple which is clearly the format returned by the method of the parent class. Its all clear enough. 

Although I wonder if this would be better:

```python
handlers = super(LiveReloadServer, self).get_web_handlers(script)
return [(handlers[0][0], _get_handler(site_dir), handlers[0][2],)]
```

akhilputhiry: Agreed

akhilputhiry:
@waylan I tested with `livereload==2.5.1` it works fine. I think all requirements file has specified livereload version 2.5.1, then why should we worry about older versions??

mkdocs/commands/serve.py
akhilputhiry: @waylan comments addressed

waylan:
I wasn't sure if Livereload released anything newer than what we specify. If so we might need to update to the current release. That all. But if it works with oldest release we already support, then its not a concern.

Issue #803, opened 2016-01-20
TOC scroll position jumps to top after clicking a link on long TOCs (RTD theme)


sbutcher:
Summary: Using ReadtheDocs theme, when your TOC is long and exceeds the browser height and a scrollbar is required, if you scroll down and click on a link, the scroll position jumps to the top again after opening your page. 

Intended behaviour: the position in the TOC is remembered for you to continue clicking through pages near the ones you are currently reading.

Example: go to http://sandboxdocs.readthedocs.org/ , and reduce your browsers height until you see the scrollbar. scroll down the left hand menu and choose one of the last links e.g. Web Service api.
You will note that instead of retaining the position in the TOC, the position has now jumped to the top again.

waylan:
Is there even any elegant way to resolve this? When a link is clicked, the entire page is reloaded, including the TOC. 

True, there are solutions like using frames (shudder) or AJAX. However, in each of those, the page is served in pieces and only the individual pieces are loaded. As MkDocs is a "static page generator" its primary feature is that each page is a static, complete page on the server. No piecemeal assembling of pages when the request is made. Therefore, the obvious solution is in direct opposition to the main feature of MkDocs. 

I feel like in the past we wouldn't even consider such a request. However, today, with the prevalence of non-static pages (especially single-page apps) on many web sites, this behavior has become "normal" and that is the only reason this wasn't rejected immediately. So how can we "fake it?"

I suppose a URL query string could work (`http://example.com/some/page?toc=5`), but do we really want that? First, that makes for ugly URLs. Then, some JavaScript would be required to scroll the TOC to position 5 (whatever that might mean).

Sure, you could use the URL hash/fragment (`http://example.com/some/page#someid`), but that should be reserved for the position on the main body of the page (rather than the TOC).

I suppose JavaScript could write data to a cookie (or some client side DB) and retrieve it later. But that feels wrong for a "static site" to me. And then when would you update the info? Constantly as the TOC is scrolled or only when a TOC link is clicked?

I'm not aware of any other ways to maintain information across pages which don't require more sophisticated server setups than "static servers."

d0ugal:
The information we need is already there - no need for cookies etc. The JavaScript on the ReadTheDocs theme would just need to scroll the left nav to the top of `li.toctree-l1.current`

d0ugal:
This hacky line isn't too far off, but the offset is wrong. I'm not sure how we get the offset within the div.

```JavaScript
[...]

PR #1156, opened 2017-03-10
Updates to mkdocs ReadTheDocs theme

SquareWaffles:
This Pull Request addresses three issues:  1) adjusts TOC scrollbar so that the bottom is not covered by the 'Next' navigation button; 2) fixes the 'GitHub' class name in the version.html file; and 3) incorporates @seanmadsen code posted on Jan 10 issue #803 (if accepted and merged can be marked as closed) which may address some concerns with issue #588.  Still needing someone to address the collapsible TOC headings discussed in issue #588.

mkdocs github update scrollbar

### DIFF
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -29,7 +29,7 @@
 .wy-nav-side {
     height: calc(100% - 45px);
     overflow-y: auto;
-    min-height: 0%;
+    min-height: 0;
 }
 
 .rst-versions{
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -37,6 +37,10 @@
 	 height: 45px;
 }
 
+.wy-nav-side.shift{
+	height: 100%;
+}
+
 /*
  * readthedocs theme hides nav items when the window height is
  * too small to contain them.
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -33,12 +33,12 @@
 }
 
 .rst-versions{
-	 border-top: 0;
-	 height: 45px;
+    border-top: 0;
+    height: 45px;
 }
 
-.wy-nav-side.shift{
-	height: 100%;
+.wy-nav-side .shift{
+    height: 100%;
 }
 
 /*
--- a/mkdocs/themes/readthedocs/css/theme_extra.css
+++ b/mkdocs/themes/readthedocs/css/theme_extra.css
@@ -22,10 +22,19 @@
  * area doesn't scroll.
  *
  * https://github.com/mkdocs/mkdocs/pull/202
+ *
+ * Builds upon pull 202 https://github.com/mkdocs/mkdocs/pull/202
+ * to make toc scrollbar end before navigations buttons to not be overlapping.
  */
 .wy-nav-side {
-    height: 100%;
+    height: calc(100% - 45px);
     overflow-y: auto;
+    min-height: 0%;
+}
+
+.rst-versions{
+	 border-top: 0;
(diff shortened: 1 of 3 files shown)


waylan:
At a quick glance this looks good. Will take a closer look when I have some time.

However, while working on #1155 earlier today I noticed that the upstream (Sphinx) theme does not even have a visible scrollbar for the side nav. See [this page][1] for example. Perhaps the correct thing here is to do the same rather than the scrollbar adjustment here.

[1]: http://docs.readthedocs.io/en/latest/getting_started.html

waylan:
At a quick glance this looks good. Will take a closer look when I have some time.

However, while working on #1155 earlier today I noticed that the upstream (Sphinx) theme does not even have a visible scrollbar for the side nav. See [this page][1] for example. Perhaps the correct thing here is to do the same rather than the scrollbar adjustment here.

[1]: http://docs.readthedocs.io/en/latest/getting_started.html

SquareWaffles:
Ah, I did not even notice that Sphinx does not have a scrollbar on the toc.  Further investigation into when the scrollbar appeared in the mkDocs RTD theme references Pull Request #202.  There wasn't any conversation as to if the scroll should be visible.  I agree with your other posts that the mkDocs RTD theme should resemble Sphinx as much as possible.  But would it be advantageous to implement minor improvements where we can?  Yes, I know that then rabbit holes into defining "minor improvements".

The issue with the absence of the scroll is that depending on your window size you may not know that there are more headings in the overflow.  For instance, when I opened the link you referenced, my default window size makes it appear that the toc ends after "Ethical Advertising".  Furthermore, some machines (mine included) do not have scrolling enabled, therefor we use the scrollbar and the up/down keys.  The auto focus is on the main content as it should be so I have no idea that there is an overflow in the toc when using the keys.  Only way I can scroll in the toc is to click somewhere that is not a link and then use they up/down keys.  If the scrollbar is visible (only when content exceeds the view), it then serves two purposes: 1) alters the reader that there is more content and 2) allows easy access to the content. 

So, this all gets back to, how far should we deviate from Sphinx?  If you want it as close to possible, then this pull request should be rejected as Sphinx does not scroll in the toc and does not auto focus the toc on load.  I can see the advantages of keeping this theme as close to Sphinx as possible and I can see the advantages of implementing minor improvements.  The decision is ultimately up to you and the direction you want to take.

waylan:
@SquareWaffles You make good points. I think for now we will leave things as-is, which means accepting these improvements.

Now regarding these changes, I took a closer look and I found one problem. When in "mobile mode" (narrow browser window), there is a blank space at the bottom of the nav bar:

screenshot composite

Note that to see this, you need to narrow the browser window until the nav bar disappears, then click the hamburger button in the top left corner to show the nav bar. In that case, the nav bar should go all the way to the bottom of the window as the next/previous bar is at the bottom of the main body rather than the bottom of the nav bar. Of course, this makes sense for someone browsing on their phone. Requiring them to use the hamburger button to get to next/previous would be ridiculous.

SquareWaffles:
Thanks for catching that @waylan.  The css has been updated and ready for re-testing.

waylan:
That doesn't seem to solve the problem for me (checked both firefox and chrome). Besides, the `shift` class is added to the nav `div` when activating the hamburger button, but it doesn't go away when the window resizes. So, for example, rotating a tablet from portrait to landscape would retain the same 100% height, but mobile mode would be deactivated --that is, if it worked at all, which it doesn't for me for some reason I can't quit make out. The browser doesn't seem to even acknowledge that the rule was defined, let alone have it override the previous rule.

In any event, I think this makes more sense:

```css
@media screen and (max-width: 768px) {
    .wy-nav-side {
        height: 100%;
    }
}
```

waylan:
I just cleaned up the commit history (swashed the fixes for Travis failures) and incorporated by fix above. @SquareWaffles thanks for your help on this. I'm closing this in favor of #1157 (don't worry, you are still listed as author in the individual commits).

PR #1170, opened 2017-03-16
Remove previously deprecated `json` command.

### LINKED ISSUE #481: Deprecate the JSON command
Now that #457 has landed we should have people use the JSON it creates and remove the JSON command which is otherwise a bit weird. 

Add a warning for 0.13
Louder warning in 0.14
Remove in 1.0

waylan:
Closes #481.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -29,6 +29,7 @@ TODO...
 
 ### Other Changes and Additions to Version 1.0.0
 
+* The deprecated `json` command has been removed (#481)
 * Support for Python 2.6 has been dropped (#165)
 
 ## Version 0.16.2 (2017-03-13)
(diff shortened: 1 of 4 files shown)


waylan:
I always found it odd there was two ways to generate JSON. This command and the search index. Finally, when removing this, I realized that this predated the search index and it made some sense. But, yeah, this could have been removed some time ago.

d0ugal:
well, it could, but it is still used by readthedocs - this change will block them from upgrading to 1.0, but that is fine I think - they could be updated to use the search index.

waylan:
Actually, I think a number of things will break for them. For example some of the template changes in 0.16. We left the old variables in with a warning but they will go in 1.0. There may be a few other things I don't recall off-hand. It's all been deprecated, so they have had time to adapt, so I'm not worried about it.

Issue #1168, opened 2017-03-15
Remove all deprecated code in prep for 1.0


waylan:
Just a reminder so we don't forget about it. There are a bunch of "TODO" comments in the code marking sections that need removed etc.

```
$ grep -n -r "TODO" *
mkdocs/commands/build.py:90:        # TODO: remove the rest in 1.0 as they are deprecated
mkdocs/commands/build.py:124:    # TODO: remove the rest in version 1.0 as they are deprecated
mkdocs/commands/build.py:138:        # TODO: remove the rest in version 1.0 as they are deprecated
mkdocs/commands/build.py:204:            # TODO: Remove this in version 1.0
mkdocs/commands/build.py:259:    # TODO: remove DeprecationContext in v1.0 when all deprecated vars have been removed
mkdocs/commands/build.py:296:    # TODO: end remove DeprecationContext
mkdocs/config/config_options.py:278:    TODO: This could probably be improved and/or moved from here. It's a tad
mkdocs/config/config_options.py:428:        # TODO: Remove in 1.0
mkdocs/config/config_options.py:472:        # TODO: delete deprecated `NumPages` class in version 1.0
mkdocs/config/defaults.py:36:    # TODO: remove this in version 1.0
mkdocs/config/defaults.py:96:    # TODO: delete deprecated `include_nav` and `include_next_previous` in version 1.0
mkdocs/legacy.py:44:    TODO: Remove in 1.0
mkdocs/tests/integration.py:12:TODOs
mkdocs/tests/legacy_tests.py:9:    # TODO: Remove in 1.0
mkdocs/utils/__init__.py:73:        # TODO: Remove this when external calls are properly cleaning up file
mkdocs/utils/__init__.py:332:    # TODO: Fix this, this is a hack. Relative urls  are not being calculated
```

Most of those are relevant here.

waylan:
We also have the `mkdocs-bootstrap` and `mkdocs-bootswatch` themes for which we [raise a special error][1] if they are not installed. Probably want to remove that. Also check that we never install them anywhere (perhaps in tests).

[1]: https://github.com/mkdocs/mkdocs/blob/a00fe577005e1198b24183fbb93cb8c31be694a2/mkdocs/config/config_options.py#L317

PR #1171, opened 2017-03-16
Remove deprecated code.

waylan:
See #1168.


d0ugal:
so good! :+1:

waylan:
Yeah, its nice to remove old code. Less to maintain and work around as we move forward.

PR #1173, opened 2017-03-18
Added base_url to config so that improper relative path to static files can be fixed in sub directories of mkdocs theme

akhilputhiry:
I was trying to fix https://github.com/tomchristie/django-rest-framework/issues/4758 which occurs due to relative path of static files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ htmlcov/
 mkdocs.egg-info/
 *.pyc
 .coverage
+.idea/
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -77,7 +77,7 @@ def get_global_context(nav, config):
     return {
         'nav': nav,
         # base_url should never end with a slash.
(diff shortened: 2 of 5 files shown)


akhilputhiry:
@waylan Your thoughts please

waylan:
Two things:

I was of the impression this was fixed in f5b5446b which is included in version 0.16. If not, could you provide a minimum example that still fails?

And finally, your proposed fix does not actually "fix" the problem. If we are generating incorrect URLs, then we should actually fix that. You can find a detailed analysis starting with [this comment][1] (and the few after it).   There should be no need to add a new config option. Everything you would need is already available from the existing options.

[1]: https://github.com/mkdocs/mkdocs/issues/77#issuecomment-242777289

akhilputhiry:
@waylan thanks. let me see.

akhilputhiry:
Things are working fine with latest release. closing this PR.

Thanks @waylan

PR #1174, opened 2017-03-20
Fix live reload

jeffreytolar:
Live reload (with `mkdocs serve`) is broken in 0.16.2; saving a file triggers a reload, but the page contents doesn't change because the server returns a 304 Not Modified.

I did a bit of testing, and it looks like efbc78922dc379cecc95beef8d7e1badc49959a2 (#1141) is the suspect commit. The `_get_handler` method introduced in that PR inherits from `tornado.web.StaticFileHandler`. livereload has a slightly customized `StaticFileHandler` which disables 304's: https://github.com/lepture/python-livereload/blob/master/livereload/handlers.py#L151-L153

This fixes the issue by using the correct `StaticFileHandler`

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -11,9 +11,8 @@ from mkdocs.config import load_config
 log = logging.getLogger(__name__)
 
 
-def _get_handler(site_dir):
+def _get_handler(site_dir, StaticFileHandler):
 
-    from tornado.web import StaticFileHandler
     from tornado.template import Loader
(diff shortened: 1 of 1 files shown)


waylan:
Awesome. Thanks. I noticed some weirdness the other day when I was working one something else. But of course, I didn't investigate because I was working on something else at the time.

mikeschinkel:
`/opt/box/mkdoc` <--- source for WPLib Box documentation

PR #1179, opened 2017-03-23
Add 404 page for RTD Theme

kylegregory:
The ReadtheDocs theme does not include a 404 page in the build output. I checked and the Sphinx theme doesn't have one either, but if wanted, here's a simple 404 page. It may have been excluded for a reason, but I couldn't locate any previous issue.

### DIFF
--- a/mkdocs/themes/readthedocs/404.html
+++ b/mkdocs/themes/readthedocs/404.html
@@ -1,0 +1,9 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+  <h1 id="404-page-not-found">404</h1>
+
+  <p><strong>Page not found</strong></p>
+
+{% endblock %}


waylan:
The ReadtheDocs service provides their own 404 page which is separate from and does not share the styling of the theme (see [an example][1]). I suppose an argument could be made that we should follow suite, but personally I don't buy it. 

Unless we get an objection, I'm inclined to accept this. However, this should be committed to the 1.0.dev branch rather than master, so just hitting the merge button won't work. I'll get to it in time.

[1]: https://docs.readthedocs.io/en/latest/foo

kylegregory:
@waylan I can change the base branch on the PR, if that will make your life easier.

waylan:
@kylegregory Yes, that would make it easier.

waylan:
Closed in favor of #1209.

PR #1201, opened 2017-04-06
Theme refactor

### LINKED ISSUE #1164: Add theme specific configuration.
Currently, in addition to building (Markdown) pages via `main.html`, we have hardcoded provisions for theme templates to build `404.html`, `search.html`, and `sitemap.xml`. However, there is no way for a theme to provide any additional pages (`sitemap.html` or Sphinx's `genindex.html`).

We do have the `extra_templates` config setting, but that specifically assumes the template files are in the `docs_dir`, not at a path in the Jinja environment (they go through `mkdocs.commands.build.build_extra_templates()` rather than `mkdocs.commands.build.build_template()`).

In the work on Plugins (#206), a plugin can alter the config and add an additional directory to the Jinja Environment. That same plugin can also add a hook which explicitly calls `mkdocs.commands.build.build_template()` with a template in its template dir (that's how I implemented search as a plugin). However, themes have no way to alter the config.

To allow something like this, we need some way to allow configuration for themes. I see a few possibilities:

1. The theme could include a config file which defines specific variables. For example an `extra_templates` variable could include a list of additional templates to pass through `build_template`.  As prior art, Sphinx themes provide for a `config.ini` file in their root dir. We might want to use a `.yml` file instead to match our config file format. Or we could use Python in the required `__init__.py` file. 

2. We could allow nested config settings in `mkdocs.yml`. Following existing conventions, perhaps something like:

    ```yaml
    theme:
        themename:
            - extra_templates: [sitemap.html]
            - include_sidebar: false
     ````

3. A combination of the two. The theme author could define some defaults and the user could override them. This would also make a great way for themes to define optional behavior (`include_sidebar` which defaults to `True` but the user can set to `False`).

Yes, we have the `extra` config setting. While that is already used by third party themes to enable/disable theme features, the theme has to provide for the setting not being defined at all. And `extra` does not provide a way for a theme to pass more templates through `build_template`.

If we just want to allow a theme to pass more pages through `build_template`, then option 1 makes the most sense to me. But personally, I would prefer option 3. Option 2 would work, but the user would need to explicitly set each setting before it would be defined at all.

waylan:
Themes now have `theme` objects, and theme specific configs.
Themes can inherit from other themes. Users (and theme authors)
can define custom static templates and variables.

Note that the `theme_dir` config option is being deprecated in favor of a nested `custom_dir` option.

This fixes #1164.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,5 @@
 include README.md
 include LICENSE
-recursive-include mkdocs *.ico *.js *.css *.png *.html *.eot *.svg *.ttf *.woff *.woff2 *.xml *.mustache
+recursive-include mkdocs *.ico *.js *.css *.png *.html *.eot *.svg *.ttf *.woff *.woff2 *.xml *.mustache *mkdocs_theme.yml
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -25,6 +25,40 @@ The current and past members of the MkDocs team.
 
 ### Major Additions to Version 1.0.0
 
+#### Theme Customization. (#1164)
+
+Support had been added to provide theme specific customizations. Theme authors
+can define default options as documented in [Theme Configuration]. A theme can
+now inherit from another theme, define various static templates to be rendered,
+and define arbitrary default variables to control behavior in the templates.
+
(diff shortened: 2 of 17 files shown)


waylan:
In #1164 I also proposed a way for templates to specify which version(s) of Mkdocs they work with. I have not yet implemented that. I think I'll leave this open to add that later. In the meantime, feedback from others (especially third party theme authors) would be welcome.

PR #1202, opened 2017-04-07
fix the link to Tom Christie in the copyright notice

NickVolynkin:
The current link is broken. I've reverted it to its previous state.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -26,5 +26,5 @@ markdown_extensions:
         permalink: 
     - admonition:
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Don't much mind either ways. I'm merging this, as it's at least better than the existing state, but perfectly happy to have them either both updated to http://www.tomchristie.com or left as-is.

NickVolynkin:
@waylan the `/about/release-notes/#maintenance-team` is the next link in the same string, and duplicating it makes little sense. That's why I chose the link to twitter.

@tomchristie  `http://www.tomchristie.com/` seems to be the best option, because that page has links both to github and twitter.

PR #1223, opened 2017-05-09
The Plugin API.

### LINKED ISSUE #206: Plugin API.

waylan:
See the included documentation for a full explaination of the API.

Search has been removed and wrapped in a plugin (named "legacy_search").
The search feature was ported as-is. None of the open issues with search
have been addressed. Presumably, a new search plugin will be built from
the ground up which addresses those issues.

Note that there is no 'default' plugins. That means that users get no
search unless they enable it in the config by adding the following to
their `mkdocs.yml` file:

    plugins:
        - legacy_search

Fixes #206


d0ugal:
I've only skimmed it quickly - overall it looks good, I really like the direction.
One question, why `legacy_search` rather than just `search`? Is there a plan for a new search plugin?

waylan:
> Is there a plan for a new search plugin?

Yes. Half of the remaining open tickets are regarding search. It would be easier to start over. If users really want the old plugin, then they can fall back to it. However, if it makes sense after we have the new plugin, we can always get ride of `legacy_search` before release.

waylan:
While working through some failing tests, The [linkchecker][1] tests made me realize that there is something missing from the API. In those tests we run through the integration tests and then run the linkchecker on the built sites. Because those sites are built without the Plugin API, the search assets are not included in the build, causing the build to fail. I could just include the plugin in each config file, but that doesn't address that issue that the template still includes script tags pointing to the search assets regardless of whether the plugin is activated.

In other words, we need a way for plugins to load scripts. We can't use `extra_scripts` because they assume a base URL of the `docs_dir` rather than a template dir. It's easy enough to add an attribute to the BasePlugin class for `extra_libs` and/or `extra_scripts`, but do we want to limit that to plugins only or open that up for themes to use as well?

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/230466826

waylan:
So `extra-javascript` works fine as a way for a plugin to add scripts. I forgot the files are copied independent of entries in the setting so it doesn't matter where the file originated. As long as the value is a valid path from the server root of the built site to an actual script, it works fine.

The real problem is that the legacy search script depends on require.js so rather than a straightforward script tag pointing to `search.js` we need a custom script tag which won't get generated from an entry in `extra_javascript`. I'll need to hack a workaround until we can refactor the plugin.

waylan:
Well the workaround was surprisingly easy. And it serves as a demonstration of how a theme could conditionally support a plugin (based on whether it is installed or not). I included the (commented out) code that uses `extra_javascript` as a reminder (and demonstration) of how to address the issue in most situations.

This passes all tests now and is ready to merge. However, I'll let it sit for a few days to give others a chance to provide feedback.

waylan:
Just pushed a better fix which ever so slightly refactors `search.js` so it works with the plugin API.

d0ugal:
Woop! Great job.

Issue #1220, opened 2017-05-09
Support query string in edit_uri


ovasquez:
The `edit_uri` setting in the `mkdocs.yml` file doesn't support the question mark, which is a valid and very common character in URLs. When a '?' is used the edit_uri contents are ignored and the edit URL is incorrectly generated.

This approach is used in the _Visual Studio Team Services_ git repository, so it would be really helpful to be able to link the files.
For example: 
```
edit_uri: '?path=/website/docs/'
```
This also happens if the '?' is added in the `repo_url` setting.

waylan:
This appears to be the behavior of `urljoin`:

>>> urljoin('http://example.com/foo/bar', 'baz/blah.txt')
'http://example.com/foo/baz/blah.txt'
>>> urljoin('http://example.com/?path=/foo/bar', 'baz/blah.txt')
'http://example.com/baz/blah.txt'

The first example is what we anticipated. The first string points to the location and the second string is the path from the document root to the file. But when the first string includes a query string (part that starts with `?`) then is strips that part out as we see in the second example. This actually makes sense. It would be wrong to have a query string in the middle of a URL and `joinurl` wouldn't necessarily consider that you might want to join a string to the query string. Consider this example:

urljoin('http://example.com/', '?path=/foo/bar', 'baz/blah.txt')
'http://example.com/?path=/foo/bar'

There it adds the query string, but then ignores everything after it. Looking at the [source], we don't use `joinurl` to join the repo_url and the edit_url, but we do use joinurl to join the file path to the combined repo_url/edit_url.

[source]: https://github.com/mkdocs/mkdocs/blob/b3ca24234b5cbd29d401dd8477fe0d9a7c8f48a4/mkdocs/nav.py#L214-L216

ovasquez:
@waylan I played around with the code and noticed that the *query string* is supported in the second parameter of *urljoin*.
Is there any case where changing this code in the source could create an issue? From this:
```
self.edit_url = utils.urljoin(
                repo_url + edit_uri,
                self.input_path)
```
to this:
```
self.edit_url = utils.urljoin(
                repo_url,
                edit_uri + self.input_path)
```
Based on the usage of the `edit_uri` it should be the root and if I understand correctly `self.input_path` is relative so there should be no overlap between them, or am I missing something?

waylan:
[...]

PR #1224, opened 2017-05-10
Adding support for query strings in the edit_uri

ovasquez:
Fixing #1220
Includes tests and documentation.

### DIFF
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -205,15 +205,19 @@ class Page(object):
 
     def set_edit_url(self, repo_url, edit_uri):
         if not repo_url.endswith('/'):
-            repo_url += '/'
+            # Skip when using query or fragment in edit_uri
+            if not edit_uri.startswith('?') and not edit_uri.startswith('#'):
+                repo_url += '/'
         if not edit_uri:
             self.edit_url = repo_url
         else:
+            # Normalize URL from Windows path '\\' -> '/'
+            input_path_url = self.input_path.replace('\\', '/')
             if not edit_uri.endswith('/'):
                 edit_uri += '/'
             self.edit_url = utils.urljoin(
-                repo_url + edit_uri,
-                self.input_path)
+                repo_url,
+                edit_uri + input_path_url)
 
 
 class Header(object):
--- a/mkdocs/tests/nav_tests.py
+++ b/mkdocs/tests/nav_tests.py
@@ -610,3 +610,149 @@ class TestLegacyPagesConfig(unittest.TestCase):
         self.assertEqual(str(site_navigation).strip(), expected)
         self.assertEqual(len(site_navigation.nav_items), 3)
         self.assertEqual(len(site_navigation.pages), 6)
+
+    def test_edit_uri(self):
+        """
+        Ensure that set_edit_url creates well formed URLs for edit_uri
+        """
+
+        pages = [
+            'index.md',
+            'internal.md',
+            'sub/internal.md',
+            'sub1/sub2/internal.md',
+        ]
+
+        # Basic test
+        repo_url = 'http://example.com/'
+        edit_uri = 'edit/master/docs/'
+
+        site_navigation = nav.SiteNavigation(pages)
+
+        expected_results = (
+            repo_url + edit_uri + pages[0],
+            repo_url + edit_uri + pages[1],
+            repo_url + edit_uri + pages[2],
+            repo_url + edit_uri + pages[3],
+        )
+
+        for idx, page in enumerate(site_navigation.walk_pages()):
+            page.set_edit_url(repo_url, edit_uri)
+            self.assertEqual(page.edit_url, expected_results[idx])
+
+        # Ensure the '/' is added to the repo_url and edit_uri
+        repo_url = 'http://example.com'
+        edit_uri = 'edit/master/docs'
+
+        site_navigation = nav.SiteNavigation(pages)
+
+        for idx, page in enumerate(site_navigation.walk_pages()):
+            page.set_edit_url(repo_url, edit_uri)
+            self.assertEqual(page.edit_url, expected_results[idx])
+
+        # Ensure query strings are supported
+        repo_url = 'http://example.com'
+        edit_uri = '?query=edit/master/docs/'
+
+        site_navigation = nav.SiteNavigation(pages)
+
+        expected_results = (
+            repo_url + edit_uri + pages[0],
+            repo_url + edit_uri + pages[1],
+            repo_url + edit_uri + pages[2],
+            repo_url + edit_uri + pages[3],
+        )
+
+        for idx, page in enumerate(site_navigation.walk_pages()):
+            page.set_edit_url(repo_url, edit_uri)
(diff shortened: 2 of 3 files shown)


mkdocs/tests/nav_tests.py
```diff
@@ -610,3 +610,84 @@ def test_indented_toc_missing_child_title(self):
...
+
+        # Test repo_url ending without '/'
+
+        repo_url = 'http://example.com/workspace'
+        edit_uri = '?path=edit/master/docs/'
+
+        site_navigation = nav.SiteNavigation(pages)
+
+        expected_results = (
+            repo_url + edit_uri + pages[0],
```
waylan: This would equal `http://example.com/workspace?path=edit/master/docs/index.md`. Is that correct or do we need to add an extra slash in there: `repo_url + '/' + edit_uri + pages[0]`. I'm not sure which way you intended. See my other comment below.

mkdocs/tests/nav_tests.py
```diff
@@ -610,3 +610,84 @@ def test_indented_toc_missing_child_title(self):
...
+
+        # Test repo_url ending with '/'
+
+        repo_url = 'http://example.com/workspace/'
+        edit_uri = '?path=edit/master/docs/'
+
+        site_navigation = nav.SiteNavigation(pages)
+
+        expected_results = (
+            repo_url + edit_uri + pages[0],
```
waylan: This would equal `http://example.com/workspace/?path=edit/master/docs/index.md`. Is that correct or do we need to remove a slash: `repo_url[:-1] + '/' + edit_uri + pages[0]`. I'm not sure which way you intended. See my other comment above.

Either way we need to normalize to one output format. Users could easily enter either in their config and we should just do the right thing.

mkdocs/utils/__init__.py
```diff
@@ -479,3 +479,47 @@ def nest_paths(paths):
...
+                path_string += '/'
+            path_parsed = path_parsed._replace(
+                query=url_query + path_string + path_parsed.query,
+                path='')
+
+    # new_path contains the query string that will be ignored from new_url
+    new_url = urlunparse(url_parsed)
+    new_path = urlunparse(path_parsed)
+
+    return urljoin(new_url, new_path)
```
waylan: This seems like an awful lot of code just to support query string URLs. Is this all really necessary? I'm not even convinced we need to use `urljoin`. If we do, fine, but if not then removing that use seems like a better option.

Also, if we explicitly support query strings, then we should explicitly support hash urls as well (`http://example.com/path/to/root/#/foo/bar/page.md`).

ovasquez:
@waylan Thanks for all the comments. Regarding the use of `urljoin` and the complex code, it might end up being easier to restrict the use of query strings and hash urls only in the `edit_uri`, that way the first and straightforward solution of concatenating the `edit_uri` and the `self.input_path` would work.

Do you consider this restriction viable for the end user? If yes, then it will be simple to support and to document, as there will not be any complex behavior involved.

waylan:
`self.input_path` is always going to be the local path from `docs_dir` to the local file. Therefore, it would never contain a query string or hash. In fact, it is not even overridable by the user. The two unknowns are the `repo_url` and `edit_uri`. The user could put anything in either config setting. We [validate][1] `repo_url` to check that `urlparse` can parse it and that it contains a `scheme`, but that's it. `edit_uri` could be any string. Although, if the `repo_url` points to a known host (only GitHub and Bitbucket at this point), then we do [auto-populate][2] `edit_uri` (if it is unset). I suspect the current implementation only expects URLs of the style used by those two services.

So to answer your question, I would expect the only location of a query string or hash to be in the `edit_uri`, but I suppose one could be found in the `repo_url` (although it seems highly unlikely to me).

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/config_options.py#L169-L182
[2]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/config_options.py#L193-L209

ovasquez:
Based on the discussion, I will add support for query and hash URLs in the `edit_uri`, which is a lot better than no support at all. 

I believe it will cover almost all cases and can implemented without any complex logic. I'll update the documentation to clarify this. Thanks for the help!

ovasquez:
@waylan I updated the code with a much simpler version, tests covering more cases and improved documentation. The *windows to posix* path change in the code is the simplest version I could find that worked in python 2.6-2.7 without going completely overkill. I didn't use `os.path.sep` (like in [`get_url_path`](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/utils/__init__.py#L204)) because I needed the Windows tests to pass in Linux.

Regarding of whether or not to include the `'/'` before the query or fragment character, after reading and searching for a while I understood that both (with or without) options are allowed by modern browsers so restricting to one style might prevent some users from using this functionality. Here's an interesting read about the topic http://stackoverflow.com/a/42193734

mkdocs/tests/nav_tests.py
```diff
@@ -610,3 +610,125 @@ def test_indented_toc_missing_child_title(self):
...
+            repo_url + edit_uri + pages[3],
+        )
+
+        for idx, page in enumerate(site_navigation.walk_pages()):
+            page.set_edit_url(repo_url, edit_uri)
+            self.assertEqual(page.edit_url, expected_results[idx])
+
+        # Ensure the '/' is added to the repo_url for simple edit_uri
+        repo_url = 'http://example.com'
+        edit_uri = 'edit/master/docs/'
```
waylan: Perhaps we should also include a test where the `edit_uri` is missing the slash at the end.

mkdocs/nav.py
```diff
@@ -205,15 +205,19 @@ def set_canonical_url(self, base):
...
+            # Skip when using query or fragment in edit_uri
+            if not edit_uri.startswith('?') and not edit_uri.startswith('#'):
+                repo_url += '/'
         if not edit_uri:
             self.edit_url = repo_url
         else:
             if not edit_uri.endswith('/'):
                 edit_uri += '/'
+            # Normalize URL from Windows path '\\' -> '/'
+            input_path_url = self.input_path.replace('\\', '/')
```
waylan: You might need to do this before added a missing slash (switch order of this and previous statement). As is, it could result in two slashes in some circumstances. Probably should add a Windows specific test for that as well.

Also, any reason why you didn't use `mkdocs.utils.path_to_url()` here?

ovasquez: Because it uses `pathname2url` which will turn the '?' and '#' to their encoded versions and generate wrong URLs.

waylan: Ahh right. Just wanted to make sure there was a good reason.

mkdocs/tests/nav_tests.py
ovasquez: Sure, I will include those tests later today. I think I'll just create 2 test functions, one containing all tests for posix-style paths and the other one for windows-style paths, to reduce the length of the code.

ovasquez:
@waylan I updated the code and tests as requested. Let me know what you think.

PR #1233, opened 2017-06-04
Removed hack that broke images on Windows

sesispla:
We are working in a team with both MacOS and Windows computers. We where facing and estrange issue where media routes where not properly generated in Windows.

If you have the following folder structure:
- mkdocs.yml
- docs
    - level1
        -  level2
            - img
                - sample.png
            - index.md

Where the content of index.md is:

```markdown
This is a route test:
![alt text](img/sample.png)
``` 
The output in MacOS is referencing the current path (right):
```html
<p> 
    This is a test
    <img src="./img/sample.png" alt="alt text" />
</p>
```

However, in Windows we where getting a reference to the upper folder (wrong):
```html
<p> 
    This is a test
    <img src="../img/sample.png" alt="alt text" />
</p>
```

After some debugging in both Windows and MacOS platforms, I found that the following hack where causing this behaviour:

```python
def create_relative_media_url(nav, url):
...

    # TODO: Fix this, this is a hack. Relative urls are not being calculated
    # correctly for images in the same directory as the markdown. I think this
    # is due to us moving it into a directory with index.html, but I'm not sure
    if (nav.file_context.current_file.endswith("/index.md") is False and
            nav.url_context.base_path != '/' and
            relative_url.startswith("./")):
        relative_url = ".%s" % relative_url

    return relative_url
```

As I mentioned before, in the Windows case, the ".%s" is adding an extra "." to the media URL, resulting in a broken image due to incorrect path:

```python
        relative_url = ".%s" % relative_url
```

I have removed the hack and checked that the behaviour of media url is still correct in both Windows and MacOS, for the following scenarios:

1. Building the documentation
2. Executing the following test scenario, with both image at same level as MD (as stated in the hack comment) and in a folder a index.md relative level:
- mkdocs.yml
- docs
    - level1
        -  level2
            - img
                - sample.png
            - index.md
            - sample2.png

This PR removes the hack from create_relative_media_url and

### DIFF
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -124,7 +124,7 @@ class UtilsTests(unittest.TestCase):
 
         assertPathGenerated("img.png", "./img.png")
         assertPathGenerated("./img.png", "./img.png")
-        assertPathGenerated("/img.png", "../img.png")        
+        assertPathGenerated("/img.png", "../img.png")
 
     def test_reduce_list(self):
         self.assertEqual(
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -7,7 +7,6 @@ import mock
 import os
 import unittest
 
-from sys import platform
 from mkdocs import nav, utils, exceptions
 from mkdocs.tests.base import dedent
 
@@ -93,10 +92,7 @@ class UtilsTests(unittest.TestCase):
             ]}
         ])
         site_navigation.url_context.set_current_url('/subpage/')
-        if platform == "win32":
-            site_navigation.file_context.current_file = "subpage\\index.md"
-        else: 
-            site_navigation.file_context.current_file = "subpage/index.md"
+        site_navigation.file_context.current_file = "subpage/index.md"
 
         def assertPathGenerated(declared, expected):
             url = utils.create_relative_media_url(site_navigation, declared)
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -102,6 +102,30 @@ class UtilsTests(unittest.TestCase):
         assertPathGenerated("./img.png", "./img.png")
         assertPathGenerated("/img.png", "../img.png")
 
+    def test_create_relative_media_url_sub_index_windows(self):
+        '''
+        test special case where there's a sub/index.md page and we are on Windows.
+        current_file paths uses backslash in Windows
+        '''
+
+        site_navigation = nav.SiteNavigation([
+            {'Home': 'index.md'},
+            {'Sub': [
+                {'Sub Home': '/level1/level2/index.md'},
+
+            ]}
+        ])
+        site_navigation.url_context.set_current_url('/level1/level2')
+        site_navigation.file_context.current_file = "level1\\level2\\index.md"
+
+        def assertPathGenerated(declared, expected):
+            url = utils.create_relative_media_url(site_navigation, declared)
+            self.assertEqual(url, expected)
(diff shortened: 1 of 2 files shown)


waylan:
Thanks for the contribution. However, with this change we do have at least one failing test in both [linux][1] and [windows][2]. I'm not sure if the test is wrong or not. I suspect not and this hack was ensuring that that test passed. In other words, we get either one broken edge case or the other. I'll need to take a closer look when I have some time and am more rested.

Even more interesting, is that the [linkchecker test][3] is also failing with many links that point to nonexistent resources in our documentation. In other words, this change breaks our own documentation. I'd say this is a pretty good indication that my hunch is correct. 

As an aside, as part of version 1.0 I am currently working on a refactor which completely rewrites the way relative URLs are calculated. In theory is should eliminate these sorts of problems. That being the case, I'm inclined to leave this as-is until then -- unless someone can provide a simple fix that can pass all our tests and resolve the reported problem.

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/239209718#L246
[2]: https://ci.appveyor.com/project/d0ugal/mkdocs/build/1314.master/job/51p0v7e06c3fo59r#L337
[3]: https://travis-ci.org/mkdocs/mkdocs/jobs/239209710#L507

sesispla:
I see @waylan :(

I'll try to develop a simple patch that solves the issue, while you work with the new version.

I have an important part of the team (how uses Windows) limited by this issue, so is important for me to give a solution.

I'll run tox in local and then push  to my master to check the progress with you CI pipeline.

sesispla:
Ok, I think I found the root issue:

As you can see in the following screens, the "endswith is false" condition is not evaluated equally in Windows and Non-Windows systems. Having all 3 conditions to "true" has causing an incorrect execution of the hack, adding an additional "." to the images in the concrete scenario I described before.

I have added an additional check, comparing platform and path specific forms for Windows and non-Windows platforms. 

Local tox tests looks good. My concrete scenario works properly. So I think this is the easy fix we are looking for. Let's see how CI tests ends. 

screen shot 2017-06-04 at 05 15 08

screen shot 2017-06-04 at 05 15 01

sesispla:
IMO the hack now looks great, but since it now should be able to work with two platforms and differentiate between slash and backslash in current_file, we have some tests to fix.

The other solution I thought, that I don't like at all, is remove the slash from the hack (nav.file_context.current_file.endswith("index.md")), which may lead to other issues if someone uses something like "my-index.md".

What do you think @waylan ?

facelessuser:
Instead of all the platform stuff, couldn't you just check it with regex?

```
re.match(r'.*(?:\\|/)index.md$', nav.file_context.current_file) is None
```

facelessuser:
I guess you could do it without regex as well:

```
nav.file_content.current_file.endswith(('/index.md', '\\index.md')) is False
```

sesispla:
Good idea @facelessuser. Now only flake8 test is broken, but since your proposal looks simpler, it would be definitely better for future maintainability.

Let me have a try.

sesispla:
@waylan @facelessuser here we are :) Sorry for the commit mess.

Finally, I used the regex option and all test passes. I have changed the utils/__init__.py in one of my Windows computers and our docs site is now generating images properly.

Hope this helps.

waylan:
Looks good. Could we add a test which failed (at least on Windows) before the change and passes now?

sesispla:
Sure. Working on it

sesispla:
Ready to go :) 

I have added a special unit test (test_create_relative_media_url_sub_index_windows) where the SiteNavigation and current_file meets the original scenario we identified.

waylan:
Awesome! @sesispla thanks for your work on this.

PR #1236, opened 2017-06-05
Open files with `utf-8-sig` to account for BOM.

### LINKED ISSUE #1186: First heading not rendered correctly
Hello,

I have a simple index markdown file:

```
# Project
## Overview
```

And it fails to render correctly in any theme.  

**Cinder**
keqxxqn - imgur

**ReadTheDocs**
amsuaqe - imgur

Am I doing something wrong?  If I change it to a level 2 heading, the same thing happens. 

I can workaround it by placing a blank line at the start of the file, but then the heading appears too low:

6xtusui - imgur

Thanks.

waylan:
Python simply discards the BOM with `utf-8-sig`. This way, users of
Microsoft text editors can have their files properly parsered. In all other
ways behaves as reading files using `utf-8` encoding. For more info see:
https://docs.python.org/2/library/codecs.html#encodings-and-unicode

Fixes #1186.

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -177,7 +177,7 @@ def _build_page(page, config, site_navigation, env, dump_json, dirty=False):
 
     # Read the input file
     try:
-        input_content = io.open(input_path, 'r', encoding='utf-8').read()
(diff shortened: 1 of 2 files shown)


waylan:
Closed in favor of #1309.

PR #1252, opened 2017-07-10
Added command 'add'

Daethe:
This command allow user to create new file for documentatin following created template.

PR #1253, opened 2017-07-13
New command 'add' to create new page from a page template.

MarcProux:
I create this PR to push the right branch. Old PR is https://github.com/mkdocs/mkdocs/pull/1252

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -10,7 +10,7 @@ from mkdocs import __version__
 from mkdocs import utils
 from mkdocs import exceptions
 from mkdocs import config
-from mkdocs.commands import build, gh_deploy, new, serve
+from mkdocs.commands import build, gh_deploy, new, serve, add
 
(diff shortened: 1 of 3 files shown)


waylan:
Thanks @Daethe. I'll review this when I have some time (it may be a while).

Daethe:
I will continue to improve the command to be able to use variables within the templates that will have to be configured either in a file '_config.yml' or in a header.

So I will create a new branch to handle these improvements and make a new pull request if it is closed.

Daethe:
Issue when running check in 'mkdocs/tests/gh_deploy_tests.py' function 'test_deploy_error'

oprypin:
So this PR is definitely stale (sorry about that). I am also inclined to reject this as a feature request altogether, because the functionality doesn't benefit from being tied to MkDocs: people can easily invent their own external approach that suits them best.
Further discussions on the merits of such a feature can still be brought up if there's more to add.

PR #1262, opened 2017-07-28
It should be deployable as a python web app

arycloud:
I have seen a lot of developers trying to deploy that mkdocs whole project so they can easily get access mkdocs functionality online across the whole team.But mkdocs comes as a python package, I think it should comes like a framework app OR some deployment configuration for example we can add remote host & port.


waylan:
> I have seen a lot of developers trying to deploy that mkdocs whole project so they can easily get access mkdocs functionality online across the whole team.

This is **not** the intended user case for the server. It is **only** intended to be a local dev server and nothing more. There is absolutely no testing or consideration given to thread safety or other such complications when a server has to interact with multiple connections. Adding this functionality would commit us to the added burden of maintaining a more complicated server environment. My recommendation is to not accept this proposed change. There is no reason why teams can't create automation scripts which (re)build the site behind a proper flat file server.

PR #1256, opened 2017-07-15
#1134 but against 1.0.dev - Follow_symlinks if they exist

warmfusion:
This is an updated PR pointing at 1.0-dev as suggested by @waylan in https://github.com/mkdocs/mkdocs/pull/1134#issuecomment-315421392

Allows symlinks within docs folder to be followed by python rather than ignored.

This corrects an inconsistency with [existing behaviour](https://github.com/mkdocs/mkdocs/issues/869) where assets in symlinked paths are followed resulting in a sub-dir holding assets but no content.

PR #1269, opened 2017-08-30
Follow symlinks in subfolders of docs_dir

ruudud:
My usecase is that I have a `docs/` dir with general files, and subdirs with project-specific documentation. These subdirs should ideally be symlinks, eg. to easen local development.

This patch makes the automatic page discovery follow symlinks, which currently is not the case.


waylan:
This needs a test. I'm not sure exactly what the test should look like (I haven;t given it much thought), but it should fail without this change. Add a test, and I'll merge this.

waylan:
Oh, this is a duplicate of #1256.

PR #1272, opened 2017-09-06
Adds a AnchorLink.level property

funkyfuture:
this allows themes to restrict the toc to a certain depth.

here's a proof-of-concept: https://github.com/squidfunk/mkdocs-material/pull/450

### DIFF
--- a/mkdocs/tests/toc_tests.py
+++ b/mkdocs/tests/toc_tests.py
@@ -132,3 +132,25 @@ class TableOfContentsTests(unittest.TestCase):
         """)
         toc = markdown_to_toc(md)
         self.assertEqual(str(toc).strip(), expected)
+
+    def test_level(self):
+        md = dedent("""
+        # Heading 1
+        ## Heading 1.1
+        ### Heading 1.1.1
+        ### Heading 1.1.2
+        ## Heading 1.2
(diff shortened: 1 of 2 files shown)


waylan:
Hmm, so this adds a `level` attribute to the `AnchorLink` class to track the "level" of each item in a template. Is that really necessary? Why can't you track that with a template variable? In other words, convince me this is the right approach to the problem.

Regardless, this would need to have some tests. Also, it should be against the 1.0.dev branch. Finally, as this would be a 1.0 feature, the template specific setting (in the associated Material PR) may want to take advantage of the new template settings implemented in #1201,

funkyfuture:
> Is that really necessary? Why can't you track that with a template variable? In other words, convince me this is the right approach to the problem.

in general i tend to keep the logic in templates at a minimum and rather pass rich context objects. imo this is easier to read in the templates. also, all theme developers and users can easily access this information w/o implementing a logic for it. it's obviously simply implemented and maintainable on the Python side.

thanks for the other hints.

waylan:
Add some tests and I'll give this another look.

funkyfuture:
a test is added. the fails are caused by missing interpreters.

PR #1273, opened 2017-09-07
Fixes a bug when 'edit_url' is not defined in the config

funkyfuture:
i encountered this tailed traceback:

```
  File "/home/user/mkdocs/mkdocs/nav.py", line 180, in __init__
    self._set_edit_url(config['repo_url'], config['edit_uri'])
  File "/home/user/mkdocs/mkdocs/nav.py", line 263, in _set_edit_url
    if not edit_uri.startswith('?') and not edit_uri.startswith('#'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

### DIFF
--- a/mkdocs/config/__init__.py
+++ b/mkdocs/config/__init__.py
@@ -1,5 +1,6 @@
-from mkdocs.config.base import load_config
-from mkdocs.config.config_options import Config
+from mkdocs.config.base import load_config, Config
 from mkdocs.config.defaults import DEFAULT_SCHEMA
 
-__all__ = ["load_config", 'Config', 'DEFAULT_SCHEMA']
+__all__ = [load_config.__name__,
+           Config.__name__,
+           'DEFAULT_SCHEMA']
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -212,15 +212,16 @@ class RepoURL(URL):
     """
     Repo URL Config Option
 
-    A small extension to the URL config that sets the repo_name, based on the
-    url if it hasn't already been provided.
+    A small extension to the URL config that sets the repo_name and edit_uri,
+    based on the url if they haven't already been provided.
     """
 
     def post_validation(self, config, key_name):
+        repo_host = utils.urlparse(config['repo_url']).netloc.lower()
+        edit_uri = config.get('edit_uri')
 
+        # derive repo_name from repo_url if unset
         if config['repo_url'] is not None and config.get('repo_name') is None:
-            repo_host = utils.urlparse(
-                config['repo_url']).netloc.lower()
             if repo_host == 'github.com':
                 config['repo_name'] = 'GitHub'
             elif repo_host == 'bitbucket.org':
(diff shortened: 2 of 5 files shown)


waylan:
Thanks for this. However, this needs a test that fails before the change and passes after before it will be accepted.

funkyfuture:
done. i realised the bug is only evoked when `repo_name` is not a default value.

funkyfuture:
locally i rebased, so the fix is HEAD, but it appears different here.

waylan:
The tests for this are [here][1]. I was expecting something similar, not an integration test.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/nav_tests.py#L614-L758

funkyfuture:
well, i looked into this. but i couldn't evoke the bug with this test:

```python
        repo_name = 'example'
        repo_url = 'http://example.com/'
        edit_uri = None

        site_navigation = nav.SiteNavigation({
            'pages': pages,
            'repo_name': repo_name,
            'repo_url': repo_url,
            'edit_uri': edit_uri,
            'docs_dir': 'docs',
            'site_dir': 'site',
            'site_url': '',
            'use_directory_urls': True
        })
```

am i missing something?

waylan:
Don't you want to test the case where `repo_url = None` (rather than `edit_uri`)?

waylan:
FYI, I just pushed a fix for the missing interpreters on Travis (see #1274 & #1275). If you rebase this, the tests should all run properly now.

funkyfuture:
> Don't you want to test the case where repo_url = None (rather than edit_uri)?

no, i'd rather test the scenario that evoked the bug, as demonstrated in the integration test. can you give me a hint how to write a test to hit that code path with a simple unit test? i'm not very deep into the code.

waylan:
Oh right. Sorry. Are you sure you rolled back your changes before trying your test? By my reading of the code, that should be correct.

Actually, I wonder if the appropriate fix would be to address this in config validation. Validation is already ensuring the `edit_uri` setting is a string (see [mkdocs/config/defaults.py#L82][1]) However, there is no default defined. Perhaps we should add `default=''` to avoid `None` as the default?

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config/defaults.py#L78-L82

funkyfuture:
> Are you sure you rolled back your changes before trying your test?

yep.

> I wonder if the appropriate fix would be to address this in config validation.

I had that thought while looking for a simple test. imo, the `edit_url` could be expanded in the config as well.

waylan:
> the `edit_url` could be expanded in the config as well.

You make a good point. While the `edit_url` is intentionally page specific (a separate URL for each page), combining the `repo_url` and `edit_uri` from the settings only needs to happen once. Perhaps that should happen in post-validation.

Truth be told, I haven't given much attention to this feature as I expect it will get ripped out and replaced with a (more flexible) plugin anyway (the plugin API is currently available in 1.0.dev). The recent changes in this space have all been from random one-time contributors. That said, we do have a bug, which should be fixed (by the way, this PR should probably be against the master branch).

funkyfuture:
okay, i moved the logic to the config. i see that the feature has its flaws and really should be moved to a plugin. well, it works for now.

waylan:
@funkyfuture Thanks for the work on this. Greatly appreciated.

funkyfuture:
@waylan shall i prepare a pr that merges this into the `1.0.dev` branch? i'd like to use this one with my use-case.

waylan:
I expect that part of the code will go through a major rewrite for 1.0, so I wasn't all that concerned about it.

funkyfuture:
i could throw out that stuff, but w/o a substituting plugin. would you accept that?

PR #1285, opened 2017-09-18
Running tests with Docker

funkyfuture:
this makes it possible to run all tests locally as on travis - provided one has a Docker client available.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+*.egg-info
+*.pyc
+__pycache__/
+.git/
+.idea/
+.tox/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,65 +1,18 @@
 sudo: false
 language: python
-dist: trusty
+cache: pip
 
-matrix:
-  include:
-    # Default Python
-    - env: TOXENV=flake8
-    - env: TOXENV=markdown-lint
-      before_install: gem install mdl
-    - env: TOXENV=linkchecker
-    - env: TOXENV=jshint
-      before_install: npm install -g jshint
-    - env: TOXENV=csslint
-      before_install: npm install -g csslint
-    # Python version specific
-    - python: '2.7'
-      env: TOXENV=py27-integration
-    - python: '2.7'
-      env: TOXENV=py27-min-req
-    - python: '2.7'
-      env: TOXENV=py27-unittests
-    - python: '3.3'
-      env: TOXENV=py33-integration
-    - python: '3.3'
-      env: TOXENV=py33-min-req
-    - python: '3.3'
-      env: TOXENV=py33-unittests
-    - python: '3.4'
-      env: TOXENV=py34-integration
-    - python: '3.4'
-      env: TOXENV=py34-min-req
-    - python: '3.4'
-      env: TOXENV=py34-unittests
-    - python: '3.5'
-      env: TOXENV=py35-integration
-    - python: '3.5'
-      env: TOXENV=py35-min-req
-    - python: '3.5'
-      env: TOXENV=py35-unittests
-    - python: '3.6'
-      env: TOXENV=py36-integration
-    - python: '3.6'
-      env: TOXENV=py36-min-req
-    - python: '3.6'
-      env: TOXENV=py36-unittests
-    - python: 'pypy'
-      env: TOXENV=pypy-integration
-    - python: 'pypy'
-      env: TOXENV=pypy-min-req
-    - python: 'pypy'
-      env: TOXENV=pypy-unittests
-    - python: 'pypy3'
-      env: TOXENV=pypy3-integration
-    - python: 'pypy3'
-      env: TOXENV=pypy3-min-req
-    - python: 'pypy3'
-      env: TOXENV=pypy3-unittests
+python:
+  - 2.7
+  - 3.3
+  - 3.4
+  - 3.5
+  - 3.6
 
 install:
-  - pip install codecov
-  - pip install tox
+  - travis_retry pip install codecov tox-travis
+  - travis_retry gem install mdl
+  - travis_retry npm install -g csslint jshint
 
 script:
   - git clean -f -d -x
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+# in case something's wrong with this script, mention @funkyfuture
+# in issues and poll requests
+
+FROM funkyfuture/nest-of-serpents
+
+ENTRYPOINT tox
+WORKDIR /src
+
+RUN apt-get -q update \
+ && apt-get install -qy --no-install-recommends nodejs-legacy npm ruby \
+ && pip2.7 install LinkChecker \
+ && pip3.6 install coverage flake8 nose mock tox \
+ && pip3.6 install click Jinja2 livereload Markdown PyYAML tornado \
+ && mkdir /home/tox \
+ && mv /root/.cache /home/tox/ \
+ && gem install -N mdl \
+ && npm install -g csslint jshint \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+
+ARG uid
+RUN useradd --uid=$uid -m tox \
+ && chown -R tox.tox /home/tox/.cache
+
+ADD . .
+RUN chown -R tox.tox .
+
+USER tox
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -54,8 +54,23 @@ to be pip installed and then the test suite can be ran for MkDocs but running
 the command `tox` in the root of your MkDocs repository.
 
 It will attempt to run the tests against all of the Python versions we
-support. So don't be concerned if you are missing some and they fail. The rest
-will be verified by [Travis] when you submit a pull request.
+support. So don't be concerned if you are missing some and they fail.
+The rest will be verified by [Travis] when you submit a pull request.
+
+### Run tests with Docker
+
+If you have a running [Docker]-daemon running you can run tests from a container
+that has the necessary interpreters and packages installed and pass arguments
+to `tox`:
+
+    # in the project's root directory
+    ./run-docker-tests -e pypy3 -e csslint
+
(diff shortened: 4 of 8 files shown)


waylan:
What does this give us (besides more of a maintenance burden) that we don't already get from using tox directly?

waylan:
Looking more closely, I can see that this installs non-Python dependencies. But it does so with apt-get. What about non-Debian based systems?

funkyfuture:
> What does this give us (besides more of a maintenance burden) that we don't already get from using tox directly?

i'm using this technique in another project for a while now and there were hardly any changes necessary over the time (see [here](https://github.com/pyeve/cerberus/commits/master/.dockerignore), [here](https://github.com/pyeve/cerberus/commits/master/run-docker-tests) and [here](https://github.com/pyeve/cerberus/commits/master/Dockerfile)). the advantage is that one can run all tests even if one's system doesn't provide all needed interpreters (that are included in the base image) and tools (do i want to bloat my system with node packages? rather not.). it is clearly a faster development process to run tests locally than to wait for results from the CI (changes need neither to be commited nor pushed, tox' test environments can be selected). though this should hardly happen, a filesystem is safe from possible left-over artifacts that may be created by the code.

> What about non-Debian based systems?

i don't get what this question is aiming at.

waylan:
> > What about non-Debian based systems?
>
> i don't get what this question is aiming at.

There are a couple calls to `apt-get` in the docker file. Of course, `apt-get` is the package management system for Debian and its variants (including Ubuntu and friends). However, there are other non-Debian Linux distributions that use different package management systems, not to mention MacOS and Windows. So how does this work on those systems? Sorry, I have never used Docker before. To me, it's always seems like overkill.

facelessuser:
I'm no docker expert, but basically it's like running things in a VM instead of on your own file system.  A matter of fact, it's essentially a fancy wrapper around VMs.  I won't go into advantages of this vs VMs in a traditional way (as I honestly don't use docker enough to understand all that, but it seems you can have a server that does all this, so it's like remote VMs?).  My only use of docker is to pre-build Python binaries on really old Linux systems so I can distribute them and have them work on most linux systems (I have a very specific reason to do this, and it is not something people would normally need to do). I'm sure others use docker for various benefits far beyond what I use them for.

From what I understand, this is just to allow a user to run tests without having to run them on their local computer infrastructure (installing Python, installing dependencies, etc.).  I don't think it is meant to cover all cases, just provide a way to run the tests without needing the system to have all that installed on it.  I mean if you are running the docker image locally, then you have to have the docker image installed locally, but if you are running it remotely, then it is on your local system at all.

Now whether we need it, is an entirely different discussion.

waylan:
As I understand it, Docker does software virtualization while a VM does hardware virtualization. In fact, as I understand it, a docker image shares the host OS's kernel (while a VM is completely separate softwarewise). Therefore, my assumption has been that the docker image needs to be of the same system type as the host. Perhaps my assumption is wrong? I bring this up because its not clear to me if it can be used across system types. If not, then I'm not interested. If so, then we can have that discussion about whether we need it.

So, yes or no? Can the docker image be used across different OSes (including Linux, MacOS and Windows)?

facelessuser:
I'm running Linux docker containers on Mac. When installing docker on Mac, it literally installs virtualbox. I imagine there is more to it than that, but that is my novice experience.

funkyfuture:
> There are a couple calls to `apt-get` in the docker file.

yes, because the root image is `ubuntu:xenial`. the host system doesn't matter.

> Sorry, I have never used Docker before. To me, it's always seems like overkill.

at least on Ubuntu and Arch it's fairly easy to install. once you're there, you can run the script.

> I'm no docker expert, but basically it's like running things in a VM instead of on your own file system.

no, Docker (or other Linux container solutions) is not about virtualization. processes run in isolating control groups and in isolated portions of the fs (somewhat like `chroot`).

> but it seems you can have a server that does all this, so it's like remote VMs?

no, you can work with a local or remote daemon.

> just provide a way to run the tests without needing the system to have all that installed on it.

it provides an environment with all necessary software installed. that software is defined in the build instructions.

> Now whether we need it, is an entirely different discussion.

have you tried it?

>  a docker image shares the host OS's kernel

correct, a container (which is based on an image) runs in a designated control group of a Linux kernel that implements cgroups.

> Therefore, my assumption has been that the docker image needs to be of the same system type as the host. Perhaps my assumption is wrong?

that's not wrong. but you can talk to remote Docker daemons from a local client.

> So, yes or no? Can the docker image be used across different OSes (including Linux, MacOS and Windows)?

yes. [here](https://docs.docker.com/engine/installation/#supported-platforms)'s a list of supported platforms.

rickpeters:
You are mostly right @waylan, it’s about software packaging. Without making it to difficult, it depends on the OS kernel. So in general you have a Linux based container that needs a Linux kernel but the OS flavor of Linux is not important. Docker delivers solutions for Linux, macOS and Windows, your container works on all of them (although technically on macOS and windows the Docker software uses a small vm for that).
On newer windows versions it will be possible to run Linux containers natively due to WSL.
I use Docker to keep all mkdocs dependency’s in the container and this makes it unnecessary to install dependencies for mkdocs in the os itself. I think it’s a great solution.
You can also find my simple Docker image for mkdocs at rickpeters/mkdocs on GitHub.

facelessuser:
Again, I have tried it for the use cases listed above. But as stated in my disclaimers, my understanding of how it works may clearly be wrong. I was mainly starting my observations.

facelessuser:
After hearing more from other people, it seems Mac uses VirtualBox to create a base environment that Docker then uses.  Much like running Docker on Windows' WSL.  Thanks @rickpeters  for explaining this.  I really didn't understand that and was confused when I'd see VirtualBox running.  It kind of distorted my view of what was actually happening.

waylan:
> > So, yes or no? Can the docker image be used across different OSes (including Linux, MacOS and Windows)?
> 
> yes. here's a list of supported platforms.

Okay, now that we got that straight... Why do we need it? Seriously, other than a few dependencies, how does this help? Generally, I run `tox -e py27-unittests` and the same for whatever the default Python3 is on my system and then let the CI Server handle the rest (on larger changes I might run `flake8` locally as well). After all, every commit goes through a PR and must run through the CI Server anyway. Now that GitHub offers "Squash and Merge" the individual commits in the PR don't even need to be pretty. Extra commits to make minor adjustments never show up in master.

As it is today, I have to wait for the CI server. If I use Docker, then I have to wait for Docker and the CI Server. My point is that I'm not likely to ever use it. And when things break and/or we get requests to modify it, that will just be extra maintenance... Extra work for no gain.

funkyfuture:
i'm not proposing to make this method to run the tests mandatory.

> Why do we need it? 

> My point is that I'm not likely to ever use it.

for me it made testing way more accessible. i mentioned some arguments in my second comment.

> Now that GitHub offers "Squash and Merge" the individual commits in the PR don't even need to be pretty. Extra commits to make minor adjustments never show up in master.

 i certainly would not have put debug statements into commits to wait until the CI finished, analyze the output and so forth. 

there's nothing more frustrating to prepare and open a pr, open a beer, watch a tv show, go to bed and get informed in the morning that an environment that one doesn't have available failed in the CI.

> If I use Docker, then I have to wait for Docker and the CI Server.

well, after *all* tests passed within the container, i was very certain the CI would pass them too. how much faster the tests run locally depends on the system. in my case it's about factor 10.

waylan:
> i'm not proposing to make this method to run the tests mandatory.

Right and I'm saying I'm not going to use it. So, as I said ...

> And when things break and/or we get requests to modify it, that will just be extra maintenance... Extra work for no gain.

And that's my whole point. I'm meh on whether it's there or not. That is, until it creates more work for me. Then its just a problem in my way. So why should I create a future problem for myself with no gain?

funkyfuture:
> So why should I create a future problem for myself with no gain?

i described the gains for a subset of the community above. you could gain more contributions to the project.

what about an explicit mentioning of me for the case that there's a problem that no one gets fixed?

funkyfuture:
> what about an explicit mentioning of me in the files for the case that there's a problem that no one gets fixed?

i added explicit mentions of me for problem cases in the relevant files as inline comments.

waylan:
So the docs in this patch state:

> You can run the script without any arguments to test the project exactly as [travis] does without having to setup anything.

If that would continue to be the case after an update to the Travis config then I'd consider this. But that's not the case. Any updates to Travis require an update also to Docker, which is maintenance I don't want. If Docker worked off of the Travis config, that might (and I stress *might*) be something I'd consider.

funkyfuture:
> If that would continue to be the case after an update to the Travis config then I'd consider this. But that's not the case.

i'm assuming that the Travis config is always kept in sync with the tox config. the latter is what defines the behaviour of the container.

*nonetheless* and following that assumption, i've changed the Travis config to a more maintable state. this addresses how the tests are invoked on travis, not in the container.

waylan:
I realize that all the tests are being run, but I liked the matrix were each test was run separately--and more importantly reported separately--by Travis. It should be clear at a glance that it is the linter that failed, not just some test run in Python 2.7. Besides, this doesn't actually address my concern. Docker is still not using the Travis config. We **must** be able to modify Travis and or Tox without concern for Docker and yet Docker still properly runs all the Travis tests. I assume that that is not possible, therefore Docker is a no-go.

funkyfuture:
what i formulated in my previous post as assumption is imo a fact. the Travis config reflects the tox config, whenever that latter is changed the former might be adjusted.

this means that the container runs the same tests. one might emphasize the '*like*' in the description to make clear that this is not a Travis clone of some kind. i wouldn't consider the axiom that such must be achieved valid for the proposed purpose.

the use of `tox-travis` was a proposal to meet your desire for few maintenance.

btw, [there are](https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=mkdocs&starCount=0) quiet a lot of container images providing `mkdocs`. the few i reviewed are unmaintained and of poor quality. @squidfunk's okay though.

Issue #474, opened 2015-04-24
Improve Configuration Validation


d0ugal:
The config validation code is getting messier with each addition and change.

https://github.com/mkdocs/mkdocs/blob/master/mkdocs/config.py#L95-L204

Trying to think how to improve this:

1. Break up into smaller functions within `config.py`.
2. Use something like [voluptious](https://pypi.python.org/pypi/voluptuous).
3. Deligate validation to the respective areas - so a function in `build.py` for validating settings it uses etc.

Any other options?

d0ugal:
Closing in favor of having a discussion with code in #493

PR #1287, opened 2017-09-18
Enhancements concerning plugins

funkyfuture:
these are several changes regarding the plugin code that were necessary or i found useful when i wrote a plugin.

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -209,25 +209,30 @@ class RepoURL(URL):
                 config['edit_uri'] = 'src/default/docs/'
 
 
-class Dir(Type):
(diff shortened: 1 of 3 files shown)


waylan:
Awesome! Some good work here. Thanks.

funkyfuture:
oh, i forget a remark; i found that the config validation has its limitations. [Cerberus](http://docs.python-cerberus.org/en/stable/) should offer all functionality needed to normalize and validate thoroughly. if you think another dependency would be okay, i could prepare a proposal at some point. disclaimer: i'm a contributor to Cerberus.

waylan:
@funkyfuture could you bring that up in a new separate issue? For previous discussions see #474, #493 and #496.

PR #1290, opened 2017-09-21
Adds 'load_markdown' event

funkyfuture:
This allows to override the loading of source files, e.g. to fetch it from a remote location or for generated contents.

### DIFF
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -11,9 +11,7 @@ using a plugin which comes with MkDocs, then it was installed when you installed
 MkDocs. However, to install third party plugins, you need to determine the
 appropriate package name and install it using `pip`:
 
-```python
-pip install mkdocs-foo-plugin
-```
+    pip install mkdocs-foo-plugin
 
 Once a plugin has been successfully installed, it is ready to use. It just needs
 to be [enabled](#using-plugins) in the configuration file.
@@ -156,7 +154,7 @@ entire site.
     be used to alter the site navigation.
 
     Parameters:
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
     : __config:__ global configuration object
 
     Returns:
@@ -170,7 +168,7 @@ entire site.
     Parameters:
     : __env:__ global Jinja environment
     : __config:__ global configuration object
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : global Jinja Environment
@@ -274,7 +272,7 @@ event.
     : __markdown:__ Markdown source text of page as string
     : __page:__ `mkdocs.nav.Page` instance
     : __config:__ global configuration object
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : Markdown source text of page as string
@@ -289,7 +287,7 @@ event.
     : __html:__ HTML rendered from Markdown source as string
     : __page:__ `mkdocs.nav.Page` instance
     : __config:__ global configuration object
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : HTML rendered from Markdown source as string
@@ -303,7 +301,7 @@ event.
     : __context__: dict of template context variables
     : __page:__ `mkdocs.nav.Page` instance
     : __config:__ global configuration object
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : dict of template context variables
@@ -319,7 +317,7 @@ event.
     : __output_content:__ output of rendered template as string
     : __page:__ `mkdocs.nav.Page` instance
     : __config:__ global configuration object
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : output of rendered template as string
@@ -358,14 +356,14 @@ Note that registering a plugin does not activate it. The user still needs to
 tell MkDocs to use if via the config.
 
 [BasePlugin]:#baseplugin
-[entry point]: #entry-point
 [config]: configuration.md#plugins
+[entry point]: #entry-point
+[env]: #on_env
 [events]: #events
+[extra_templates]: configuration.md#extra_templates
 [Global Events]: #global-events
 [Page Events]: #page-events
-[Template Events]: #template-events
-[static_templates]: configuration.md#static_templates
-[extra_templates]: configuration.md#extra_templates
-[env]: #on_env
-[post_template]: #on_post_template
 [post_build]: #on_post_build
+[post_template]: #on_post_template
+[static_templates]: configuration.md#static_templates
+[Template Events]: #template-events
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -246,15 +246,29 @@ event.
     Parameters:
     : __page:__ `mkdocs.nav.Page` instance
     : __config:__ global configuration object
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : `mkdocs.nav.Page` instance
 
+##### on_page_read_source
+
+:   The `on_page_read_source` event can replace the default mechanism to read
+    the contents of a page's source from the filesystem.
+
+    Parameters:
+    : __page:__ `mkdocs.nav.Page` instance
+    : __config:__ global configuration object
+
+    Returns:
+    : The raw source for a page as unicode string. If `None` is returned, the
+      default loading from a file will be performed.
+
 ##### on_page_markdown
 
-:   The `page_markdown` event is called after the page is loaded from file and
-    can be used to alter the Markdown source text.
+:   The `page_markdown` event is called after the page's markdown is loaded
+    from file and can be used to alter the Markdown source text. The meta-
+    data has been stripped off and is available as `page.meta` at this point.
 
     Parameters:
     : __markdown:__ Markdown source text of page as string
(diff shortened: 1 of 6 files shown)


waylan:
Interesting. But I do have a few general notes:

1. It would be preferable if all of the 'page events' had the word "page" in their name. 

2.  What does this accomplish that you can't already do with the existing events? Seems to me the `page_markdown` and/or `page_context` events would be able to alter the content.

This API is brand new and mostly unused. If we should be altering the existing events, or replacing them with different events, now is the time to figure that out before we release it. I can't imagine every use case, so I'd like to see a use case for any proposed changes. Without use cases, its more difficult to provide an API which works for a larger audience.

Regarding this specific proposal, it provides a way to replace the default behavior (by overriding file loading). None of our events provide that type of functionality. They all provide ways to alter the data after it is loaded. I understand the attractiveness of each approach, but I think we need to stick with one or the other. If we accept this, then the same should be done for other parts of the API. Again, use cases would help clear up which approach will work better for our goals and the needs of users.

funkyfuture:
my use-case is to generate a markdown stream from multiple yaml-files. so far, i used the `pre_build` event. at that point an actual file is written to disk. this is not necessary, creates fs artifacts and   it significantly increases the build duration.

the `pre_page` and the `page_markdown` events aren't suited as the `Page.load_markdown` method is called between and raises an exception if no file is found.

another approach would be to make the `load_markdown` method more tolerant and return `None` if there's no file. but that would avoid the `meta.get_meta` call and there would be no exception in case of an actually missing file. anyway, i find that just overriding a method that loads a unicode stream is far more simple than to deal with more complex strucures like `Page`.

i don't have my code at hand right now, but roughly it does so:

```python

def on_load_markdown(stream, page, config):
    if page.title == "Catalogue":
        return generate_catalogue_contents()
```

as written in the commit message, other uses of such event could be to fetch the stream from a remote location.

> It would be preferable if all of the 'page events' had the word "page" in their name.

waht would you suggest? `load_page_markdown` seems redundant to me as a page is always derived from a stream formatted with markdown.

funkyfuture:
a'ight, i amended the changed name.

docs/user-guide/plugins.md
```diff
@@ -246,11 +244,21 @@ event.
...
+##### on_load_page_markdown
+
+:   The `load_page_markdown` event can replace the default loading mechanism to 
+    load the contents of the file contents. If the plugin(s) return `None`, the
+    default loading will be performed.
+
+    Parameters:
+    : __page:__ `mkdocs.nav.Page` instance
+    : __config:__ global configuration object
+
```
waylan: Just noticed this is missing the `Returns`  description which exists in the other events. Perhaps the description should specify that the Markdown string should include any meta-data (if it exists) as that has not yet been removed.

docs/user-guide/plugins.md
```diff
@@ -246,11 +244,21 @@ event.
...
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : `mkdocs.nav.Page` instance
 
+##### on_load_page_markdown
+
+:   The `load_page_markdown` event can replace the default loading mechanism to 
+    load the contents of the file contents. If the plugin(s) return `None`, the
```
waylan: Lets change this line :

> load the contents of the file <del>contents </del><ins>from the file system</ins>. If the <del>plugin(s)</del><ins>event</ins> return<ins>s</ins> `None`, the

docs/user-guide/plugins.md
waylan: Perhaps:

    Returns:
        : Content of file as string of Markdown text with any meta-data still attached.

funkyfuture: thanks for spotting. fixed it.

docs/user-guide/plugins.md
```diff
@@ -246,11 +244,24 @@ event.
...
-    : __site_navigation__: global navigation object
+    : __site_navigation:__ global navigation object
 
     Returns:
     : `mkdocs.nav.Page` instance
 
+##### on_load_page_markdown
+
+:   The `load_page_markdown` event can replace the default loading mechanism to
+    load the contents of the file contents.
```
waylan: I realize you removed some of the text I suggested changing (by adding it to the "return" description -- which makes sense), but we still have "contents of the file contents" here. There are to many "contents" which makes it awkward. Let's change it to "contents of the file from the system". I realize your use case may not actually be reading "from the system" but that makes it clear which default behavior this overrides.

funkyfuture: thanks again, i think by employing a subjunctive it is described peoperly.

waylan: But that implies that it should not read from an actual file. As if the loading mechanism **must** fake file reading. Of course, it can do actual file reading.

Let's try this instead:

> The `load_page_markdown` event can replace the default mechanism of reading the contents of a file from the file system.

Here, we are stating what the default is and that it can be replaced. Nowhere do we state what to replace it with. That is left as an exercise for the reader to work out (as it should be). No need to imply artificial limits in the documentation.

Interestingly, by changing the verb to "read" it occurs to me that perhaps a better name for the event would be `on_read_page` (or `on_page_read`) as that is actually what it is doing. "Loading" involves both reading and separating the meta-data. I didn't suggest "load_page" before because that would give the impression of a much larger operation than simply reading the file.

funkyfuture: the proposal lgtm. in my understanding loading and reading would be synonymous in this context, but you're the native speaker. what's bugging me is, that 'page', imo, points to the `Page` class which is the agent here, but not the object that is read. what about `page_read_markdown`?

funkyfuture: that's what i have now, notice the amendment to `on_page_markdown`.

```markdown
##### on_page_read_markdown

:   The `page_read_markdown` event can replace the default mechanism of reading 
    the contents of a file from the file system.

    Parameters:
    : __page:__ `mkdocs.nav.Page` instance
    : __config:__ global configuration object

    Returns:
    : The raw source for a page as unicode string. If `None` is returned, the
      default loading from a file will be performed.

##### on_page_markdown

:   The `page_markdown` event is called after the page's markdown is read and
    can be used to alter the Markdown source text. The meta-data has been 
    stripped off and is available as `page.meta` at this point.
```

waylan: > what's bugging me is, that 'page', imo, points to the Page class which is the agent here, but not the object that is read.

Fair point. Which reminds me of my original objection. We seem to be having trouble fitting this event in with the other events, which suggests that maybe it doesn't belong as-is. Each of us are suggesting slightly different takes on the documentation. In each case, we each are compromising on a different aspect of the implementation. Neither of us are satisfied as there is a shortfall either way. The way I see it, if we can't write documentation that fits clearly, then we don't have a solution that makes sense. And that's why I'm making such a big deal out of this. Normally I wouldn't be so picky about a single sentence in documentation.

So, let me break down how I see this new event [^1] ... 

This event is called when populating an instance of the `Page` class. And this event would be called once for each "page". It even has the (unpopulated) `Page` instance passed in as a parameter. Therefore, it is a "page" event and "page" needs to be in the name. 

Its actual function (by default) is to read the contents of a file. Therefore it is a "read" event and "read" should be in the name.

I had wanted "markdown" in the name before because I hadn't clarified in my mind the distinction  between "reading" and "loading", but now that we are using "read", my reasons for including  "markdown" aren't that compelling. We don't want to confuse this event and the event which immediately follows: `on_page_markdown`. The later  is where modification of the Markdown text itself should happen. Whereas this event should simply be returning the contents of a file (whether a real file or some pseudo file).  It's not really the concern of this event if the contents of the file happen to not actually be Markdown[^2]. Therefore, I think we can drop the use of "markdown" from the name. 

However, to make it clear that the event does not "read" the `Page` instance, but something else, I agree that including a third word after "read" helps[^3]. I considered  `on_page_read_file`, but that feels slightly contradictory as the whole purpose of this event is to allow reading from something other than a file. What about `on_page_read_src`? This event fires when a `Page` instance 'reads' its 'source', regardless of what that source may happen to be.


[^1]: In fact, this comment went through multiple drafts as I have clarified my take on this. Thats often how I develop concepts and processes--by typing out an explanation. And that explains why many of my comments are so lengthy. 

[^2]: For example, a file could contain Markdown mixed with template variables. Those variables should only be swapped out for their values in the `on_page_markdown` event, after the meta-data is available. In that case, this "read" event would not actually return valid Markdown, but the raw file source.

[^3]: That's also why I suppose you suggested `page_read` rather than `read_page`. I agree, that is an important distinction to make, yet one I hadn't considered previously.

funkyfuture:
i went with `page_read_source` and renamed the method accordingly.

waylan:
Looks good. Fix the test failures and I'll merge this.

mkdocs/nav.py
```diff
@@ -243,14 +243,18 @@ def is_homepage(self):
...
-            log.error('file not found: %s', self.abs_input_path)
-            raise
-
-        self.markdown, self.meta = meta.get_data(input_content)
+    def read_source(self, config):
+        source = config['plugins'].run_event(
+            'page_read_source', None, config=config, page=self)
+        if source is None:
+            try:
+                with open(self.abs_input_path, 'r', encoding='utf-8') as f:
```
waylan: You need to use `io.open` here as was used in the code you are replacing. `io.open` guarantees feature parity between Python2 & 3. And then both the Python2.7 and flake8 tests will pass.

funkyfuture:
lol, what's a Freudian incident, i didn't want to make a point here, but i made two. ;-)

closing the file after reading is okay though?

waylan:
Groan. Now the Windows CI tests appear to be [failing][1] do to the use of the now deprecated `ez_setup.py` script. Regardless, that is not the fault of this PR. Thanks for your work on this and patience while I worked my way though it.

[1]: https://ci.appveyor.com/project/d0ugal/mkdocs/build/1377.1.0.dev/job/6xo34ihqylvo1cyy

PR #1302, opened 2017-10-05
Document how to generate/install manpages.

waylan:
This probably needs some improvement as the manpage generation script uses `python setup.py...` but we only document installing via pip so there is no setup.py script to call/run without more work.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -123,7 +123,7 @@ mkdocs, version 0.15.3
     generate and install them for you. Simply run the following two commands:
 
         pip install click-man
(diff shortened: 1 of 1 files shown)


waylan:
I think we should hold off on this until we see click-contrib/click-man#16 accepted and released. Then users will no longer need to call the `setup.py` file. We can simply tell users to run the following commands to generate and install man pages (after MkDocs is installed):

```bash
pip install click-man
click-man --target path/to/man/pages mkdocs
```

PR #1304, opened 2017-10-10
Make the repo link an "edit" link, if edit_uri is set.

### LINKED ISSUE #1129: GitHub repo link appears, but no "Edit on GitHub" link appears
Hello,

Although a number of these options appear to have suitable defaults, I have the following in my `mkdocs.yml` file:

```
repo_name: GitHub
repo_url: https://github.com/chrisdeeming/reponame/
edit_uri: edit/master/docs/
```

I have only just recently updated to `mkdocs, version 0.16.1` and unfortunately no edit link appears (using the default, unmodified theme).

Assuming it should appear in the breadcrumb area, all I see is this:

image

The "GitHub" link just links to `https://github.com/chrisdeeming/reponame/` as per the `repo_url`. Does displaying the edit link require some sort of manual action on each page?

Many thanks,

Chris

jkinkead:
Fixes #1129 .

I'm happy to write a unit test for this, if there are any on the themes. I couldn't spot them, though.

### DIFF
--- a/mkdocs/themes/mkdocs/nav.html
+++ b/mkdocs/themes/mkdocs/nav.html
@@ -71,7 +71,18 @@
               {%- endblock %}
 
               {%- block repo %}
-                {%- if config.repo_url %}
+                {%- if page and page.edit_url %}
+                    <li>
+                        <a href="{{ config.edit_url }}">
+                            {%- if config.repo_name == 'GitHub' %}
+                                <i class="fa fa-github"></i>
+                            {%- elif config.repo_name == 'Bitbucket' -%}
+                                <i class="fa fa-bitbucket"></i>
+                            {%- endif -%}
+                            Edit on {{ config.repo_name }}
+                        </a>
+                    </li>
+                {%- elif config.repo_url %}
                     <li>
                         <a href="{{ config.repo_url }}">
                             {%- if config.repo_name == 'GitHub' %}


waylan:
Thank you for your contribution. This looks good with one problem. It was clearly stated in #1129 (and the linked comment in #269) that we do *not* want to replace the existing link, but add a second new "edit" link elsewhere on the page. Unfortunately, as this proposed change replaces the link, we cannot accept it as-is.

jkinkead:
Do you have any preference for placement? Side-by-side with the existing link?

waylan:
> Do you have any preference for placement?

I suspect this is why we don;t yet have a solution. No one had made that decision. Personally, I haven't given it much thought and don't really have an opinion on where it should go. A few proposals/mock-ups (preferably as screenshots) might make it more clear.

jkinkead:
I'll close this for now; I'll add PR that updates the documentation on the setting, since it's a bit of a time-waste trying to make it work for the default theme.

waylan:
I've reopened this as we should offer at least an option. In fact, this would be a great use of the new theme config settings---config settings specific to a given theme (see 75350da). 

Users seem to have one of the following opinions:

1. The "GitHub link"  should point the "project" on GitHub
2. The "GitHub link" should be a "edit on GitHub" link.
3. Both types of links should be on the page.

So why not provide all three options? Probably leave the default as-is. I'm thinking two boolean options: `show_repo_link` and `show_edit_link`.  If they are both `True`, then the two are displayed side-by-side.

As a reminder, these should be set on the theme only, not in MkDocs proper. And then we probably need to add a section to the documentation for the `mkdocs` theme defining settings unique to that theme.

waylan:
Okay, I just took another look at this. I'm not sure I understand @d0ugal's original objection. As this PR is currently implemented, if the user has a `repo_url` set, then a "repo link" (GitHub/Bitbucket/etc) is displayed. However, if the user also defined an `edit_uri`, then an "edit this page" link is displayed instead. The only use-case not supported is displaying both, which I see very limited value in anyway -- especially if they were to be displayed side-by-side (and there is not really any other logical location to display such a link in the `mkdocs` theme). While we could add more config settings, I don't see any value in doing so. 

Therefore, baring any objection, I'm inclined to accept this PR as-is.

waylan:
Just pushed to #1313 with documentation and release notes updates. Closing this in favor of #1313. Thanks @jkinkead, I used your template changes (with credit).

jkinkead:
Thanks! :)

PR #1306, opened 2017-10-12
Add a note that edit_uri doesn't do anything with the mkdocs theme.

jkinkead:


waylan:
We have every intention of supporting edit links in the MkDocs theme. Therefore, I'm closing this as wontfix.

jkinkead:
It's fairly poor user experience for the documentation to be deliberately incorrect. It cost me an annoying amount of time to track down why this didn't work, and I'm clearly not the only user who's experienced this.

The documentation should stay inline with the project, not inline with the eventual goals of the project.

waylan:
While I generally agree, there are two reasons why I took that action I did here:

1. MkDocs is beta software (until version 1.0) and various pieces are still undocumented. Anything undocumented is subject to change. If we document it, then we have promised to support that behavior going forward. This not a promise we want to make.
2. The documentation (on the website) is only ever updated on a release. If we do add the feature before the next release, then we would need to remember to remove this documentation which has never been published. Our time would be better spent focusing on an actual solution rather than documenting a missing feature.

waylan:
However, it does occur to me that it may be reasonable to note in the documentation that an "edit this page" link will only be generated if the theme you are using supports it. Note that such a statement should not name any specific themes. After all, there are many third-party themes which we have no control over and any of them may or may not support the feature. When you don't see the edit link, at least the documentation would point you to what to check next. 

If you would like to update this PR with a generic statement, I'll reconsider.

PR #1370, opened 2018-01-04
Use exclude keyword to skip markdown files

brandongc:
This change slightly modifies the functionality of the copy_media_files() function by changing the behavior of the unused exclude option to be a list extensions rather than a list of Unix shell patterns.

This function is only called in `mkdocs/commands/build.py` and does not use the functionality.

I also add a variable to define markdown extensions in a single place.

This change will *greatly* simplify life for a plugin developer who wishes to reused this functionality (namely me).

So my plugin to copy markdown files can simply consist of:

```python
utils.copy_media_files(config['docs_dir'], config['site_dir'], exclude = [], dirty=True)
```

in a `post_build` event.

I understand this may get rejected as there is a private refactor in progress for the next release, but please consider allowing for a similarly flexible util function in the future.

### DIFF
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -41,6 +41,14 @@ else:                           # pragma: no cover
 
 log = logging.getLogger(__name__)
 
+markdown_patterns = [
+    '*.markdown',
+    '*.mdown',
+    '*.mkdn',
+    '*.mkd',
+    '*.md'
+]
+
 
 def yaml_load(source, loader=yaml.Loader):
     """
@@ -145,11 +153,12 @@ def clean_directory(directory):
             os.unlink(path)
 
 
-def copy_media_files(from_dir, to_dir, exclude=None, dirty=False):
+def copy_media_files(from_dir, to_dir, exclude=markdown_patterns, dirty=False):
     """
     Recursively copy all files except markdown and exclude[ed] files into another directory.
 
     `exclude` accepts a list of Unix shell-style wildcards (`['*.py', '*.pyc']`).
+
     Note that `exclude` only operates on file names, not directories.
     """
     for (source_dir, dirnames, filenames) in os.walk(from_dir, followlinks=True):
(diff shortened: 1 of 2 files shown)


waylan:
First of all, `exclude` is not unused (see [build.pyL276][1]). It serves a very important function for themes that relies on Unix shell patterns (note that we are passing in more than file extensions). Apparently, we are not properly testing that as no tests are failing with your changes.

Second, why would you want to switch from the more powerful Unix shell patterns to file extensions only? That would be a regression to a less powerful solution.

That said, having Markdown extensions defined in one place makes sense. And using the `exclude` keyword to skip Markdown files seems like it could be a reasonable compromise. I might consider that change as long as we keep support for Unix shell patterns.

[1]: https://github.com/mkdocs/mkdocs/blob/f90c44a20d12635db83021ebd11c31bbac4e28d3/mkdocs/commands/build.py#L276

brandongc:
Strange that I missed that call with `grep` - possibly due to it being multiline.

For the change away from patterns there is no motivation other than it appeared unused and was convenient for passing a list of extensions.

Some design choices:

*  Is it acceptable for `exclude=['.ext']` to be interpreted as `exclude=['*.ext']`?

   If yes then I can add the logic for that to the copy_media_files().

   If no I can change `markdown_extensions=[]` to a list of shell patterns instead of extensions and change `utils.is_markdown_file()` accordingly.

*  Is having a default `exclude=markdown_extensions` ok?

brandongc:
The failed test is `pip install codecov` timing out

mkdocs/utils/__init__.py
```diff
@@ -41,6 +41,14 @@
 
 log = logging.getLogger(__name__)
 
+markdown_patterns = [
+    '*.markdown',
+    '*.mdown',
+    '*.mkdn',
+    '*.mkd',
+    '*.md'
+]
```
waylan: Let's leave these as "extensions" (and call them that).

mkdocs/utils/__init__.py
```diff
@@ -145,11 +153,12 @@ def clean_directory(directory):
             os.unlink(path)
 
 
-def copy_media_files(from_dir, to_dir, exclude=None, dirty=False):
+def copy_media_files(from_dir, to_dir, exclude=markdown_patterns, dirty=False):
```
waylan: Here you can use a list comprehension to convert the list of extensions to patterns:

```python
 exclude=['*{0}'.format(x) for x in markdown_extensions]
```

mkdocs/utils/__init__.py
```diff
@@ -222,13 +230,7 @@ def is_markdown_file(path):
...
     """
     ext = os.path.splitext(path)[1].lower()
-    return ext in [
-        '.markdown',
-        '.mdown',
-        '.mkdn',
-        '.mkd',
-        '.md',
-    ]
+    return any(fnmatch.fnmatch(ext, pattern) for pattern in markdown_patterns)
```
waylan: I like this, but I don't think we need to break out the extension from the path. Just pass `path` to `fnmatch` and remove line 232.

waylan: Or, leave it unchanged as we have a list of extensions, not patterns. It probably should come down to which is more performant. If `fnmatch` fast enough, then converted the extensions to pattern for each call might not matter. Otherwise, don't use `fnmatch` here.

waylan:
I restarted the timed-out job on Travis and it is good now.

PR #1376, opened 2018-01-13
Paths in config file are relative to the config file.

### LINKED ISSUE #543: Paths are relative to you, not the config file
You can do `mkdocs build --config-file=path/to/mkdocs.yml` but any configuration values in that file are relative to your current working directory.

This seems wrong to me, but it's only something I've ran into when trying to automate MkDocs builds (for integration testing). I'm not sure if we want to change this, or what people really expect. The common use case is that you are in the same directory as the config file anyway, so that doen't matter.

Something like this should be changed or at least decided for a 1.0 release.

Thoughts?

djeebus:
Ran into #543 today, saw that no one had started working on this, so decided to take a stab at it. I used `pre_validation` to coerce relative paths to absolute paths, which worked for my use case. 

Two things I wasn't sure of where to go with:
1. What happens if `Config.load_file` is passed something without a filename? I found a few tests where that was the case, but couldn't think of a use case otherwise. I reverted to the current behavior in this scenario (relative paths are left alone, so they're relative to the current working directory).
2. A few of the tests relied on the current behavior (they processed config files in `/tmp`, and relied on  the existence `./docs` in order to pass validation. I cleaned up the tests, but used `tempfile.TemporaryDirectory`, which doesn't exist in py<3.2. Installing `backports.tempfile` took care of that, but made the imports a little more complex (you'll see them in the `*_tests.py` files).

Thanks for making this project available!

Fixes #543

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -25,7 +25,7 @@ The current and past members of the MkDocs team.
 
 ### Major Additions to Development Version
 
-#### Path Based Settings are Relative to Configuration File (#543).
+#### Path Based Settings are Relative to Configuration File (#543)
 
 Previously any relative paths in the various configuration options were
 resolved relative to the current working directory. They are now resolved
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -25,6 +25,25 @@ The current and past members of the MkDocs team.
 
 ### Major Additions to Development Version
 
+#### Path Based Settings are Relative to Configuration File (#543).
+
+Previously any relative paths in the various configuration options were
+resolved relative to the current working directory. They are now resolved
+relative to the configuration file. As the documentation has always encouraged
+running the various MkDocs commands from the directory that contains the
+configuration file (project root), this change will not affect most users.
+However, it will make it much easier to implement automated builds or otherwise
+run commands from a location other than the project root.
+
+Simply use the `-f/--config-file` option and point it at the configuration file:
+
+```sh
+mkdocs build --config-file /path/to/my/config/file.yml
+```
+
+As previously, if no file is specified, MkDocs looks for a file named
+`mkdocs.yml` in the current working directory.
+
 #### Refactor Search Plugin
 
 The search plugin has been completely refactored to include support for the
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -21,14 +21,14 @@ class Config(utils.UserDict):
     for running validation on the structure and contents.
     """
 
-    def __init__(self, schema):
+    def __init__(self, schema, fname=None):
         """
         The schema is a Python dict which maps the config name to a validator.
         """
 
         self._schema = schema
         self._schema_keys = set(dict(schema).keys())
-        self.fname = None
+        self.fname = fname
         self.data = {}
 
         self.user_configs = []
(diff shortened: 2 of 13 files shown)


waylan:
Awesome! I'll have to take a closer look when I have some time.

djeebus:
Just found an issue: plugins aren't getting normalized properly. Should have it fixed soon too.

djeebus:
All ready now!

mkdocs/config/base.py
```diff
@@ -21,13 +21,14 @@ class Config(utils.UserDict):
...
 
-    def __init__(self, schema):
+    def __init__(self, schema, fname=None):
         """
         The schema is a Python dict which maps the config name to a validator.
         """
 
         self._schema = schema
         self._schema_keys = set(dict(schema).keys())
+        self.fname = fname
```
waylan: I'm not crazy about naming this `fname`. Considering its intended use, we don't care about the "name" of the file, but rather the "path" to the file. Using `fname` indicates to me that I only need to pass in the name of the file, not the full path. However, in actual use, I need to pass in the absolute path (and it might even work if I left off the filename). Therefore, let's indicate that with the name used.

mkdocs/config/config_options.py
```diff
@@ -291,6 +291,24 @@ def __init__(self, exists=False, **kwargs):
...
+            return
+
+        if config.fname is None:
+            # Unable to determine absolute path of the config file; fall back
+            # to trusting the relative path
+            return
+
+        config_dir = os.path.dirname(config.fname)
+        value = os.path.join(config_dir, value)
+        config[key_name] = value
```
waylan: Rather than a pre-validation method, perhaps an `abspath` method would make more sense. Then anywhere we need an absolute path, just pass the value into the `self.abspath` method. And let's put it on the `BaseConfigOption` class so its available everywhere.

Note that you missed the `custom_dir` sub-setting of the theme setting [here][1]. That is not in a `FilesystemObject` subclass. And its entirely possible that Plugins could use various of the `BaseConfigOption` sub-classes or even create their own. to validate their own sub-settings. The point is, this ability to convert a relative path to an absolute path should be available from everywhere.

[1]: https://github.com/eventray/mkdocs/blob/e7d1c70eaee033e7735a5bae3f32293e200982ee/mkdocs/config/config_options.py#L449

mkdocs/config/base.py
djeebus: Maybe `config_path`?

mkdocs/config/config_options.py
waylan: And I just realized the problem with this. The config (which contains the base path) is only available from the `pre_validation` and `post_validation` methods. Therefore, we can only do this from there.

waylan:
The py33 tests are being handled in #1453. Guess we can move on with this issue by updating the docs.

docs/about/release-notes.md
```diff
@@ -25,6 +25,25 @@ The current and past members of the MkDocs team.
 
 ### Major Additions to Development Version
 
+#### Path Based Settings are Relative to Configuration File (#543)
+
+Previously any relative paths in the various configuration options were
+resolved relative to the current working directory. They are now resolved
+relative to the configuration file. As the documentation has always encouraged
```
flynneva, line 32: @waylan I couldn't find it anywhere in the docs, but what about for inherited config files? Are the paths in those inherited config files relative to the parent config file or the child config file?

oprypin: Please don't necro-bump and especially don't tag people while doing so.

The answer is here
https://github.com/mkdocs/mkdocs/blob/ae4a639f6e6fa3648e033f8d2f2cb7679886887b/docs/user-guide/configuration.md?plain=1#L1047-L1048

oprypin: (you can ask a question as a Discussion instead)

PR #1379, opened 2018-01-16
Add a title attribute to the search input

Rudloff:
Currently the input does not have any label or title which means is is not correctly identified by assistive technologies such as screen reader.
(You can read more about it in [this WCAG guideline](https://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140408/H65).)

### DIFF
--- a/mkdocs/themes/mkdocs/search-modal.html
+++ b/mkdocs/themes/mkdocs/search-modal.html
@@ -12,7 +12,7 @@
                 </p>
                 <form role="form">
                     <div class="form-group">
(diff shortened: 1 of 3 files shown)


waylan:
This change should address both of the included themes. Could you add this to the `mkdocs` theme as well?

As an aside, we'll commit this now (assuming the above request is met) to get the benefits ASAP, but I'm concerned this might get missed/lost in the upcoming [search refactor][1].

Rudloff:
I updated my PR. I also added it in `readthedocs/search.html` (I missed it before).

PR #1380, opened 2018-01-16
Defer scripts

Rudloff:
Currently, page rendering is delayed until all scripts are loaded. But these scripts are not essential to the main content.
The `defer` attribute allows the browser to render the page before the scripts are completely loaded so the user can see the webpage earlier.

### DIFF
--- a/mkdocs/themes/mkdocs/base.html
+++ b/mkdocs/themes/mkdocs/base.html
@@ -33,9 +33,9 @@
             <script src="https://oss.maxcdn.com/libs/respond.js/1.3.0/respond.min.js"></script>
         <![endif]-->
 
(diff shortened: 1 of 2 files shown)


waylan:
I'm not opposed to this change, but I suspect the search script is cause for much more of a delay that these scripts are. Of course, that is being addressed as part of a larger [search refactor][1] in #859.

Regardless, this sort of change should be done in both of the included themes. If you could add the same changes to the `mkdocs` theme, then I'll accept this.

Rudloff:
Indeed it won't help with freezes caused by the search script but it makes every page appear faster when loaded over a slow 3G connection.

I updated my PR to include a change for the mkdocs theme too.

PR #1388, opened 2018-01-20
Prevent reinjection of extra Javascript files on reload

yeraydiazdiaz:
The search plugin would add its extra Javascript on each reload polluting the HTML


waylan:
Thanks for this. I noticed the problem when working on something else and then promptly forgot about it.

PR #1389, opened 2018-01-21
Workaround Safari bug when zooming to < 100%

### LINKED ISSUE #1050: Collapsing sidebar when zooming out
**Zooming out** with Safari 10.0 on macOS Sierra **collapses the side bar** of the MkDocs, Cinder, Yeti, etc. styles into the main text.

cinder

yeraydiazdiaz:
Fixes #1050 

Tested on Safari 11.0.2 on Mac OS X Sierra.

Haven't tried with Cinder but I assume it's the same problem, Safari simply ignores the `width: 25%` of the `col-md-3` and pushes the `col-md-9` over it. Overriding the padding and width seems to work without affecting other browsers. It's a fairly hideous workaround but aren't they all.

### DIFF
--- a/mkdocs/themes/mkdocs/css/base.css
+++ b/mkdocs/themes/mkdocs/css/base.css
@@ -181,6 +181,13 @@ footer {
 
 /* Show and affix the side nav when space allows it */
 @media (min-width: 992px) {
+    /* Workaround a Safari bug when zooming to < 100%
+       https://github.com/mkdocs/mkdocs/issues/1050 */
+    .col-md-9 {
+        box-sizing: border-box;  /* csslint allow: box-sizing */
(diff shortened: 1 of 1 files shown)


waylan:
@yeraydiazdiaz thanks for these. We don't really have anyone working on front end stuff right now, so any and all help in the area is much appreciated.

lebigot:
Thanks @yeraydiazdiaz! Did you report this bug to Apple as well (unless already reported)?

yeraydiazdiaz:
@waylan no worries, thanks to Tom and all the contributors for your efforts :)

@lebigot I haven't, no, I'll see if I can test it on a newer version of Safari first

Issue #1172, opened 2017-03-17
Document that all internal URLs must be relative


zxcware:
In a site, referring an image with an absolute url, for example, `/img/the-image.png`, doesn't work. The site interprets it into `http://gobblin.readthedocs.io/img/the-img.png` but expects `http://gobblin.readthedocs.io/en/latest/img/the-img.png`. Why is there the part `/en/latest` injected into the url? I searched through the documentation but didn't find any related configuration. 

Semantically, that absolute url should work. The root of the site should be  `http://gobblin.readthedocs.io/en/latest/` instead of `http://gobblin.readthedocs.io/` in this case.

Any thought?

waylan:
All internal URLs must be relative not absolute. Issue #192 addresses whether we should allow absolute URLs. This could be a duplicate of that; however, I checked the [documentation][1] and noticed that while all of the examples use relative URLs, we don't explicitly state that they must be relative. Therefore, I will leave this open to remind us to clarify the documentation regardless of what decision is made in #192.

[1]: http://www.mkdocs.org/user-guide/writing-your-docs/#images-and-media

PR #1392, opened 2018-01-23
Refactor writing-your-docs.md

### LINKED ISSUE #1385: Is the flavor of Markdown used a secret?
I can't find a single reference in any of the docs to what flavor of markdown is supported and what tokens are acceptable.  

I thought that would be one of the very first and most critical things documented prominently for this type of project as there is obviously no such thing in markdown as a "standard", CommonMark is the closest thing we have today which unfortunately MKDOCS doesn't appear to support.

There is a help section entitled "Writing your docs" subtitled "How to write and layout your markdown source files." 

**That** is where I would have expected to see some indication of what markdown I can use.

waylan:
Present the sections in a logical manner:

1. File Layout (where to save documents)
   * Configure Navigation...
2. Writing with Markdown
   * Supported extensions...

Added documentation for all included extensions. Also added some basic
information about Markdown; pointing to the implementation and rules.
Fixes #1385.

### DIFF
--- a/docs/user-guide/writing-your-docs.md
+++ b/docs/user-guide/writing-your-docs.md
@@ -1,67 +1,16 @@
 # Writing your docs
 
-How to write and layout your markdown source files.
+How to layout and write your Markdown source files.
 
 ---
 
-## Configure Pages and Navigation
-
-The [pages configuration](/user-guide/configuration.md#pages) in your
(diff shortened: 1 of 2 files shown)


waylan:
Also fixed #316 and #1172.

Issue #1394, opened 2018-01-27
Low frame rate when scrolling on large screens


rpkilby:
Scrolling the docs on a large screen (2560x1440) has a very noticeable impact on frame rate (drops to between 10 and 20 fps). This is due to using `background-attachment: fixed` for the body's background image. Disabling that property fixes the issue. Of course, that also disables the fixed background. 

There is a solution presented in [this article](https://www.fourkitchens.com/blog/article/fix-scrolling-performance-css-will-change-property/), but it's not clear to me how to adapt this to a background for the entire body.

yeraydiazdiaz:
Hi @rpkilby, thanks for reporting this, I made a quick stab at implementing your suggested solution but I can't test it myself on big screen, any chance you could test it?

PR #1395, opened 2018-01-27
Avoid `background-attachment: fixed` for better performance

yeraydiazdiaz:
Attempt at addressing #1394 

I can't test it myself on a big screen, could @rpkilby or others test if it indeed solves the problem?

### DIFF
--- a/mkdocs/themes/mkdocs/css/base.css
+++ b/mkdocs/themes/mkdocs/css/base.css
@@ -1,8 +1,18 @@
 body {
     padding-top: 70px;
-    background: url(../img/grid.png) repeat-x;
-    background-attachment: fixed;
+}
+
+body::before {
+    content: ' ';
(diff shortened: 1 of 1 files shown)


rpkilby:
Thanks @yeraydiazdiaz - that works. 

For whatever reason when I was mucking around earlier, I couldn't get the styles to apply correctly. 

¯\\\_(ツ)_/¯

yeraydiazdiaz:
Glad to hear that 👍

rpkilby:
Also, not sure if it's worth adding a note explaining the change. eg,

```css
/* Replacement for `body { background-attachment: fixed; }`,
   which has noticeable performance issues on large displays. */
```

waylan:
Thank you @rpkilby for the report and suggested comment. And thank you @yeraydiazdiaz for the fix.

Issue #1218, opened 2017-04-30
require.js Error: Mismatched anonymous define() module 


lierisme:
# use MkDocs

## mydocs.yml:
site_name: My Docs

extra_javascript:
    - https://cdnjs.cloudflare.com/ajax/libs/raphael/2.2.7/raphael.min.js
    - https://cdnjs.cloudflare.com/ajax/libs/flowchart/1.6.5/flowchart.min.js
markdown_extensions:
    - pymdownx.superfences

## test.md
\```flow
st=>start: Start:>http://www.google.com[blank]
e=>end:>http://www.google.com
op1=>operation: My Operation
sub1=>subroutine: My Subroutine
cond=>condition: Yes
or No?:>http://www.google.com
io=>inputoutput: catch something...

st->op1->cond
cond(yes)->io->e
cond(no)->sub1(right)->op1
\```

## run: mkdocs serve
http://127.0.0.1:8000

## result:
**Uncaught Error: Mismatched anonymous define() module: function ()** ......

why???
help!
thanks!

facelessuser:
This is better suited for the repo containing SuperFences. https://github.com/facelessuser/pymdown-extensions

waylan:
> This is better suited for the repo containing SuperFences. 

I agree. Please follow up there.

lierisme:
Thank you for your.

facelessuser:
@waylan I do think this might need some kind of fix in the default `mkdocs` theme.  It appears some scripts don't play nice with being loaded after `require.js` and the `mkdocs` theme loads all extra scripts after `require.js`.  I'm not sure if and how you'd guys would approach this, but the current setup limits what can be loaded in extra javascript.  So feel free to reopen this if needed, but this seems to be an issue that can only be worked around by overriding the `mkdocs` theme `base.html`.  See comment https://github.com/facelessuser/pymdown-extensions/issues/67#issuecomment-298495186.

facelessuser:
Official statement from require.js on this error: http://requirejs.org/docs/errors.html#mismatch.

waylan:
IIRC, we only use `require.js` for search, which is getting a refactor. I had every intention of trying to remove the dependency on `require.js` in the refactor. This is another reason to add to the list. See [Search Refactor](https://github.com/mkdocs/mkdocs/projects/1) for more info.

facelessuser:
[...]

PR #1396, opened 2018-01-27
Use web worker to build Lunr index

yeraydiazdiaz:
Addresses #1218, #859 and #1127 

Reading the discussion on #859 I decided to give the Web Workers API a try and got a working solution using them. In the process I managed to remove require.js and mustache from the dependencies which seemed overkill for our use case.

I'm not completely replacing the previous code since it's possible the user's browser [might not the Web Workers API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API#Browser_compatibility). Based on our target browser versions we can discuss if we  load or remove the previous version.

I also realised it'd be quite simple to upgrade to the current version Lunr and maybe include [additional language support](https://lunrjs.com/guides/language_support.html) but would require some changes to the configuration, we can discuss it outside the context of this PR.

### DIFF
--- a/mkdocs/contrib/legacy_search/__init__.py
+++ b/mkdocs/contrib/legacy_search/__init__.py
@@ -22,9 +22,8 @@ class SearchPlugin(BasePlugin):
         if not ('search_index_only' in config['theme'] and config['theme']['search_index_only']):
             path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'templates')
             config['theme'].dirs.append(path)
-            for extra_js in ('search/require.js', 'search/search.js'):
-                if extra_js not in config['extra_javascript']:
-                    config['extra_javascript'].append(extra_js)
+            if 'search/main.js' not in config['extra_javascript']:
+                config['extra_javascript'].append('search/main.js')
         return config
 
     def on_pre_build(self, config, **kwargs):
--- a/mkdocs/contrib/legacy_search/templates/search/main.js
+++ b/mkdocs/contrib/legacy_search/templates/search/main.js
@@ -1,0 +1,63 @@
+function getSearchTermFromLocation() {
+  var sPageURL = window.location.search.substring(1);
+  var sURLVariables = sPageURL.split('&');
+  for (var i = 0; i < sURLVariables.length; i++) {
+    var sParameterName = sURLVariables[i].split('=');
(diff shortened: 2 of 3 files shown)


waylan:
Awesome. I am going to hold off merging this though. This has been part of a [larger refactor][1] of search and I'd like to get everything worked out before committing to how any one part is implemented. For example, I didn't merge #1061 because I wanted it to work with the webworker stuff (it's also out-of-date but that's another matter).

My intention is to build a **new** search plugin and leave the existing "legacy search" plugin as-is. The two can then reside side-by-side.

yeraydiazdiaz:
That's fine, I was trying to see if the approach would work and how much of the current search funcionality we actually need, turns out it's not a lot.

How would you like to proceed? Feels these changes should be part of #1061 which in turn is part of the larger refactor, but its PR is against master, so I'm a little confused. Is there a branch for this refactor that I'm missing?

waylan:
#1061 was against master before search was broken out into a plugin and was also an experiment to see if the approach would work (it does, although I would do some things a little differently). Today it only exists as a point of reference. I would be more inclined to start working off of this and reimplement #1061 on top of this. In fact, I'm reopening this so it doesn't get lost and forgotten about.

yeraydiazdiaz:
Sounds good, let me know if you'd like some help with it 👍

waylan:
I took this and re-implemented it as a new plugin in the [waylan:new-search][1] branch (see https://github.com/mkdocs/mkdocs/compare/master...waylan:new-search for comparison). I left bea4fcf20 as a separate commit to show how I implemented the non-worker fallback ('ll probably squash it later).

I should note that all of the URLs need to be relative as the root of the site may not be hosted at the server root (ex: `username.github.io/projectname/`). This turned out to be the trickiest thing to work around (when loading the index from the worker the JSON files URL needs to be relative to the worker, but when loading from the main thread, the JSON file's URL needs to be relative to the page). If course, once I had a solution, it was an easy fix, but it may be non-obvious from the code why I did that. 

The entire thing is a work in progress. I've updated the [project][2] to reflect what has been resolved and what is outstanding.  Any feedback and/or improvements are welcome.

[1]: https://github.com/waylan/mkdocs/tree/new-search
[2]: https://github.com/mkdocs/mkdocs/projects/1

PR #1400, opened 2018-02-03
Remove last slash from repository URL

raimon49:
I think that it is more natural for this repository URL linked from the project document to have no last slash.

For example, the last with no slash URL even search results of Google have been indexed.
google-search

What do you think?

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -3,7 +3,7 @@ site_url: http://www.mkdocs.org
 site_description: Project documentation with Markdown.
 site_author: MkDocs Team
 
-repo_url: https://github.com/mkdocs/mkdocs/
+repo_url: https://github.com/mkdocs/mkdocs
 edit_uri: ""
 
 pages:


waylan:
Actually, those of us who are old-school think the slash is more natural. Used to be you couldn't have a URL that pointed to a directory that didn't end with a slash. If a URL did not end with a slash then it pointed to a file (with a file extension). Of course, today servers are more flexible and can be configured in all sorts of ways.

Regardless, Github does not redirect to the URL without the slash (which would make for unnecessary bandwidth usage) so there is no herm in leaving it. And, the URL itself doesn't get displayed anywhere (so who cares what "looks" better). Finally, when generating "Edit on GitHub" links for pages, the slash would need to be added anyway when joining the path to a given file.

In other words: who cares.

raimon49:
@waylan 

> Actually, those of us who are old-school think the slash is more natural.

OK, I understood very well.

Thank you for your reply.

Issue #778, opened 2015-12-23
External links in menu


chipsenkbeil:
Is there any workaround for adding external links in the pages menu generated by mkdocs? I've got a Scala project that is using mkdocs and have some externally-hosted Scaladoc content that I'd love to be able to reference in the menu itself. 

My current workaround is to create pages for the documentation with metadata and Javascript to redirect to the documentation.

d0ugal:
You could add custom JavaScript and have that add the links you want. http://www.mkdocs.org/user-guide/configuration/#extra_javascript

Otherwise, what you have done works. There isn't anything in MkDocs that supports this.

gsouf:
Adding javascript to redirect is an ugly workarround, and that needs to set a page that will live here for nothing. 

Could you consider reopening the issue?

waylan:
Actually, the proper way to do this would be with a custom theme. You could insert the external links directly in the theme. I suppose, with some work on #607 this could be even easier.

Outside of that, the only way to support such a feature would be by creating a new setting, which we do not want to do (we would prefer to have less settings, not more). Perhaps a Plugin could address this if/when that becomes available (see #206).

In either case, I see no reason to open this issue. There is nothing for us to do outside of those other issues.

gsouf:
@waylan  thanks,  about the plugin API, is there any contribution than can be done to help to speed up its development?

waylan:
@gsouf unfortunately, I don't have a good answer for that question. Neither @d0ugal or myself have a big block of time to spend reviewing the Plugin API and we don't want to only create a half-built system (which might not work when the other half gets implemented later). I realize that we have punted quite a large number of feature requests to the Plugin API and we should probably focus out attention there, but without the time (we work on this in our spare time), I can't give any indication of when things will move forward on that front.

PR #1406, opened 2018-02-15
Allow broken Nav without failing building the rest of the docs

seanch87:
We've noticed when many teams are contributing to a single mkdocs file, broken TOC files will eventually happen. and the mkdocs build breaks. It'd be nice to have the option to continue the build, and replace broken TOC pages with a 404 page. 

Looking for input on what exactly that 404 page should look like, for now I've literally just put the text "404" on the page!

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -23,7 +23,21 @@ The current and past members of the MkDocs team.
 
 ## Development Version
 
+* Compress `sitemap.xml` (#1130).
+* Defer loading JS scripts (#1380).
+* Add a title attribute to the search input (#1379).
+* Update RespondJS to latest version (#1398).
+* Always load Google Analytics over HTTPS (#1397).
+* Improve scrolling frame rate (#1394).
+* Provide more version info. (#1393).
+* Refactor `writing-your-docs.md` (#1392).
+* Workaround Safari bug when zooming to &lt; 100% (#1389).
+* Remove addition of `clicky` class to body and animations. (#1387).
+* Prevent search plugin from reinjecting `extra_javascript` files (#1388).
 * Refactor `copy_media_files` util function for more flexibility (#1370).
+* Remove PyPI Deployment Docs (#1360).
+* Update links to Python-Markdown library (#1360).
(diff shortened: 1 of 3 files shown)


waylan:
First of all, the "TOC" is the list of headers within a page. You appear to be referring to the "`pages` config setting" which we often refer to as the "nav" . Any such references to "toc" should be updated to refer to the "nav".

Second, we are very resistant to adding new settings. I question whether this new `allow_missing_pages` setting is really needed. We already have a `strict` mode for the build which fails on warnings. Seems like that might be a sensible approach (missing page issues a warning and that warning results in an error only in strict mode).

Third, I'm not sure we want to do this at all. As per #699 we intend to support building pages which are not listed in the `pages` config setting. The config setting will simply become a mechanism to define the global navigation for the site. Why would anyone add items to the nav which don't exist? Seems backwards to me. In any event, as per #778, we also will be adding the ability to include external links in the nav, so at that point, we will have items in the nav which we won't even know if they exist.

Given the above, my inclination is to address this with #699 and #778, if at all. Therefore, it makes no sense to apply this patch (or update it). I'll leave this open and add it to the Pages Refactor project so we don;t forget about it.

waylan:
Oh and we would not create the page with the body of "404" or anything else (the themes already provide a 404 page). Instead, simply move on to the next page. But for that to happen we need #699 which separates the page building from the nav.

seanch87:
thanks for the quick response! 

1) re: toc vs pages, understood
2) rather than a new config i can leverage the strict config setting
3) with our builds, we have teams adding misconfigured toc.yml's in their respective project which is breaking the whole site. i'd like to be able to relax this so other documentation can still be updated 
4) noted on the 404 page, maybe I can just have it display the theme's 404 pagr for now?

waylan:
> maybe I can just have it display the theme's 404 pagr

It already does for any non-existent page which is requested as long you the server is configured correctly (the dev server is).

> for now

Unless you are forking locally, there is no need. The next release we do will have the above discussed changes. An interim fix will never make it to a release.

PR #1410, opened 2018-02-21
Expand {sha} and {version} is user commit message

abitrolly:


waylan:
Interesting. I didn't even realize this was done with the default message. We should add some documentation that this is supported [here](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/__main__.py#L85).

waylan:
Also, a line should be added to the [release notes](https://github.com/mkdocs/mkdocs/blob/master/docs/about/release-notes.md#development-version)

abitrolly:
Done.

mkdocs/__main__.py
```diff
@@ -83,7 +83,7 @@ def common_options(f):
 no_reload_help = "Disable the live reloading in the development server."
 dirty_reload_help = "Enable the live reloading in the development server, but only re-build files that have changed"
 commit_message_help = ("A commit message to use when commiting to the "
```
trel: not part of this original PR, but this should also be 'committing' with two t's

PR #1418, opened 2018-02-27
Refactor search plugin

### LINKED ISSUE #826: utf-8 search support

### LINKED ISSUE #859: Large search indexes can cause Lunr to freeze the page while it creates the index.

### LINKED ISSUE #867: word separators in searching
Suppose in the markdown file, I have a "word": "Foo.bar", the current search function is not able to pick "Foo.bar" if I search for "bar". May be some settings in lunr.js will do the job?

### LINKED ISSUE #1127: Run search only for query having 3 and more characters

### LINKED ISSUE #1218: require.js Error: Mismatched anonymous define() module

### LINKED ISSUE #1319: Upgrade to lunr 2.x
At some point recently, lunr 2.0 was released. A summary of changes are here: https://lunrjs.com/guides/upgrading.html

More importantly, there is a page on [Pre-Building Indexes](https://lunrjs.com/guides/index_prebuilding.html) which includes an example script. Unfortunately, I still don't see a spec (of the index format) so that a Python lib could reliably build the index.

In any event, when search is refactored, we should use lunr 2.x.

waylan:
* Use a web worker in the browser with a fallback (fixes #1396).
* Optionally pre-build search index (fixes #859 & closes #1061).
* Upgrade to lunr.js 2.x (fixes #1319).
* Support search in languages other than English (fixes #826).
* Allow the user to define the word separators (fixes #867).
* Only run searches for queries of length > 2 (fixes #1127).
* Remove dependency on require.js (fixes #1218).
* Compress the search index (fixes #1128).


waylan:
Anyone reviewing this (especially theme authors @squidfunk) should take a close look at the structure of the `search_index.json`. The docs explain:

> The search index is written to a JSON file at `search/search_index.json` in the
> [site_dir]. The JSON object contained within the file may contain up to three
> objects.
> 
> ```json
> {
>     config: {...},
>     data: [...],
>     index: {...}
> }
> ```
> 
> If present, the `config` object contains the key/value pairs of config options
> defined for the plugin in the user's `mkdocs.yml` config file under
> `plugings.search`. The `config` object was new in MkDocs version *1.0*.
> 
> The `data` object contains a list of document objects. Each document object is
> made up of a `location` (URL), a `title`, and `text` which can be used to create
> a search index and/or display search results.
> 
> If present, the `index` object contains a pre-built index which should load much
> faster than building the index in the browser on each page load. Note that the
> pre-built index is only created if the necessary dependencies (node.js) are
> available when `mkdocs build` is run. Any failures are ignored silently with the
> expectation that the index can be built in the browser. The `index` object was
> new in MkDocs version *1.0*.

Note that the `data` object preexisted and any existing third-party themes which build their own index should continue to work as-is. Themes may choose to make no modifications and continue to ignore the other items or, check for their existence and only act on them if they exist. 

A word of caution: if the user does not set `prebuild_index=False` in their `plugins.search` config, the prebuilt index will be included even if the theme doesn't need it. While the theme can easily ignore it, that could potentially waste a lot of bandwidth. There is currently no way to set/override that setting from the themes config. In my experience, the pre-built index significantly increases the size of the file for a small site, but has a relatively small effect on the size of the file for a large site. See the discussions in #859 and #1061 for more details about that.

squidfunk:
Thanks for the notice! So as this is a downward compatible change I assume that everything should work as before. Can the thme disable the building of the index? Material supports 23 languages with different configurations and stemming support - how is this reflected in the prebuilt index? Is there only support for English?

squidfunk:
Ok so I read your comment again and the theme can not disable it - that's pretty bad, because some very huge projects use Material and they would all have to set the flag explicitly - can we add some way for the theme like we did with the serving of the search assets etc? See https://github.com/squidfunk/mkdocs-material/blob/2a48e17d2dfb0199fe482c539acc8ece9a8da304/src/mkdocs_theme.yml#L69

waylan:
@squidfunk the plugins new config options are all documented [here].(https://github.com/mkdocs/mkdocs/pull/1418/files#diff-e3c9b75ef08163744e78a99035b50be2)

As you can see, the new plugin supports the 17 languages supported by lunr-langauge (lunr-language recently added Thai which material appears to be missing).  You said material supports 23 languages, but I see 16 [listed in your docs][0] (23 languages are listed under [Localization][1]). Of course, the languages would need to be listed in the `plugins.search.lang` config setting which is duplicative of your theme's setting. As an aside, I'm running validation on the values provided to the plugin settings and raising a ValidationError on any "unknown" languages. So, for example, listing `zh` in `plugins.search.lang`  would fail. I did consider allowing the user to list `zh` and just using the `ja` stemmer under the hood, but decided against it.

I hesitate to make the `plugins.search.prebuild_index` setting controllable by a theme because there would be two places to set it. Which one get's precedence? If the theme overrides the plugin setting, then the user can't use the plugin setting to override the theme if she really wants to. Or do we not want to allow the user to do that? I wanted to copy the `theme.search_index_only` setting to `plugins.search.index_only`, but we would have the same duplication problem so I didn't.

I notice that material includes support for an `extra.search.tokenizer` setting. And the new search plugin supports the same thing at `plugins.search.separator`. However, the two are ignorant of each other, which could confuse users. In that case, I expect that material could be updated to simply check for the `plugins.search.separator` setting first, and if not present, fall back to `extra.search.tokenizer`. However, not all settings are that straightforward to deal with.

It may help to remember that search is a plugin. While it is included with MkDocs, I expect most plugins to be third party. As some third party plugins become more popular, themes may need to support them. However, MkDocs won't be able to do anything special for individual settings of third party themes. The current system of `theme.search_index_only` probably won't carry through to other plugins very well. Users will need to set those sorts of things themselves in the `plugins` settings for the plugins they are using. Whatever solution we come up with should take this larger picture into consideration.

One possible solution I considered is to make all plugin settings overridable by a theme. Using some nesting or prefix, the theme settings could indicate they are for a given plugin and if set, would override the theme's defaults. The problem is that such a solution would not likely be backward compatible with the existing config scheme. While it would give you a lot more control long-term, it would be more work up front for material to support and would requires existing users to make a lot of changes to their configs when they upgrade to the next release. Of course, the alternative ti leave things as-is and require users to manually configure the plugins settings to work with they theme. Any thoughts?

[0]: https://squidfunk.github.io/mkdocs-material/getting-started/#site-search
[1]: https://squidfunk.github.io/mkdocs-material/getting-started/#localization

squidfunk:
@waylan Well, I think upgrade complexity and necessary effort are very important aspects. When I'm running an Open Source project as large as Material I try always to minimize breakage across versions, even minor versions. I try to keep them to major releases and so far there have been two.

I don't want my users to bother with more setup than they currently have, only to support search which all users take for granted and which works very well for Material. Introducing such a change would incur a silent penalty (prebuilt index is on by default) that probably nobody would notice. It would mean that the setup for every theme is getting more complex and there are more potential traps to fall into. There are a lot of users asking simple questions on the issue tracker already.

When you want themes to adopt different plugins, you also need themes to explicitly state what they want to support for those plugins. This is the very reason we have the search index configurations in the top-level theme object which were added in 0.17.1. It would be perfect if themes can define which features they support and turn on/off default flags for plugins, if they are available. Otherwise, if the user specifies the plugin themselves, I would expect that the user's choices take precedence.

The configuration options I added are partly there because MkDocs doesn't support some features. Search is probably a topic which Material addresses in a very specific way and therefore added a lot of options. I will do my best to merge them with the new options for the search plugin.

Regarding the languages - Material integrates the languages from lunr-languages and adds support for more languages by configuring stemmers, trimmers and tokenizers in specific ways - see the language files. Language support doesn't include stemming, yes, but otherwise search wouldn't work at all, that's the reason Material does that.

waylan:
> I think upgrade complexity and necessary effort are very important aspects. When I'm running an Open Source project as large as Material I try always to minimize breakage across versions, even minor versions. I try to keep them to major releases and so far there have been two.

I agree. However, it is important to remember that you are currently developing against a pre-1.0 version of Mkdocs. At this point I'm more interesting in getting things "right" than backward compatibility. The next release of MkDocs (which will include the refactored search) will be 1.0, so from my perspective, if we are going to break anything, now is the time to do it. What we settle on now is what we're stuck with going forward.

> I don't want my users to bother with more setup than they currently have

That's understandable and a consideration of mine as well. I realize search is a special case, but it is also the first plugin using our new API so its the testing ground for the API. I want to get that API right and how plugins and themes interact is a much bigger factor than I originally anticipated.

> It would be perfect if themes can define which features they support and turn on/off default flags for plugins, if they are available. Otherwise, if the user specifies the plugin themselves, I would expect that the user's choices take precedence.

Thank you. That's helpful feedback. And that more-or-less is what I am proposing. Unfortunately, I expect that means backward incompatible changes with the 1.0 release of MkDocs but I havn't fleshed out the details yet.

waylan:
@yeraydiazdiaz we have a couple [jshint errors][1] for your code that I'm not clear how to resolve. Any assistance would be appreciated. As a warning, I rebased the code since your last commit, so you may not be able to merge cleanly.

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/347345868#L505

yeraydiazdiaz:
Should be fixed via https://github.com/yeraydiazdiaz/mkdocs/commit/8597619665bd33352fe5a8a2cc9badc7df629969

mkdocs/contrib/search/prebuild-index.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  var idx = lunr(function () {
+    if (lang.length === 1 && lang[0] !== "en" && lunr[lang[0]]) {
+      this.use(lunr[lang[0]]);
+    } else if (lang.length > 1) {
+      this.use(lunr.multiLanguage.apply(null, lang));
+    }
+    this.field('title', { boost: 10 });
```
olivernn: As has been mentioned in https://github.com/mkdocs/mkdocs/issues/1319 and https://github.com/olivernn/lunr.js/issues/312 field boosts are not currently supported by Lunr 2.x. The second parameter here will be ignored, I'm not sure that does any harm, but perhaps a comment might help.

When I get around to implementing field boosts it will almost certainly use this interface (same as Lunr 1.x) so you _could_ leave this here to take advantage of field boosts when they arrive.

mkdocs/contrib/search/templates/search/lunr.js
```diff
@@ -0,0 +1,2977 @@
+/**
+ * lunr - http://lunrjs.com - A bit like Solr, but much smaller and not as bright - 2.1.5
```
olivernn: FYI Lunr [2.1.6](https://github.com/olivernn/lunr.js/blob/master/CHANGELOG.md#216) was released which includes a performance improvement when building indexes. I don't know how you intend on keeping up with releases of Lunr, but thought you might appreciate faster building!

waylan:
Thinking about this some more, I think maybe we should change the default for `prebuild_index` to `False`. The index building script is written in JavaScript and runs in Node.js. If the `node` command is not available on the system `PATH`, then it fails silently and no prebuilt index is generated (we get the same behavior as `prebuild_index: False`).

It occurs to me that this could present a problem in that a user could build the same project from two different machines and get two different results if one has node installed and the other doesn't. The way things are now, there would be no warning to the user. Instead, to prebuild the index, we should require the user to explicitly request it and then issue either a warning or error if it fails. I'm thinking perhaps a warning (which would still fail in "strict mode") as things will still "work" without it. 

While it's true that we loose the prebuilt index as a default, it really only benefits users with large sites anyway. Those users are going to be willing to do the extra work to set the config setting and install Node.js. If, at some point in the future, we get a Python solution for pre-building the index, then we can reconsider the default at that time.

squidfunk:
> While it's true that we loose the prebuilt index as a default, it really only benefits users with large sites anyway. Those users are going to be willing to do the extra work to set the config setting and install Node.js.

That's a very valid point and great idea. Users having problems with the search will want to go the extra mile. You could add a section on performance improvements for search in the documentation and add the `prebuild_index` option there. All other users won't notice anything. I'm also not a fan of the "do it if available or leave it out" method (regarding Node availability). The proposed behavior is far more transparent. Big up!

mkdocs/contrib/search/templates/search/lunr.js
waylan: Thanks for the heads up. Might as well release with the most recent version.

mkdocs/contrib/search/prebuild-index.js
waylan: Thanks. I realize boost is being ignored. But your right, no point in leaving it there. It will just cause confusion later when we forget that.

waylan:
The `prebuild_index` default has been changed. This feels much better. Baring any other concerns, I'll squash and merge this in a few days.

yeraydiazdiaz:
> If, at some point in the future, we get a Python solution for pre-building the index, then we can reconsider the default at that time.

Well, now that you mention it, I just published [v0.1.1a1 of Lunr.py](https://pypi.org/project/lunr/) 😄 

I honestly thought it'd take me longer to port but @olivernn did an excellent job with the JS version, especially with the unit tests which helped quite a bit to get the building blocks sorted quickly.

There's still a few places in the API I'd like to tweak to make it more Pythonic, but I'm happy to say it produces the same results searches on the MkDocs `search_index.json` that the JS version, both in Python 2.7 and 3. I'll make some tests to ensure serialized indexes can be passed back and forth between versions as a starting point for pre-building the index in pure Python.

Have a look at [the repo](https://github.com/yeraydiazdiaz/lunr.py) if you like, of course any feedback is appreciated.

josuahunziker:
@waylan Many thanks for your efforts here - I hope this makes the search also usable for my rather large documentation without blocking the browser. Do you already know when this is going to be released officially?

waylan:
@josuahunziker this will be part of the 1.0 release. Version 1.0 will be released when the [milestone][1] is complete.

[1]: https://github.com/mkdocs/mkdocs/milestone/4

josuahunziker:
@waylan Ok, many thanks! I'm looking forward! :smiley:

PR #1423, opened 2018-03-02
Use .svg badges in README instead of .png

benjamingorman:
The png versions of the badges can appear blurry on certain screen sizes. For me they look like this:

image

This PR uses the svg versions instead, which fixes that issue. It also uses the rounded version of the landscape.io 'health' badge to match the others.

image


waylan:
Do we care that svg support is not consistent across all browsers? I realize it’s mostly older browsers that are a problem. Personally I’m meh on this change.

benjamingorman:
https://caniuse.com/#feat=svg

Support for basic svg is nearly 98%. Should be fine?

PR #1426, opened 2018-03-06
Support Unicode characters in file paths

Friday21:
when I have unicode directory name under docs,  such as '导航', when I change the content under the direcory, and the mkdoc server auto detect the change and make build, it will throw UnicodeDecodeError:
```
[E 180306 16:13:35 ioloop:638] Exception in callback <bound method type.poll_tasks of <class 'livereload.handlers.LiveReloadHandler'>>
      File "/usr/lib64/python2.7/site-packages/tornado/ioloop.py", line 1026, in _run
        return self.callback()
      File "/usr/lib/python2.7/site-packages/livereload/handlers.py", line 67, in poll_tasks
        filepath, delay = cls.watcher.examine()
      File "/usr/lib/python2.7/site-packages/livereload/watcher.py", line 73, in examine
        func and func()
      File "/usr/lib/python2.7/site-packages/mkdocs/commands/serve.py", line 112, in builder
        build(config, live_server=live_server, dirty=dirty)
      File "/usr/lib/python2.7/site-packages/mkdocs/commands/build.py", line 259, in build
        utils.clean_directory(config['site_dir'])
      File "/usr/lib/python2.7/site-packages/mkdocs/utils/__init__.py", line 138, in clean_directory
        if entry.startswith('.'):
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 in position 0: ordinal not in range(128)

```

### DIFF
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -321,3 +321,30 @@ class UtilsTests(unittest.TestCase):
                     os.chmod(src, stat.S_IRUSR | stat.S_IWUSR)
             shutil.rmtree(src_dir)
             shutil.rmtree(dst_dir)
+
+    def test_unicode_clean_directory(self):
+        temp_dir = tempfile.mkdtemp()
+        utf8_directory = os.path.join(temp_dir, '导航')
+        utf8_utf8_directory = os.path.join(utf8_directory, '导航')
+        ascii_directory = os.path.join(temp_dir, 'bar')
+        ascii_ascii_directory = os.path.join(ascii_directory, 'bar')
+
+        os.makedirs(utf8_utf8_directory)
+        os.makedirs(ascii_ascii_directory)
+
+        utf_8_path = os.path.join(utf8_utf8_directory, 'foo.txt')
+        ascii_path = os.path.join(ascii_ascii_directory, 'foo.txt')
+
+        with open(utf_8_path, 'w') as f:
+            f.write('content')
+
+        with open(ascii_path, 'w') as f:
+            f.write('content')
+
+        try:
+            utils.clean_directory(utf8_directory)
+            utils.clean_directory(ascii_directory)
+            self.assertTrue(os.path.exists(utf_8_path), False)
+            self.assertTrue(os.path.exists(ascii_path), False)
+        finally:
+            shutil.rmtree(temp_dir)
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -344,7 +344,9 @@ class UtilsTests(unittest.TestCase):
         try:
             utils.clean_directory(utf8_directory)
             utils.clean_directory(ascii_directory)
-            self.assertTrue(os.path.exists(utf_8_path), False)
-            self.assertTrue(os.path.exists(ascii_path), False)
+            self.assertFalse(os.path.exists(utf_8_path))
+            self.assertFalse(os.path.exists(ascii_path))
+            self.assertFalse(os.path.exists(utf8_utf8_directory))
+            self.assertFalse(os.path.exists(ascii_ascii_directory))
         finally:
             shutil.rmtree(temp_dir)
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -12,6 +12,7 @@ import stat
 
 from mkdocs import nav, utils, exceptions
 from mkdocs.tests.base import dedent, load_config
+from mkdocs.utils.ghp_import import dec
 
 
 class UtilsTests(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


waylan:
Thanks for the patch. However, we should include a test with this. Also, I wonder if we should actually decode `entry` and pass it on rather than only decoding within the conditional test for "starts with a dot".

mkdocs/tests/utils/utils_tests.py
```diff
@@ -321,3 +321,30 @@ def test_copy_files_without_permissions(self):
...
+            f.write('content')
+
+        with open(ascii_path, 'w') as f:
+            f.write('content')
+
+        try:
+            utils.clean_directory(utf8_directory)
+            utils.clean_directory(ascii_directory)
+            self.assertTrue(os.path.exists(utf_8_path), False)
+            self.assertTrue(os.path.exists(ascii_path), False)
```
waylan: Shouldn't this be:

    self.assertFalse(os.path.exists(utf_8_path))
    self.assertFalse(os.path.exists(ascii_path))

Might also want to add:

    self.assertFalse(os.path.exists(utf8_utf8_directory))
    self.assertFalse(os.path.exists(ascii_ascii_directory))

Friday21: yeah, you're right, I use assertTrue wrong, I use it as assertEqual.

Friday21:
Wonder why this failed in appveypr,  in Python2.7 it passed both on my local and travis
```
42  File "C:\projects\mkdocs\mkdocs\tests\utils\utils_tests.py", line 354, in test_unicode_clean_directory
43    shutil.rmtree(temp_dir)
44  File "c:\python27\Lib\shutil.py", line 266, in rmtree
45    onerror(os.remove, fullname, sys.exc_info())
46  File "c:\python27\Lib\shutil.py", line 264, in rmtree
47    os.remove(fullname)
48WindowsError: [Error 123] The filename, directory name, or volume label syntax is incorrect: 'c:\\users\\appveyor\\appdata\\local\\temp\\1\\tmpry7a63\\??'
```

mkdocs/tests/utils/utils_tests.py
```diff
@@ -321,3 +321,32 @@ def test_copy_files_without_permissions(self):
                     os.chmod(src, stat.S_IRUSR | stat.S_IWUSR)
             shutil.rmtree(src_dir)
             shutil.rmtree(dst_dir)
+
+    def test_unicode_clean_directory(self):
+        temp_dir = tempfile.mkdtemp()
+        utf8_directory = os.path.join(temp_dir, '导航'.decode('utf-8'))
```
waylan: This isn't necessary. Up at the top of the file we have `from __future__ import unicode_literals` which ensures that all literal strings defined in in the module are Unicode strings. Its basically the same as the `u` prefix: `u'unicode string'`.

What we need t check is that `temp_dir` is Unicode. Or any other path which is not created by us.

Friday21: Thanks, now I figure out that, when user os.path.join(arg1, arg2), if both arg1 and arg2 is str or unicode, it will work, but if arg1 and arg2 are mixed of str and unicode, python will encode the str arg with system defaultencoding, which on windows is not utf-8, and it throws exceptions.   
as tempfile didn't work with utf-8 well , so I make utf-8 dir myself:  
```
temp_dir = tempfile.mkdtemp()
utf8_temp_dir = os.path.join(dec('tmp'), dec('tmp_utf8'))
```

waylan:
Another solution is to use a Path object library, such as [path.py][1] or [pathlib][2]. By wrapping all paths in the Path object, we get support for all path related operations out-of-the-box. It would allow us to remove much of our current utils methods. Of course, it would require a later refactor. Perhaps this would make sense as part of the [Pages Refactor][3]

Here are some differences between the two libs, in no particular order:

* path.py objects are `str` (`unicode` in PY2) subclasses and work with standard string operations. pathlib objects require conversion to a string first (`str(path)`).
* path.py is a third party library with PY2.7+ support. pathlib is in the standard library for PY3.4+ with some newer features only available in PY3.5 or PY3.6. However, [pathlib2][4] is an up-to-date backport with PY2.7+ support. 
* path.py includes features wrapping shlib, etc. for a fuller feature set.
* pathlib has different object types for Windows vs. Posix paths. It also has pure path types (no relation to actual files) with limited functionality and concrete path types (connected to actual files) with additional functionality. path.py only has one object type which offers all functionality. Both libraries just-work on supported systems.

[1]: https://github.com/jaraco/path.py
[2]: https://docs.python.org/3/library/pathlib.html
[3]: https://github.com/mkdocs/mkdocs/projects/2
[4]: https://github.com/mcmtroffaes/pathlib2

waylan:
Closing this in favor of #1517. This was treating the symptoms, while the changes in #1517 and 92995d4 of #1504 treat the problem.

PR #1427, opened 2018-03-07
Tornado 5.0 raises error on install with older Python versions.

wiesendaniel:
changed to `"tornado>=4.1,<5.0"` in setup.py

This broke installation via pip for me.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
         'livereload>=2.5.1',
         'Markdown>=2.3.1',
         'PyYAML>=3.10',
-        'tornado>=4.1',
+        'tornado>=4.1,<5.0',
     ],
     entry_points={
         'console_scripts': [


waylan:
As per the [release notes][1], Tornado 5.0 was released yesterday (2018/03/05). Some noteworthy points include:

* Python 3.3 is no longer supported.
* Versions of Python 2.7 that predate the ssl module update are no longer supported. (The ssl module was updated in version 2.7.9, although in some distributions the updates are present in builds with a lower version number.)
* Versions of Python 3.5 prior to 3.5.2 are no longer supported due to a change in the async iterator protocol in that version.

Without an error report, I'm not sure if the reported problem is due to one of those issues or something else. If it is one of those issues, then you should be updating your Python version, not staying with an older Tornado version.

If, on the other hand, the error is related to some other problem, then our code should be updated to work with the new way forward. I'm not accepting this PR as-is, but I'll leave it open as a bug report. We need to investigate what specifically is causing the problem before determining the proper approach.

[1]: http://www.tornadoweb.org/en/stable/releases/v5.0.0.html

waylan:
FYI, I just copied the above fix ("changed to `tornado>=4.1,<5.0`") to the `0.17` branch in #1428 so we can do an immediate release (0.17.3). That way, new users won't immediately run into an error. Of course, in master, we'll fix the issue properly for the 1.0 release.

waylan:
MkDocs version `0.17.3` is now available.

waylan:
On one system with an older "insecure platform" (Python 2.7.1), I get the following when trying to install Tornado 5.0:

```
$ pip install -U tornado
Collecting tornado
  Downloading tornado-5.0.tar.gz (504kB)
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "c:\users\wlimberg\appdata\local\temp\pip-build-qucodk\tornado\setup.py", line 146, in <module>
        raise ImportError("Tornado requires an up-to-date SSL module. This means "
    ImportError: Tornado requires an up-to-date SSL module. This means Python 2.7.9+ or 3.4+ (although some distributions have backported the necessary changes to older versions).
```

That error makes it pretty clear that the user needs to update to a "secure" version of Python. I'm okay with that. Will need to test on other platforms later when I have time/access.

Btw, in the next major release (version 1.0) of MkDocs, I expect we will be dropping support for any Python versions which Tornado doesn't support (we might even change the dependency to `tornado>=5.0`). So, I'm okay with an install of future MkDocs versions failing on those Python versions. I'm more interested in if any errors occur on a supported Python version.

waylan:
I just tested on a system which has Python 2.7.13 and 3.6.1 installed and I got no errors. I was able to install Tornado 5.0 and I was able to run `mkdocs serve` without error (MkDocs version 0.17.2). I also tried the `--no-livereload` option and that worked fine as well. It seems pretty clear that any errors are simply related to older python versions which are not supported by Tornado 5.0.

That being the case, MkDocs 1.0 will drop support for any Python version not supported by Tornado 5.0.

wiesendaniel:
Sorry for taking so long to get back.
My report should have bin more detailed i guess.
My error wasn't related to an old python version as i was using a fresh copy of the python:3 docker image. Should have saved the log, tho.
Now it seems fixed even with 0.17.2.
For reference my Dockerfile:
```
FROM python:3-slim-stretch

RUN apt-get update && apt-get install -y --no-install-recommends \
        git \
    && rm -rf /var/lib/apt/lists/* \
    && pip install mkdocs

ENTRYPOINT ["/usr/local/bin/mkdocs"]
```

Thank you really much for you quick response and actions taken.

pquentin:
The problem with requiring tornado < 5.0 is that if another dependency requires tornado and is installed first, then it's going to install tornado 5.x anyways since pip does not resolve dependencies. And at this point, even if you don't care about tornado, you cannot build the documentation any longer!

```
mkdocs build --clean --site-dir=build/docs/
  File "/venv/bin/mkdocs", line 11, in <module>
    sys.exit(cli())
  File "/venv/lib/python3.5/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/venv/lib/python3.5/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/venv/lib/python3.5/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/venv/lib/python3.5/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/venv/lib/python3.5/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/venv/lib/python3.5/site-packages/mkdocs/__main__.py", line 155, in build_command
    site_dir=site_dir
  File "/venv/lib/python3.5/site-packages/mkdocs/config/base.py", line 181, in load_config
    errors, warnings = cfg.validate()
  File "/venv/lib/python3.5/site-packages/mkdocs/config/base.py", line 97, in validate
    run_failed, run_warnings = self._validate()
  File "/venv/lib/python3.5/site-packages/mkdocs/config/base.py", line 52, in _validate
    self[key] = config_option.validate(value)
  File "/venv/lib/python3.5/site-packages/mkdocs/config/config_options.py", line 131, in validate
  File "/venv/lib/python3.5/site-packages/mkdocs/config/config_options.py", line 638, in run_validation
    plgins[item] = self.load_plugin(item, {})
  File "/venv/lib/python3.5/site-packages/mkdocs/config/config_options.py", line 647, in load_plugin
    Plugin = self.installed_plugins[name].load()
  File "/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2290, in load
    self.require(*args, **kwargs)
  File "/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2307, in require
    items = working_set.resolve(reqs, env, installer)
  File "/venv/lib/python3.5/site-packages/pkg_resources/__init__.py", line 858, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.VersionConflict: (tornado 5.0 (/venv/lib/python3.5/site-packages), Requirement.parse('tornado<5.0,>=4.1'))
abort: command failed with mkdocs build --clean --site-dir=build/docs/: 1
```

Is it possible to fix `mkdocs build` with tornado 5.x? Thanks for mkdocs.

waylan:
@pquentin the problem is twofold:

1. I have not been able to reproduce any errors with Tornado 5.0 and no one has provided any usable error reports. Therefore, there is no known issue to fix.
2. Those using Python versions not compatible with Tornado 5.0 need the `<5.0` requirement. And this is not "fixable" in MkDocs but is a limitation of Tornado itself.

Currently the only version of MkDocs which contains the `<5.0` dependency requirement is `0.17.3` and the only difference between `0.17.3` and 0.17.2` is that dependency requirement (no code changes were made). In your situation I would suggest using MkDocs `0.17.2`.

To be clear, as stated multiple times above, the next major release of MkDocs will expect Tornado 5.0+ (although it should still work fine with older versions).

pquentin:
@waylan Thanks, that makes sense! I appreciated the calm explanation. I now realize the tone of my message was not appropriate: please accept my apologies.

Issue #1378, opened 2018-01-16
Improve edit_uri flexability


jkobti:
Hi guys, 
According to the documentation, I should be able to only add the link to my bitbucket repository to get edit links. 
I added it like this: repo_url: https://bitbucket.org/josephkobti/testdocs/
but the links I'm getting are not working
i.e. https://bitbucket.org/josephkobti/testdocs/src/default/docs/index.md

I think maybe I should add the repo_url in another way, any suggestions?

waylan:
What URL are you expecting to get?

IIRC, Bitbucket doesn't have an "edit" URL for documents in your repo. We can only point to the document, then the user needs to click on the "edit" button which uses a script on Bitbucket's site to switch to edit mode (there is no change to the URL). In other words, this is a limit of Bitbucket which we have no control over.

If, on the other hand, Bitbucket does offer an "edit" URL for documents, we would be happy to update MkDocs to support it. Please let us know what that URL scheme is and we'll add support.

waylan:
Closing this due to a lack of information to act on.

elvismdev:
@jkobti In my case I set `edit_uri` value to `src/master/docs` and that should point the users directly to the documentation file inside the Bitbucket repository source code. Therefore they can click on "edit" from the Bitbucket UI to make changes to the documentation .md file.

@waylan If that helps, this is how an edit URL in Bitbucket would open directly the file from the source code editor: https://bitbucket.org/elvismdev/my-project/src/master/docs/index.md?mode=edit&spa=0

waylan:
@elvismdev thank you for providing Bitbiucket's edit URL format.

It would seem that MkDocs' current implementation is inadequate for general use. As it currently stands, an "edit URL" is composed by joining the following three parts:

    { repo_url }/{ edit_url }/{ page_url }

However, Bitbucket would need:

    { repo_url }/src/{ branch }/{ page_url }?mode=edit&spa=0

And based on this recent report in #1437, a GitHub wiki would need:

    { repo_url }/wiki/{ page_url }/_edit

Given the above, it seems that rather than having a setting for `edit_uri` it would be more effective to have a setting for a `edit_url_template` which the user could define. Then MkDocs could use Python's `string.format()` to pass in a the page path. I suppose it could accept the `repo_url` as well, but that shouldn't be necessary as the user could just include that in the `edit_url_template`.

waylan:
[...]

PR #1435, opened 2018-03-14
Add support for GitLab repositories

tribut:
This adds support for GitLab repositories in the same way that already exists for GitHub and Bitbucket.

Note that the icons in the themes will not show up until FontAwesome is updated to at least version 4.6 and/or an up-to-date version of the upstream readthedocs css is imported.

This commit also fixes a slight documentation error regarding the default value of `repo_name` and removes some tautological tests.

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -113,7 +113,9 @@ edit_uri: root/path/docs/
     (`blob/master/docs/`) to open a read-only view, which supports anonymous
     access.
 
-**default**: `edit/master/docs/` for GitHub and GitLab repos or `src/default/docs/` for a Bitbucket repo, if `repo_url` matches those domains, otherwise `null`
+**default**: `edit/master/docs/` for GitHub and GitLab repos or
+`src/default/docs/` for a Bitbucket repo, if `repo_url` matches those domains,
+otherwise `null`
 
 ### site_description
 
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -35,7 +35,8 @@ URL to the generated HTML header.
 
 ### repo_url
 
-When set, provides a link to your GitHub or Bitbucket repository on each page.
+When set, provides a link to your repository (GitHub, Bitbucket, GitLab, ...)
+on each page.
 
 ```yaml
 repo_url: https://github.com/example/repository/
@@ -45,10 +46,10 @@ repo_url: https://github.com/example/repository/
 
 ### repo_name
 
-When set, provides a link to your GitHub or Bitbucket repository on each page.
+When set, provides the name for the link to your repository on each page.
 
-**default**: `'GitHub'` or `'Bitbucket'` if the `repo_url` matches those
-domains, otherwise `null`
+**default**: `'GitHub'`, `'Bitbucket'` or `'GitLab'` if the `repo_url` matches
+those domains, otherwise the hostname from the `repo_url`.
 
 ### edit_uri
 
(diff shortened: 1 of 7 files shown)


docs/user-guide/configuration.md
```diff
@@ -105,15 +106,16 @@ edit_uri: root/path/docs/
...
+    the page in the online editor. This functionality requires that the user
+    have and be logged in to a GitHub/GitLab account. Otherwise, the user will
+    be redirected to a login/signup page. Alternatively, use the "blob" path
     (`blob/master/docs/`) to open a read-only view, which supports anonymous
     access.
 
-**default**: `edit/master/docs/` or `src/default/docs/` for GitHub or Bitbucket
-repos, respectively, if `repo_url` matches those domains, otherwise `null`
+**default**: `edit/master/docs/` or `src/default/docs/` GitHub, GitLab or
+Bitbucket repos, respectively, if `repo_url` matches those domains,
```
waylan: Missing "for" before "GitHub". Although, "respectively" doesn't really work as there are two items in the first list and three in the second. Sure, users in the know will understand, but we should not assume users will know that GitHub and Gitlab share the same URL scheme. How about: 

    **default**: `edit/master/docs/` for GitHub and GitLab repos or `src/default/docs/` for a Bitbucket repo...

waylan:
In view of #1378 I'm wondering if we should even bother with this. Yes, this gives GitLab equal footing with GitHub, which is not a bad thing. However, every time a new service comes along, I don't want to have to add support for it like this. That would be a maintenance nightmare. Perhaps if we take the approach proposed in #1378, then we might be able to avoid the need for this.

What I would really like to do is to rip all of the repo specific stuff out and implement it as a plugin. That gives us the freedom to add all sorts of configuration without adding to the main config. And as new services come along, any third party could create a plugin which worked with that services' specific needs. My hesitation involves how to deal with themes. Plugins can't inject template code into an existing theme unless they override/replace an entire file. And getting a plugin to provide template code that works across all themes, including third party themes, would be nigh impossible. I'm thinking perhaps the solution would be to have the plugin provide a documented list of template variables and templates would need to provide the logic to display those variables (this is more-or-less the approach taken by the search plugin). For example, one of the variables could be the "name" of the icon to display. What names would be valid would depend on the theme you use. For example, a theme which uses FontAwesome might support different names than a theme that used some other mechanism for displaying icons. Consider the following template snippet which would work in the readthedocs theme:

```django
class="icon icon-{{ icon_name }}"
```

Thoughts?

tribut:
Hey - I've made the change you suggested. My wording was indeed a bit wonky.

As far as the larger issue is concerned it is certainly true that maintaining the list of potential services here and replicated in all the themes seems unsustainable. I've only opened this PR because I noticed the [icon was missing in cinder](https://github.com/chrissimpkins/cinder/pull/42) and that bugged me. Adding a variable for the service/icon name seems reasonable, and maybe a service name is enough (currently the icon name would only be `service_name | lower` and that already assumes everyone is using FontAwesome).

However, unless you are planning to implement this soon, I'd still suggest merging this PR. Ripping it out again later does no harm.

tribut:
One more thing I just realized: Currently I can set `repo_name: GitLab` and the icon will be picked up. This works, even if I am using a privately hosted GitLab instance. It would be great if this flexibility would be preserved in a future, plugin-based approach.

waylan:
As this is a fully working implementation for GitLab we'll merge this now. That way, if/when we address the other issues, we'll be sure to include GitLab. Otherwise it may get forgotten about.

@tribut thanks again for doing the work on this.

PR #1446, opened 2018-03-27
Update version number in --version example

coliff:


waylan:
Thanks for your submission. However, I have been intentionally holding off on doing this as I don't want to have to update this every time we do a release.

Additionally, the next release from the `master` branch will be version 1.0. And we only ever update the site on a release. Therefore, it would make more sense to show `1.0.0` as the version number that is output.

coliff:
ok no problem. I'll close this -really looking forward to v1.0!

PR #1447, opened 2018-03-28
Update Font Awesome to 4.7.0 (MkDocs Theme)

### LINKED ISSUE #1438: Update FontAwesome in themes
FontAwesome should be updated to latest for the 1.0 release. 

See #1435 for at least one icon (GitLab) missing from the currently included version.

coliff:
Fixes #1438

Note: I'm looking into updating Font Awesome on the ReadTheDocs theme (the upstream RTD theme already has Font Awesome 4.7.0 included - but the CSS for the theme of the site is bundled together and is different to the one in MkDocs now so more work may be required to make sure the themes are in sync)


waylan:
Updating the readthedocs theme with upstream is the subject of #588, which no one seems to want to work on. I see two paths. FontAwesome could be updated along with #588 or we could update FontAwesome only for the readthedocs theme and leave #588 for later. Either way, I don't think it makes sense to update FontAwesome on one theme but not the other.

coliff:
> I don't see the prefixes updated in mkdocs/themes/mkdocs/nav.html#L78-L82. Shouldn't they be updated from fa to fab?
I upgraded Font Awesome on the MkDocs theme to v4.7.0 which is the same as what the current readthedocs theme uses. No prefix changes are needed at all.. it just means we get access to many more icons. (https://fontawesome.com/v4.7.0/icons/)

I agree the RTD theme needs updating too, but looks like the whole theme needs updating so maybe that theme should be updated in a separate PR.

There are more changes which should ideally be made to the MkDocs theme too - it's using a very old version of Bootstrap which doesn't support jQuery 3. I'd like to make a seperate PR for that updating all the assets (Bootstrap, jQuery). Would actually be nice to update to Bootstrap 4...

waylan:
> I upgraded Font Awesome on the MkDocs theme to v4.7.0

Right. I noticed that and then promptly forgot when I was doing my review. My apologies. Although, why not upgrade to v5? If so , I think just using the [CSS solution][1] is fine. No need for the [JS/SVG solution][2], which apparently has additional [complexities when used with JQuery][3]. 

[1]: https://fontawesome.com/how-to-use/web-fonts-with-css
[2]: https://fontawesome.com/how-to-use/svg-with-js
[3]: https://fontawesome.com/how-to-use/svg-with-js#with-jquery
[4]: https://fontawesome.com/how-to-use/server-side-rendering

I couldn't help but notice that FontAwesome supports [prerendering][4] the JS/SVG solution on Node based sites. It would be interesting to get similar behavior here. But that would require cloning the JS code in Python and should probably be implemented as some sort of Jinja tag which is developed as a third party general purpose solution... So, probably not practical to do at this stage.

> Would actually be nice to update to Bootstrap 4.

IIRC, there is a very specific reason why we don't use Bootstrap 4. Seems that there is some feature that Bootstrap 4 removed support for that we use/need. Unfortunately I couldn't find it in a search. Perhaps I'm thinking of something else.

coliff:
Hi @waylan - I'd been working on updating this PR to Font Awesome 5 as you suggested.
I'm nearly done and expect to update my PR for that tomorrow.

coliff:
Adding Font Awesome 5 fonts would mean adding a new `webfonts` dir and hosting 15 fonts (brands regular and solid - in 5x file formats). We could move them to the `fonts` dir where the Bootstrap 3 fonts are, but that would mean manually updating the paths of the CSS files.
So for now I updated to Font Awesome 4.7.0.

PR #1451, opened 2018-03-29
Add .editorconfig

### LINKED ISSUE #1449: Editorconfig
It'd be good to add a `.editorconfig` file so that anyone contributing to MkDocs can follow the same rules for tabs/spaces, newlines, charsets etc.
Editorconfig (http://editorconfig.org/) is very popular and it's become standard for open-source projects to include one.
I would imagine it should be something like this:
```
# editorconfig.org

root = true

[*]
charset = utf-8
end_of_line = lf
indent_size = 4
indent_style = space
insert_final_newline = true
trim_trailing_whitespace = true

[*.md]
trim_trailing_whitespace = false
```
Shall I open a PR?

coliff:
Fixes #1449

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,8 +4,7 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = lf
-indent_size = 4
+indent_size = 2
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,10 +4,13 @@ root = true
 
 [*]
 charset = utf-8
-indent_size = 2
+indent_size = 4
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.js]
+indent_size = 2
 
 [*.md]
 trim_trailing_whitespace = false


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
```
waylan: How does `end_of_line` interact with Git for Windows? IIRC, the default behavior is for git to convert all line endings to windows style (`rflf`) on fetch/pull and to convert back on push. Setting the editor to use `lf` line endings is likely to mess with this.

Perhaps the Editorconfig people have worked this out. I don't know, I haven't checked. But I'm not comfortable with this until that is confirmed.

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_size = 4
```
waylan: I believe our JS files use `indent_size = 2`. Using `4` as the default makes sense, but we might want to add a `*.js` override. Or perhaps we should consider changing to use `4` everywhere. We really need to "contributing guide" which defined these things. Then this file could just match whatever is defined there.

.editorconfig
coliff: I'm not too sure on that - if we're unsure we can simply remove that line.

.editorconfig
coliff: I'm fine with either 2 or 4 really...  2 seems slightly more common.
`mkdocs\mkdocs\themes\mkdocs\js\base.js` uses 4 so I set it as that but we could change to 4.

.editorconfig
coliff: A few of the biggest and related projects' editorconfig files for comparison:
https://github.com/squidfunk/mkdocs-material/blob/master/.editorconfig
https://github.com/h5bp/html5-boilerplate/blob/master/.editorconfig
https://github.com/twbs/bootstrap/blob/v4-dev/.editorconfig
https://github.com/Microsoft/vscode/blob/master/.editorconfig

.editorconfig
waylan: All the work in the recent search refactor uses 2 for JS files. We have a lot more JS files that use 2 than 4, so let's use 2.

coliff: Cool. updated the PR to 2.

.editorconfig
coliff: Removed `end_of_line = lf`

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+indent_size = 2
```
waylan: Sorry if I wasn't clear. I still want the default to be 4. We just need to have 2 for `*.js` files only.

coliff: no problem. PR updated.

PR #1453, opened 2018-04-03
Drop official support for Python 3.3.

waylan:
Also up tornado version to >=5.0. Fixes #1427.


waylan:
Note that other than dependencies (namely tornado) dropping support for PY3.3 and some CI servers (AppVeyor runs the py33 tests on py27) no longer supporting PY3.3, there should be no reason MkDocs won't run on PY3.3.  However, given that Python has officially deprecated PY3.3, then it makes sense for us to as well. Especially given that our next release will be 1.0.

Issue #1452, opened 2018-04-02
Clarify documentation


charlesreid1:
Hello,

First, mkdocs is a great tool and I've been really enjoying it, so thanks to the developers.

The [Customizing a Theme](http://www.mkdocs.org/user-guide/styling-your-docs/#customizing-a-theme) section of the documentation could use some clarification. I have spent about an hour tinkering with it, but still can't understand how it works.

* If I want to add an external CSS file, docs tell me I can use the `docs_dir`. Okay.
* Docs give me a simple example CSS file that I can drop into `docs_dir/extras.css`. Okay.
* Note in the docs says, if I'm deploying with readthedocs (which I'm not), I need to include the CSS with the `extra_css: [extra.css]` directive explicitly in my YAML file. I'm not using readthedocs, so I don't.

When I serve the documentation with `mkdocs serve`, the CSS file is not picked up. The CSS file is not included in the HTML at all. So, **the blue note mentioning readthedocs is incorrect or not clear enough**.

The documentation picks up after that: "After making these changes, they should be visible when you run mkdocs serve" - but **it's unclear what changes the documentation is referring to**. The `extra_css` change mentioned in the note? Or creating the `extras.css` CSS file?

I ended up having to add the `extra_css: [extra.css]` bit into my YAML configuration file to get it to add `extra.css` to my index, so the blue note mentioning readthedocs apparently applies to **all** cases.

However, even then it took more tinkering, because `extra_css: [extra.css]` was adding a link to `extra.css` in the current directory, not the `css/` directory where it was actually located.

I could keep tinkering, e.g., by changing `extra_css: [css/extra.css]` (which I think is valid), but that leads to another problem, which is that `extra.css` is not being copied into the `css/` directory when I create the site.

Without better documentation, I'm stuck with banging my head against a wall... or just picking a different theme.

I've been really happy with mkdocs, all the other features are pretty straightfoward. It's just this feature I'm having trouble with, mainly because of vague/confusing documentation.

waylan:
Thanks for pointing this out. As the [release notes][1] mention, the behavior of the `extra_css` setting has recently changed. It appears that we updated the [configuration][2] docs, but not the Theme Customization docs.

> I could keep tinkering, e.g., by changing extra_css: [`css/extra.css`] (which I think is valid), but that leads to another problem, which is that `extra.css` is not being copied into the `css/` directory when I create the site.

[...]

PR #1463, opened 2018-04-08
Improving documentation

charlesreid1:
Clarify the Customizing a Theme section describing "extras.css"

This makes the following changes:
- explain the default value of `docs_dir` to clarify (see #1452)
- put custom CSS into `custom.css` and not `extra.css` to reduce confusion from `extra_css` vs `extra.css` vs `extra/css` etc.
- add the actual YAML config block that is needed to add extra css

Also mention that custom themes require a configuration file. (TODO: Need clear information about the configuration file that's required. Leave that for another PR.)

### DIFF
--- a/docs/user-guide/custom-themes.md
+++ b/docs/user-guide/custom-themes.md
@@ -19,9 +19,10 @@ and their usage.
 
 ## Creating a custom theme
 
-The bare minimum required for a custom theme is a [theme configuration file](#theme-configuration) 
+The bare minimum required for a custom theme is a 
+[theme configuration file](#theme-configuration) 
 and a `main.html` [Jinja2 template]
-file which is placed in a directory that is *not* a child of the [docs_dir].
+file, both placed in a directory that is *not* a child of the [docs_dir].
 Within `mkdocs.yml`, set the theme.[custom_dir] option to the path of the
 directory containing `main.html`. The path should be relative to the
 configuration file. For example, given this example project layout:
--- a/docs/user-guide/styling-your-docs.md
+++ b/docs/user-guide/styling-your-docs.md
@@ -58,7 +58,7 @@ customizations, including overriding templates, you will need to use the theme
 
 ### Using the docs_dir
 
-The [extra_css] and [extra_javascript] configuration options can be added to 
+The [extra_css] and [extra_javascript] configuration options can be added to
 `mkdocs.yml` to add additional CSS and Javascript files to existing themes.
 Include CSS or Javascript files within `docs_dir` (`docs/` by default) and
 itemize them under the `extra_css` heading in `mkdocs.yml`.
@@ -75,7 +75,7 @@ h1 {
 
 Then add the following to your `mkdocs.yml`:
 
-```
+```yaml
 extra_css:
   - custom.css
 ```
@@ -89,9 +89,9 @@ extra_css:
         extra_css: [extra.css]
 
 After addding these files and making the changes to `mkdocs.yml`,
-the changes should be visible when you run `mkdocs serve` - 
-if you already had this running, you should see changes to the 
-YAML file be automatically picked up and the documentation should be 
+the changes should be visible when you run `mkdocs serve` -
+if you already had this running, you should see changes to the
+YAML file be automatically picked up and the documentation should be
 updated.
 
 !!! note


docs/user-guide/styling-your-docs.md
```diff
@@ -37,6 +39,8 @@ A clone of the default theme used by the [Read the Docs] service. This theme
 only supports features in its parent theme and does not support any MkDocs
 [theme configuration options] in addition to the defaults.
 
+The mkdocs theme is located in the mkdocs repository at `mkdocs/themes/readthedocs/`.
```
waylan: This should be: "The **readthedocs** theme..."

docs/user-guide/styling-your-docs.md
```diff
@@ -77,9 +88,11 @@ h1 {
 
         extra_css: [extra.css]
 
-After making these changes, they should be visible when you run
-`mkdocs serve` - if you already had this running, you should see that the CSS
-changes were automatically picked up and the documentation will be updated.
+After addding these files and making the changes to `mkdocs.yml`,
+the changes hsould be visible when you run `mkdocs serve` -
```
waylan: hsould => should

waylan:
A few notes:

A theme configuration file is not needed in a `custom_dir`. They are only necessary in an "installed" theme. If it helps, "installed" themes have a name, whereas the `custom_dir` has no name (as the name is set to a null value). Any configuration happens in the project's `mkdocs.yml` config file.

Second, for future reference, it was not necessary to create all those new issues. They could have all been introduced with this PR.

docs/user-guide/styling-your-docs.md
charlesreid1: Oops, good catch!

docs/user-guide/styling-your-docs.md
```diff
@@ -77,9 +88,11 @@ h1 {
 
         extra_css: [extra.css]
 
-After making these changes, they should be visible when you run
-`mkdocs serve` - if you already had this running, you should see that the CSS
-changes were automatically picked up and the documentation will be updated.
+After addding these files and making the changes to `mkdocs.yml`,
```
waylan: This change is completely unnecessary. Part of the "changes" includes adding the files. Why do we need to mention that specifically?

docs/user-guide/styling-your-docs.md
```diff
@@ -54,21 +58,28 @@ customizations, including overriding templates, you will need to use the theme
 
 ### Using the docs_dir
 
-The [extra_css] and [extra_javascript] configuration options can be used to
-make tweaks and customizations to existing themes. To use these, you simply
-need to include either CSS or JavaScript files within your [documentation
-directory].
+The [extra_css] and [extra_javascript] configuration options can be added to
+`mkdocs.yml` to add additional CSS and Javascript files to existing themes.
+Include CSS or Javascript files within `docs_dir` (`docs/` by default) and
```
waylan: Why are we removing the link here? Wouldn't it be better to leave the link to `docs_dir` rather than add that the default is `docs/`. 

A lot of your confusion seems to stem from contradictory documentation which was introduced when changes were made. By keeping things documented in one location (and linking to is from elsewhere) we can avoid that in the future.

A good rule of thumb I have been trying to follow is whenever a new setting is introduced within a section, link to that setting on the configuration page. That way, users can easily discover what that setting is for, what valid values it accepts and what it defaults to.

True, the docs don't do that everywhere, but every section I'v recently edited does. Slowly the docs have been getting better in that regard.

docs/user-guide/custom-themes.md
```diff
@@ -19,8 +19,10 @@ and their usage.
 
 ## Creating a custom theme
 
-The bare minimum required for a custom theme is a `main.html` [Jinja2 template]
-file which is placed in a directory that is *not* a child of the [docs_dir].
+The bare minimum required for a custom theme is a 
+[theme configuration file](#theme-configuration) 
+and a `main.html` [Jinja2 template]
+file, both placed in a directory that is *not* a child of the [docs_dir].
```
waylan: The line length is 80 chars. I realize the long link messes things up a little, but this could look better. May I suggest using your editors tools to automatically hard wrap the lines.

charlesreid1:
mkdocs documentation sucks. I made my attempt to improve it. If getting a PR merged and improving the documentation requires arguing over every change I propose, I don't have the time. Thanks.

trel:
I haven't seen any argumentation.  Only suggests and rationale.

waylan:
@charlesreid1 I am sorry you feel that way. I meant no disrespect. Review and improve is a normal part of the development process. Unfortunately, that has not always happened in the past, which is a contributing factor to the documentation being somewhat splintered. If we had followed this process with every submission, perhaps we would have caught some of the existing errors in the documentation before they were ever introduced.

I see you closed this issue. However,  I am willing to take what you have and finish it when I have the time. Therefore, I'm reopening this. Thanks for the time you have put in. It is much appreciated.

PR #1467, opened 2018-04-12
Warn users deploying with older version.

### LINKED ISSUE #640: Warn users if they are using an old/different version

waylan:
Fixes #640.

This was a pretty straightforward change. but its not clear how this should be tested. Obviously, mocks can be used to simulate different return values (for the versions), but the behaviors we need to test are checking for various logging messages.

At this point, let's see if the current changes break any existing tests.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -66,6 +66,7 @@ authors should review how [search and themes] interact.
 
 ### Other Changes and Additions to Development Version
 
+* Add MkDocs version check to gh-deploy script (#640).
 * Improve Markdown extension error messages. (#782).
 * Drop official support for Python 3.3 and set `tornado>=5.0` (#1427).
 * Add support for GitLab edit links (#1435).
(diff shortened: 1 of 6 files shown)


waylan:
Apparently [unittest.TestCase.assertLogs][1] was added in Python 3.4. A PY2.7 backport can be found [here]. The [Unittest2][3] lib also contains a backport, however, I don;t see the need for that entire lib for one feature.

There is also the [testfixtures][4] lib, which includes some [logging capture tools][5], along with a bunch of other stuff that could simplify a bunch of our tests.

[1]: https://docs.python.org/3.4/library/unittest.html#unittest.TestCase.assertLogs
[2]: https://stackoverflow.com/a/29938398/866026
[3]: https://pypi.python.org/pypi/unittest2
[4]: http://testfixtures.readthedocs.io/en/latest/index.html
[5]: http://testfixtures.readthedocs.io/en/latest/logging.html

PR #1478, opened 2018-05-01
[WIP] extra vars in markdown

rosscdh:
Thanks for the great lib 
have a usecase that requires the "extra" vars to be made available in the markdown. This is a major feature not sure why its not by default?

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -199,6 +199,7 @@ def build_extra_templates(extra_templates, config, site_navigation=None):
             'template_context', context, template_name=extra_template, config=config
         )
 
+        print(context)
         output_content = template.render(context)
 
         # Run `post_template` plugin events.
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -7,6 +7,7 @@ This consists of building a set of interlinked page and header objects.
 """
 
 from __future__ import unicode_literals
+from jinja2 import Template
 import datetime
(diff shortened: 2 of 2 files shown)


waylan:
This would make a great candidate for a [plugin][1]. Personally I would never want this myself and am not interested in maintaining it.

[1]: http://www.mkdocs.org/user-guide/plugins/

rosscdh:
Thanks for the tip ill take a look

waylan:
You might want to check out [fmaida/hello-dolly-mkdocs-plugin][1] which is a very basic plugin and would contain all the boilerplate code for a feature like this.

[1]: https://github.com/fmaida/hello-dolly-mkdocs-plugin

rosscdh:
Thanks for the tip again, long time dev here figured it out just from the docs Heres one i created just then https://github.com/rosscdh/mkdocs-markdownextradata-plugin

rosscdh:
As a matter of interest, where does one get a list of the mkdocs plugin? http://www.mkdocs.org/user-guide/plugins/ doesn't seem to have a list?

waylan:
The Plugin API was brand new with the last release. We haven't had any plugins to add to the list until now. We should probably create a publicly editable wiki page so that anyone can add their own (much like we have for themes).

rosscdh:
Plus one..

rosscdh:
Oh hey thanks for that! :)

PR #1493, opened 2018-05-23
Make test_doc_dir_in_site_dir work with custom build path

### LINKED ISSUE #1490: test_doc_dir_in_site_dir fails when the checkout directory name is not ‘mkdocs’
If you clone this repository as `mkdocs`, then all the tests pass. However if you clone it with a different name, e.g. `mkdocs-0.17.3`, then one of the tests fails:
```
FAIL: test_doc_dir_in_site_dir (mkdocs.tests.config.config_tests.ConfigTests)
  File "/tmp/mkdocs-0.17.3/mkdocs/tests/config/config_tests.py", line 264, in test_doc_dir_in_site_dir
    self.assertEqual(len(errors), 1)
AssertionError: 0 != 1
```

This happens because the test expects this error:
https://github.com/mkdocs/mkdocs/blob/7e06ca68d100d27c0323a59a13a9a692c2ee537f/mkdocs/config/config_options.py#L369-L373

for this line:
https://github.com/mkdocs/mkdocs/blob/7e06ca68d100d27c0323a59a13a9a692c2ee537f/mkdocs/tests/config/config_tests.py#L242

But this happens only when `../mkdocs` points back to the current directory.

Note: I run the tests using `python3 setup.py build; PYTHONPATH=. python3 -m nose`. This also affects the Debian package build.

mitya57:
The previous version worked only when `../mkdocs` was pointing back to the current directory.

Fixes #1490.

### DIFF
--- a/mkdocs/tests/config/config_tests.py
+++ b/mkdocs/tests/config/config_tests.py
@@ -239,7 +239,7 @@ class ConfigTests(unittest.TestCase):
             {'docs_dir': '.', 'site_dir': '.'},
             {'docs_dir': 'docs', 'site_dir': ''},
             {'docs_dir': '', 'site_dir': ''},
-            {'docs_dir': j('..', 'mkdocs', 'docs'), 'site_dir': 'docs'},
+            {'docs_dir': 'docs', 'site_dir': 'docs'},
         )
 
         conf = {


waylan:
That was an easy fix. I seem to remember this not being that easy before. I suspect #1376 has something to do with that.

mitya57:
Yes, this was the easiest fix that came to my mind. Of course I could retrieve the current directory name and replace `'mkdocs'` with it, but in my opinion that would be an over-complication.

Issue #1382, opened 2018-01-17
Date in title regression when moving from v16 to v17 (readthedocs)


styfle:
I updated `mkdocs` from 0.16.0 to 0.17.2 and noticed the titles of my documents have changed.

I have a folder in the root of my site called `meetings` with several files that start with a date such as `2017-08-25-adfs-example.md`. Some of them are just a date such as `2017-09-12.md`.

## Before (v16)

In version 16, the hyphens were stripped and this was fine.

![v16](https://user-images.githubusercontent.com/229881/35050382-afc4e604-fb70-11e7-9c77-5f24e618648c.png)

## After (v17)

In version 17, the titles randomly change depending on which one I click. Sometimes it has a date in it, sometimes it doesnt.

![v17](https://user-images.githubusercontent.com/229881/35050532-0de2b5ae-fb71-11e7-9fe8-9cf729709699.png)

waylan:
Yes, in #1180 the way page titles are determined was refactored (see [nav.pyL213][1]). As explained in the docstring:

```
Get the title for a Markdown document
Check these in order and return the first that has a valid title:
- self._title which is populated from the mkdocs.yml
- self.meta['title'] which comes from the page metadata
- self.markdown - look for the first H1
- self.input_path - create a title based on the filename
```

If you were relying on the filename for page titles, that is the last possible method for determining a title. Presumably, one of the other methods are returning a title for some of your pages (my best guess is that some of your pages start with an `h1` header). If you manually define a title for those pages using a method with higher priority, then you will be guaranteed to always have the title you want.

Regarding the "regression" claim... The new feature (allow alternate ways to define page titles) has been a long-standing feature request (see #316). And using filenames has always been a fallback. Not defining titles and relying on the fallback is always going to be the most likely to run into unexpected changes.  I'm inclined to not consider this a regression.

[1]: https://github.com/mkdocs/mkdocs/pull/1180/files#diff-da047634a950ee20ecb0da37c95e229dR213

styfle:
I understand now. Some of the documents didn't have an `<h1>` so they still show the date.

However there appears to be a bug that is modifying the title in the menu bar depending on which document I click. It seems like when I click one file, the remaining files listed below go back to displaying  the file name (as a date) instead of the `<h1>`.

It's hard to explain so maybe I can try to reproduce with a new repository.

waylan:
[...]

Issue #1373, opened 2018-01-07
link a file at the root of the project


jodumont:
Hi everyone;

I tried to make a file which is at the root of my project into mkdocs
let's me try to explain

mkdocs render everything inside ./docs/*
my file is at the root level which like ./norendering.md instead of ./docs/rendering.md

an example I found is https://github.com/mkdocs/mkdocs/blob/master/CONTRIBUTING.md what if I want this file appear inside the website like this one https://github.com/mkdocs/mkdocs/blob/master/docs/index.md

regards!

Jonathan

ps: if rendering is not the appropriate words tell me how to speak python/jinga/mkdocs :)

waylan:
This appears to be related to #1134. As there is currently no open issue for this, I'll leave this open so we don't forget about it.

jodumont:
wow I just saw (https://github.com/mkdocs/mkdocs/commit/34ef3ca6d0390959080ce93a695361eea1649272#diff-55f7e7bab4ecafd524c20778d6920fdaR43)

I can't wait to test it :)

Thanks!

waylan:
@jodumont a beta has been released. Please do test it out. To install do:

    pip install --upgrade mkdocs=1.0b1

deltahedge1:
after sometime I found some extensions which did the trick, using `pymdown-extensions` and specifically the `snippets` ext you can now include files like README.md and changelog.md from your root directory. 

https://facelessuser.github.io/pymdown-extensions/extensions/snippets/

ssbarnea:
@deltahedge1 Sadly this approach does not work with mkdocs monorepos and probably neither with simple inclusion.

Issue #989, opened 2016-07-04
Allow absolute links to external sites in `pages` section of `mkdocs.yml`


nicksnyder:
e.g. mkdocs.yml

```
site_name: MyCoolSite

pages:
- Home: index.md
- Guide:
  - Know this: know-this.md
  - Other info: "http://www.otherinfo.com"
```

Currently `build_pages` in `build.py` fails

```
DEBUG   -  Building page http:/www.google.com 
ERROR   -  file not found: /Users/nsnyder/code/linkedin/layoutkit-opensource/docs/http:/www.google.com 
ERROR   -  Error building page http:/www.google.com 
```

It would be easy to skip the call to `_build_pages` if the value starts with `http` but also need to generate the correct absolute link in the nav. Happy to work on this if you agree.

If so, can you point me to the correct place for generating the link?

waylan:
No, this is currently not possible. A few similar requests (#797 and #575) have been punted to the upcoming Plugin API (#206).

However, thinking about this, any such plugin would need to know where in the pages structure the new nav item would need to go. We  could just say "add it within the plugin, but with multiple levels of nesting, that really limits what users can do. Either the plugin needs to know the structure of the pages and set through until it guesses the right location, or the plugin would just tack any new nav items on the end at the root level only. In other words, it's either really complex or extremely simple and limited. Perhaps we need to have some native support for a placeholder of some kind. Not really sure what that would look like though. Maybe a title with no data is understood to be a placeholder and just ignored. Then a plugin can use that placeholder as it sees fit (link to generated content, or an external resource, or ...).

nicksnyder:
Perhaps a plugin could register a handler for values.

Something like this? (warning, I haven't written Python in a while so read this like pseudocode).

```
for page in site_navigation.walk_pages():
        try:
            log.debug("Building page %s", page.input_path)
            for plugin in page_handler_plugins:
                if plugin.can_handle_page(page):
                    build_result = plugin.handle_page(page)
                    break
  
            if !build_result:
                build_result = _build_page(page, config, site_navigation, env, dump_json)
            
            html_content, table_of_contents, _ = build_result
            search_index.add_entry_from_context(
                page, html_content, table_of_contents)
        except Exception:
            log.error("Error building page %s", page.input_path)
            raise
```
In the meantime, it is easy to me to just create a dummy markdown page that does a redirect using a meta refresh tag.

erichgoldman:
[...]

PR #1504, opened 2018-06-08
Pages Refactor

waylan:
Probably still needs some tweaks, but I believe this is fully functional. I wanted to get this in front of others for review, especially third party theme authors (@squidfunk, @notpushkin, @chrissimpkins, @hfagerlund, @cjsheets, @michaeltlombardi).

Some of the backward incompatible changes do not have a graceful upgrade path, but the underlying structure is so drastic that its really any way to support both the old and new. As this is intended to be included in the 1.0 release, there is no requirement to maintain compatibility (we are making a major release) and we need to get this correct now as we will be stuck with it throughout the 1.x release cycle.

The documentation changes and release notes provide a decent summary of the changes made. I've copied the release notes below. Some additional information, including notes regarding some of the decisions made are included in the commit message for the second commit. The first commit is from @tomchristie's work in #689, which this builds off of.

> #### Internal Refactor of Pages, Files, and Navigation
> 
> Internal handling of pages, files and navigation has been completely refactored.
> The changes included in the refactor are summarized below.
> 
> * Support for hidden pages. All Markdown pages are now included in the build
>   regardless of whether they are included in the navigation configuration
>   (#699).
> * The navigation can now include links to external sites (#989 & #1373).
> * Page data (including titles) is properly determined for all pages before any
>   page is rendered (#1382).
> * Automatically populated navigation now sorts index pages to the top. In other
>   words, The index page will be listed as the first child of a directory, while
>   all other documents are sorted alphanumerically by file name after the index page (#1042).
> * A `README.md` file is now treated as an index file within a directory and
>   will be rendered to `index.html` (#608).
> * An [on_files] plugin event has been added, which could be used to include
>   files not in the `docs_dir`, exclude files, redefine page URLs (i.e.
>   implement extensionless URLs), or to manipulate files in various other ways.
> 
> ##### Backward Incompatible Changes
> 
> As part of the internal refactor, a number of backward incompatible changes have
> been introduced, which are summarized below.
> 
> ###### URLS have changed when `use_directory_urls` is `False`
> 
> Previously, all Markdown pages would be have their filenames altered to be index
> pages regardless of how the [use_directory_urls] setting was configured.
> However, the path munging is only needed when `use_directory_urls` is set to
> `True` (the default). The path mungling no longer happens when
> `use_directory_urls` is set to `False`, which will result in different URLs for
> all pages that were not already index files. As this behavior only effects a
> non-default configuration, and the most common user-case for setting the option
> to `False` is for local file system (`file://`) browsing, its not likely to
> effect most users. However, if you have `use_directory_urls` set to `False`
> for a MkDocs site hosted on a web server, most of your URLs will now be broken.
> As you can see below, the new URLs are much more sensible.
> 
> | Markdown file   | Old URL              | New URL        |
> | --------------- | -------------------- | -------------- |
> | `index.md`      | `index.html`         | `index.html`   |
> | `foo.md`        | `foo/index.html`     | `foo.html`     |
> | `foo/bar.md`    | `foo/bar/index.html` | `foo/bar.html` |
> 
> Note that there has been no change to URLs or file paths when
> `use_directory_urls` is set to `True` (the default), except that MkDocs more
> consistently includes an ending slash on all internally generated URLs.
> 
> ###### The `pages` configuration setting has been renamed to `nav`
> 
> The `pages` configuration setting is deprecated and will issue a warning if set
> in the configuration file. The setting has been renamed `nav`. To update your
> configuration, simply rename the setting to `nav`. In other words, if your
> configuration looked like this:
> 
> ```yaml
> pages:
>     - Home: index.md
>     - User Guide: user-guide.md
> ```
> 
> Simply edit the configuration as follows:
> 
> ```yaml
> nav:
>     - Home: index.md
>     - User Guide: user-guide.md
> ```
> 
> In the current release, any configuration which includes a `pages` setting, but
> no `nav` setting, the `pages` configuration will be copied to `nav` and a
> warning will be issued. However, in a future release, that may no longer happen.
> If both `pages` and `nav` are defined, the `pages` setting will be ignored.
> 
> ###### Template variables and `base_url`
> 
> In previous versions of MkDocs some URLs expected the [base_url] template
> variable to be prepended to the URL and others did not. That inconsistency has
> been removed. All  URLs must now be joined with the `base_url`. As previously, a
> slash must be included between the `base_url` and the URL variable. For example,
> a theme template might have previously included a link to the `site_name` as:
> 
> ```django
> <a href="{{ nav.homepage.url }}">{{ config.site_name }}</a>
> ```
> 
> And MkDocs would magically return a URL for the homepage which was relative to
> the current page. That "magic" has been removed and the `base_url` must now be
> explicitly included:
> 
> ```django
> <a href="{{ base_url }}/{{ nav.homepage.url }}">{{ config.site_name }}</a>
> ```
> 
> This change applies to any navigation items and pages, as well as the
> `page.next_page` and `page.previous_page` attributes. For the time being, the
> `extra_javascript` and `extra_css` variables continue to work as previously
> (without `base_url`), but they have been deprecated and the corresponding
> configuration values (`config.extra_javascript` and `config.extra_css`
> respectively) should be used with `base_url` instead.
> 
> Any URL variables which should not be used with `base_url` are explicitly
> documented as such.


chrissimpkins:
Thanks for the summary and poke.  Much appreciated! I will have a look at this over the weekend.  What is time frame for 1.0 rollout?

squidfunk:
Sounds great and looks good to me!

waylan:
The time frame for 1.0 is "when it's done." This PR should resolve about half of the open issues on the 1.0 milestone. Most everything else is just housekeeping type stuff, so I expect by the end of summer. That said, this is being worked in in my spare time so no guarantees.

chrissimpkins:
Wasn't a criticism, rather it was to place changes on our end on my TODO list radar in a timely fashion.  Sounds good to me.  The more time you need the better from my standpoint :)

tomchristie:
Holy smokes

waylan:
We're getting a [test failure][1] in the "integration tests" which do a "strict" build of the MkDocs documentation (the same issue is causing the linkchecker test to fail) . Of course, a "strict" build errors if any warnings are issued. And a build of the MkDocs documentation currently includes the following warning:

```
WARNING -  Documentation file 'index.md' contains a link to 'img\favicon.ico' 
which does not exist in the documentation directory.
```

That error is correct. However, the image does exist in the generated site as it comes from the theme. #872 requests the ability to validate image links in documentation pages and this refactor makes that very easy to implement (it appears that I missed listing that feature and issue in the release notes). The tricky part is when the link points to an internal file which comes from an external source (external links are ignored in this check). Should we include the non-template theme files in the check? A [comment][2] I inserted in the code seems to indicate I had intended to do something about this at one point. Perhaps we should even include non-template theme files in the `Files` collection? The problem is that the `Files` collection assumes there is only ever one file for any output location. However, it is documented and understood that theme files get written first, then files from the `docs_dir` can overwrite a file of the same name.

I was thinking I could deal with this later, but didn't consider that it would cause the Integration tests to fail.

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/389848158#L546
[2]: https://github.com/mkdocs/mkdocs/pull/1504/commits/ac6581b8ef338e301f65c72ecd720863cfd5cacf#diff-30fed26d76866f92435a3b102e91c182R219

charlesreid1:
Re #1504 and this change:

> A README.md file is now treated as an index file within a directory and
will be rendered to index.html (#608).

does this implement the behavior you described in [this comment](https://github.com/mkdocs/mkdocs/issues/608#issuecomment-328311230)?

> when no index.md is found within a dir within the docs_dir, but README.md is found, then README.md is used and treated as if it was index.md

waylan:
> is the index.md name being done away with entirely in favor of using README.md?

No. I never said that and why would we do that. README.md is an optional alternate index file for those people who want it.

hfagerlund:
Thanks @waylan. Looking forward to trying this out.

waylan:
While working on this, I discovered a weird error which I reported in #1516. As the error also exists in master, it appears to not be related to this work, so I'll address that separately.

waylan:
FYI to those following this, **especially theme authors**, the above described behavior for URLs and the use of `base_url` is being changed slightly due to an oversight on my part. See #1526 for details.

timani:
Thanks for all the updates @waylan, these are all awesome and welcome changes. Is there a way to test a dev build or an ETA of when this will be available as a stable release?

waylan:
@timani this will be included in the next release (1.0) when the [1.0 Milestone][1] is complete (we are almost there). In the meantime, you can always install the latest in the master branch directly from GitHub:

```
pip install https://github.com/mkdocs/mkdocs/archive/master.zip
```

[1]: https://github.com/mkdocs/mkdocs/milestone/4

timani:
great! I'll subscribe to the [1.0 Milestone](https://github.com/mkdocs/mkdocs/milestone/4) 

I also see [1.0 Beta](https://pypi.org/project/mkdocs/1.0b1/) is already out so that should do the trick
 
```
pip install mkdocs==1.0b1
```

thanks @waylan

PR #1505, opened 2018-06-11
docs: Replace screen_mobile.png reference with a working one.

mitya57:
See rtfd/readthedocs.org#3438.

Or maybe we should include a screenshot in this repository, to be independent of possible future changes?

### DIFF
--- a/docs/user-guide/styling-your-docs.md
+++ b/docs/user-guide/styling-your-docs.md
@@ -68,7 +68,7 @@ theme:
         - rust
 ```
 
-![ReadTheDocs](http://docs.readthedocs.io/en/latest/_images/screen_mobile.png)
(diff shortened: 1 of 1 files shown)


waylan:
I hadn't noticed the old link was broken, so thank you for that. However, the linked image is of a Sphinx theme. It would make more sense to provide an image of the MkDocs theme. In fact, given the long-running status of #588, its not likely that future changes to the Sphinx theme will make into the MkDocs theme very quickly. We could easily have a situation where there is a mismatch between the third party image and the actual theme (due to a future change on RTD's end).

Given the above, let's host our own image.

waylan:
Turns out we already have a [local image][1] in our docs.

[1]: https://github.com/mkdocs/mkdocs/blob/master/docs/img/readthedocs.png

PR #1518, opened 2018-06-28
Fix Python 3.7 and PEP 479 incompatibility

moigagoo:
With [PEP 479](https://www.python.org/dev/peps/pep-0479/) implemented in Python 3.7, raised `StopIteration` is now translated into `RuntimeError`, which breaks `mkdocs build` command.

The recommended way to exit generators is to use `return` instead of raising `StopIteartion`: https://www.python.org/dev/peps/pep-0479/#id39


waylan:
Thanks for your contribution. However, this is being closed in favor of #1504 which completely replaces the offending code with a new implementation that does not exhibit the problem.

waylan:
Including a mailing list report, we have received 5 separate reports in 8 days. I guess we should probably apply this fix and do a bugfix release. However, this PR needs to be against the 0.17 branch for that and it's currently against master.

moigagoo:
> However, this PR needs to be against the 0.17 branch for that and it's currently against master.

moigagoo:
@waylan 

> However, this PR needs to be against the 0.17 branch for that and it's currently against master.

Fixed.

moigagoo:
@waylan Oh, I just noticed that you had changed the base yourself back and forth, sorry :-(

waylan:
Notice that by switching the branch we get all of the commits on master up to and including the relevant commit. However, we only want the one relevant commit. Therefore, I cloned your repo locally, cherry-picked your commit and pushed it to #1536. No need for you to do do anything else at this point.

waylan:
Fixed in #1536.

waylan:
Version 0.17.5 is now available with this fix.

PR #1519, opened 2018-06-29
Support Py3.7: Replace StopIteration by return

dirkroorda:
In Python 3.7 it is 
[deprecated](https://www.python.org/dev/peps/pep-0479/)
 to exit a generator by `raise StopIteration`, you should say `return`.

I stumbled on this by upgrading to Python 3.7 and running mkdocs. There are three occurrences in the nav.py of the current release (0.17). Replacing them by `return` fixes the issue of mkdocs not running in 3.7.

But I saw that the code has been refactored since then, and the StopIterations are gone anyway.

So this is only a quick fix for now.


waylan:
The next release will be 1.0 which will contain the code in the master branch. If we were to release any bugfix releases (0.17.x), then we would only backport a simple fix for that specific bug only. For example, to fix the reported issue here, we might consider a backport of the fix in #1518 only to the 0.17 branch. I say "might" because 0.17 never claimed to support Python 3.7 and I intended to reserve official support till 1.0. Of course, at the time I didn't realize MkDocs wouldn't even run on Python 3.7.

waylan:
Version 0.17.5 is now available with the fix in #1536.

PR #1522, opened 2018-06-29
Document the builtin themes' lack of support for multi-level nav.

### LINKED ISSUE #1107: nested nav submenu not working in mobile mode
Mkdocs allows for nested menu's. But as described in the below bootstrap issue nested nav menu's are not working in mobile mode (you need to hover the item which is impossible on mobile).

https://github.com/twbs/bootstrap/issues/16387

For my mkdocs usages I am hoping to be able to use the following menu structure. But it should also work on mobile. Anyone a suggestion ?

```
- Products
  -- Product1 
      -- Datasheet
      -- Manual
      -- Quickstart guide
  -- Product2
     -- etc.
```

waylan:
Fixes #1107.

### DIFF
--- a/docs/user-guide/styling-your-docs.md
+++ b/docs/user-guide/styling-your-docs.md
@@ -25,8 +25,8 @@ the [Customizing a Theme][customize] section below.
 ### mkdocs
 
 The default theme, which was built as a custom [Bootstrap] theme, supports most
-every feature of MkDocs. It only supports the default
-[theme configuration options].
+every feature of MkDocs. It only supports the default [theme configuration
+options] and only officially supports two levels in the navigation (see #1107).
 
 ![mkdocs](/img/mkdocs.png)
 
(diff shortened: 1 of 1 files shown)


Vampire:
Tbh, I wouldn't say this "fixes" it, but ok. :-D
But one question, you also write for the RTD theme that only 2 levels are supported.
The RTD theme has the full tree expanded and the navigation works fine with it.
So multiple levels work fine.
At least with 0.17.5 that I installed through pip.
Or did this change for 1.0?

waylan:
The RTD theme is not stable yet.  I know two levels work, but cannot guarantee that any more will work. Thus the note in the documentation. When #588 is resolved, this should be cleared up.

> I wouldn't say this "fixes" it,

Fair enough. I  generally use `fixes` as that triggers GitHub's automation. GitHub does support `closes` as well, but I guess I use "fix" out of habit.

Vampire:
At least in 0.17.5 with the three levels in Gitolite docs it worked fine.

If #588 refers to the theme e. g. seen at https://mcforge.readthedocs.io/en/latest/models/introduction/, at least that one seems to properly support deeper levels. :-)

PR #1523, opened 2018-06-30
Fix search result url in default theme

tsotnikov:
After the pages refractor #1504 (good work btw) the resulting urls are malformed (relatives) when using the search feature in the default theme.

It can be reproduced with a simple search. Given this directory structure:
```
docs/
├── foo.md
└── index.md
```
Searching for `foo` will result in `http://127.0.0.1:8000/.foo/` instead of `http://127.0.0.1:8000/foo/`


waylan:
Good catch. And thanks for the fix. I did check that at one point, but I suspect I made some more path related changes after that and forgot to check again.

PR #1526, opened 2018-07-02
Add a url filter to templates.

waylan:
Currently all local URLs must be prepended with `base_url` while external 
links must not. However, we can have both types of links in the same 
collection. As prepreprocessing of extra_css and extra_javascript 
settings is deprecated, we need a way to indicate the URL should be 
prepended, but only if it needs to be. The same applies to external 
links included in the site navigation. 

Therefore, a new temaplte filter is provided which any URL can be passed 
to. If the URL is absolute, it is returned unaltered. If it is relative 
and the template context includes a `page` object, then a URL is 
returned reltive to the page object. Otherwise, the URL is returned with 
`base_url` prepended. This also results in cleaner output as paths are 
normailized (redundant `.` are removed). Of course, the `base_url` template 
variable remains (it is not being deprecated -- in fact the new filter 
relies on its existance) and template authors may continue to use it if 
they desire.

Docs still need updated and we may need a few tests.

### DIFF
--- a/mkdocs/theme.py
+++ b/mkdocs/theme.py
@@ -111,4 +111,5 @@ class Theme(object):
         loader = jinja2.FileSystemLoader(self.dirs)
         env = jinja2.Environment(loader=loader)
         env.filters['tojson'] = filters.tojson
+        env.filters['url'] = filters.url_filter
         return env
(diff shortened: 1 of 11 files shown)


waylan:
Sigh. The tests are failing due to a [new rule][1] in [Markdownlint 0.5.0][2]. But that will be handled separately from this PR.

[1]: https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md046---code-block-style
[2]: https://github.com/markdownlint/markdownlint/blob/master/CHANGELOG.md#v050

PR #1534, opened 2018-07-06
Optimize utils.copy_file based on file size.

### LINKED ISSUE #1331: Slow documentation building with images
I have a folder with 44 images under `docs/images` (made of png, jgf, gif), weighting a total of 2.1MB.

Since I added that image folder, `mkdocs serve` takes a very long time to build (on startup only). I'd say 10-15 minutes, while it was taking only a couple of seconds before.

I don't find anything relevant in the doc, am I missing something?

waylan:
Fixes #1331... maybe...

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -202,6 +202,7 @@ authors should review how [search and themes] interact.
 
 ### Other Changes and Additions to Development Version
 
+* Optimize `mkdocs.utils.copy_file` based on file size (#1331).
 * Add highlight.js configuration settings to built-in themes (#1284).
 * Close search modal when result is selected (#1527).
 * Add a level attribute to AnchorLinks (#1272).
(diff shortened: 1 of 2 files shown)


waylan:
Well, it does improve performance, but not by much. I don't think the minimal gains are worth the extra maintenance burden.

```
$ python3 test_copy_file.py
Custom Big: 103.38457301200833
Builtin Big: 105.15060644998448
Custom Small: 40.39226396899903
Builtin Small: 52.93522975000087
```

The test code is [here](https://gist.github.com/waylan/339f102126c150544c6ac8afcbed6ba2).

waylan:
For kicks I tried calling `cp` as a subprocess and that didn't help much:

```
CP Big: 112.57949149602791
CP Small: 48.655360768025275
```

I suppose it's possible that my system has enough resources available that not a problem to begin with and the results may be different on a less powerful system. But from my perspective, there's not much improvement to be gained.

PR #1536, opened 2018-07-06
Fix Python 3.7 and PEP 479 incompatibility

waylan:
With [PEP 479](https://www.python.org/dev/peps/pep-0479/) implemented in Python 3.7, raised `StopIteration` is now translated into `RuntimeError`, which breaks `mkdocs build` command.

The recommended way to exit generators is to use `return` instead of raising `StopIteartion`: https://www.python.org/dev/peps/pep-0479/#id39

Fixes #1518.

### DIFF
--- a/mkdocs/nav.py
+++ b/mkdocs/nav.py
@@ -334,7 +334,7 @@ def _follow(config_line, url_context, config, header=None, title=None):
             header.children.append(page)
 
         yield page
-        raise StopIteration
+        return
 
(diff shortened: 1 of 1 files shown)


waylan:
The failing tests are related to upstream changes that we are not addressing in the 0.17 branch. Therefore, this is being accepted with the failing tests.

waylan:
Version 0.17.5 is now available with this fix.

PR #1545, opened 2018-07-12
Add official support for PY37.

waylan:


facelessuser:
You can add `3.7-dev` to test Py3.7 until Travis catches up and formally supports the latest 3.7.  That is what I currently do.

PR #1547, opened 2018-07-12
Test the Python 3.7 release in Travis

jdufresne:


waylan:
According to travis-ci/travis-ci#9815 this appears to be the correct way to get tests to run on Python 3.7. Thanks for the contribution.

Issue #1374, opened 2018-01-08
Title doubles


Serhioromano:
I have a file in subdirectory. Let's say I call it ACL.md. In config it is used like this

```
pages:
- Developer:
    - ACL: ACL.md
```

The file starts with H1 tag and then other titles like a structure. The H1 title is the same meaning as the title in config file. So in navigation title doubles and create unnecessary level. I have only one H1 per file.

Is it possible to tell mkdocs not to use first H1 in menu navigation? I have a screenshot, not sure if it is helpfull because it is on russian.

![2018-01-08_13-24-05](https://user-images.githubusercontent.com/650741/34661992-989b69e4-f477-11e7-963b-9f418c7775d3.png)

1 - is a title in configuration

    - Метод API: 'developer/apiendpoint.md'

2 and 3 is H1 title in the document. how to avoid this?

algo99:
++
Looks like feature request for me as well.

We solve this though not using H1 headers in our index.md files.

waylan:
At this time there is no builtin way to account for this. However, you might be able to address this through [theme customization][1]. I haven't looked at the template, not sure if it would be a simple fix or not.

[1]: http://www.mkdocs.org/user-guide/styling-your-docs/#using-the-theme-custom_dir

squidfunk:
Material uses a hack to determine if a `h1` is included in the Markdown source, see:

https://github.com/squidfunk/mkdocs-material/blob/5bc0875e7e3bb5edfb984a516b22cc74acb5d257/src/partials/toc.html#L29-L37

waylan:
@squidfunk interesting approach. Not sure how that it supposed to work in the layout for the readthedocs theme though. We would then have children with no parent. I suspect the right approach would be to skip/exclude the child link to the `h1`. The parent link to the "page" would get you there anyway.

squidfunk:
This hack has been in Material for a long, long time and it works reasonably well. It just kills the first layer of TOC if it is an `h1`, but it also assumes that there is only one `h1` per document, which is semantically correct.

However, just wanted to let you know how I solved the problem.

Serhioromano:
@squidfunk Of course it is solvable with hack. i can override navigation file and exclude that. But that way I rather solve my own problem but not the problem itself.

On the other hand Python is not my strongest side. If it would be PHP or JS, I would create PR instead of issue.

cjsheets:
[...]

PR #1594, opened 2018-08-16
Update readthedocs theme with upstream

### LINKED ISSUE #588: Update Read The Docs theme with upstream

waylan:
This updates the `readthedocs` theme to more closely match the [upstream][1] Sphinx theme (at version 0.4.1). This should resolve #588 and #1374. 

The following theme config settings were added:

* __`include_homepage_in_sidebar`__: Lists the homepage in the sidebar menu. As
  MkDocs requires that the homepage be listed in the `nav` configuration
  option, this setting allows the homepage to be included or excluded from
  the sidebar. Note that the site name/logo always links to the homepage.
  Default: `True`.

* __`prev_next_buttons_location`__: One of `bottom`, `top`, `both` , or `none`.
  Displays the “Next” and “Previous” buttons accordingly. Default: `bottom`.

* __`navigation_depth`__: The maximum depth of the navigation tree in the
  sidebar. Default: `4`.

* __`titles_only`__: If True, removes headers within a page from the sidebar.
  Default: `False`.

[1]: https://github.com/rtfd/sphinx_rtd_theme/


Blendify:
There is no need to feel sorry, I am glad to help. I will take a look.

Blendify:
Some toc options were not included these include:

- `includehidden = True`
- `collapse_navigation = True`
- `sticky_navigation = True`

Is there a reason these are not supported?

waylan:
> Some toc options were not included these include:
>
> * `includehidden = True`

MkDocs' concept of hidden pages is very different from Sphinx and therefore this option makes no sense. As I understand it, Sphinx page objects include a `hidden` attribute. MkDocs page object have no such attribute. A hidden page is simply not included in the `nav` at all. Therefore, there is no way to include such pages.

Or am I misunderstanding the purpose of this setting?

> * `collapse_navigation = True`
> * `sticky_navigation = True`

I haven't worked on these. Basically, I need to determine what is different about the HTML with these options enable/disabled and add the same logic to our template. Personally they are not high priority items so I haven't got to them yet. And the assumption is they could always be added at any time.

waylan:
`stick_navigation` has been added.

I'm struggling to see the difference between `collapse_navigation` and `titles_only`. The documentation seems clearly different, but in practice both setting simply exclude TOC items from the nav. As I already implemented that feature in `titles_only`, what value is there is also adding `collapse_navigation` ?

As an aside, the [documentation][2] for `collapse_navigation`  states: 

> * `collapse_navigation` Bool. With this enabled, you will lose the `[+]` drop downs next to each section in the sidebar.

But that is not strictly true. You still get `[+]` drop downs, just not as many levels of them (tested with the docs in [rtfd/sphinx_rtd_theme][1]). I find that very misleading and its not something I'm inclined to replicate... unless someone can explain to me the value of the setting and/or what it is supposed to be doing.

[1]: https://github.com/rtfd/sphinx_rtd_theme/
[2]: https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html#toc-options

waylan:
@Blendify thank you. Those docs are must more clear. I now have the following implemented:

> * __`include_homepage_in_sidebar`__: Lists the homepage in the sidebar menu. As
>   MkDocs requires that the homepage be listed in the `nav` configuration
>   option, this setting allows the homepage to be included or excluded from
>   the sidebar. Note that the site name/logo always links to the homepage.
>   Default: `True`.
> 
> * __`prev_next_buttons_location`__: One of `bottom`, `top`, `both` , or `none`.
>   Displays the “Next” and “Previous” buttons accordingly. Default: `bottom`.
> 
> * __`navigation_depth`__: The maximum depth of the navigation tree in the
>   sidebar. Default: `4`.
> 
> * __`collapse_navigation`__: Only include the page section headers in the
>   sidebar for the current page. Default: `True`.
> 
> * __`titles_only`__: Only include page titles in the sidebar, excluding all
>   section headers for all pages. Default: `False`.
> 
> * __`sticky_navigation`__: If True, causes the sidebar to scroll with the main
>   page content as you scroll the page. Default: `True`.

Blendify:
I will take another look again, but from my first look over, it looked pretty good.

Issue #1563, opened 2018-07-22
Update mkdocs theme from Bootstrap 2 to 4


ajhstn:
Please include `badges` (and other) very useful styles used in Documentation projects. 

- [badges](https://getbootstrap.com/docs/4.1/components/badge/)

waylan:
A [quick look][1] at Bootstrap 4 indicates that like Bootstrap 3, only a single level of nesting is supported in menus. However, as we are documenting that the `mkdocs` theme only supports that many levels of nesting (see #1522), I suppose we don't have any reason to not update to a more recent Bootstrap version. The one possible issue with that is that some users may be using more than one level of nesting and have not noticed that mobile mode is broken. With the update, their nav would be broken in both modes.

In any event, I'm changing the scope of this issue to be about updating to Bootstrap 4.

[1]: https://getbootstrap.com/docs/4.1/components/navs/#pills-with-dropdowns

waylan:
As a reminder, the `mkdocs` theme is a custom [Bootswatch][1] theme (Bootswatch themes are built on top of bootstrap). The problem with doing this is that no one has the original source `scss` files used to build the custom Bootswatch theme. I'm not even sure which Bootswatch theme was used as the starting point, although I suspect the "default".

There are two ways to build a custom Bootswatch theme.

1. Start with the raw source files from the [thomaspark/bootswatch][4] repo. Edit `_variables.scss` (and possibly `_bootswatch.scss`) and then build the theme with a `grunt` command. See [the docs][2] for details. 
2. Use Bootstrap's [builder tool][3] available through a web interface. This will let you import your bootswatch theme of choice to start, and then you can edit variables in a web form. When done, you can download a built set of theme files.  IIRC, you can export your custom settings and save for later import.

What we need is for someone to go through the existing theme and figure out what all the variables were when it was first built. Then we can import them into a newer version to get an updated theme. And we also should save those variables for the next time we want to update.

[1]: https://bootswatch.com/
[2]: https://bootswatch.com/help/
[3]: https://bootstrap.build/
[4]: https://github.com/thomaspark/bootswatch

mitya57:
> The problem with doing this is that no one has the original source scss files used to build the custom Bootswatch theme. I'm not even sure which Bootswatch theme was used as the starting point, although I suspect the "default".

It was the “cerulean” theme. Our CSS file is identical to that in 3.0.3+1 tag in Bootswatch Git repository:

[...]

PR #1597, opened 2018-08-20
Update to bootswatch 4.1.3.

waylan:
This updates the CSS and JS files. However, updates need to be made to 
the templates or something. Everything looks wonky.

Addresses #1563.


waylan:
This is what it looks like with the changes so far:

untitled

Someone needs to go through everything and make sure all the HTML is up-to-date with the latest bootstrap/bootswatch API. And double check all the tweaks in base.css.

mitya57:
@waylan Please look at the [bs-update branch in my fork](https://github.com/mitya57/mkdocs/commits/bs-update). With my changes it looks like this:

mkdocs documentation screenshot

Not pixel-to-pixel equivalent to the Bootstrap 2 version, but in my opinion it looks good enough. Do you want me to open a new pull request for it, or you can just merge my branch into your own?

waylan:
@mitya57 as a "contributor" you should be able to push to any branch from any user which has an open pull request (unless the user selected the box disabling that - but no one does). Therefore you can clone my branch, merge your changes and push back up to my repo. Then your changes will appear in this PR.

waylan:
The search modal is much narrower than before. Not sure if that is a problem, but it is an obvious difference.

There is also some weirdness with the sidebar.  At the middle of the three supported widths, it looks wrong:

untitled

Whereas the narrowest width is fine:

untitled1

I checked the old version and it only supported two widths. Looks like we should have the middle width act the same as the narrow width. Although, It might make sense to have the TOC collapsible in these situations. However, collapsible TOC would be a new feature so that's nice-to-have, but not necessary for this PR.

There is also some weirdness when the sidebar runs off the bottom:

untitled2

mitya57:
I do not have push access to your fork ­— probably only project members have it.

Issues with code blocks and narrow search modal were easy to fix — pushed as 5d9137412af322f3 and 2e557ea650093a7b.

The TOC is more tricky. The current code had its own implementation of `col-md-3` and `col-md-9` (using padding) which is quite hacky and error-prone. Also it had its own implementation of sticky position.

As I did not want to add even more hacks, I searched for a solution which will rely on Boostrap or browser features, and found [`position: sticky`](https://getbootstrap.com/docs/4.0/migration/#components). I made use of it in 473a0f31bdb9f81c and it works fine (tested Firefox, Chrome and Android browser), but has some drawbacks:
- It is [not supported by older browsers](https://caniuse.com/#search=sticky) — in particular by IE (we can add a [polyfill](https://github.com/wilddeer/stickyfill) if we care about it).
- csslint complains about it — see CSSLint/csslint#719. From that issue it looks like csslint is not maintained properly, maybe [stylelint](https://stylelint.io/) is a better alternative.

Please let me know what you think about this solution. If it is not acceptable then I will try to fix width by adding more hacks to the existing CSS code.

I also tried to make TOC collapsible but it does not look good. Instead, I can completely hide it on small screens. I pushed that as 0e33dc7385b6bcc1. If you do not agree with this approach then don’t pick that commit and I will remove it.

facelessuser:
I've personally found ` position: sticky` to behave differently on chrome between Windows and Mac, not to mention across browsers. I know some people use it, but I currently find it buggy.

waylan:
In a rather quick  searching I found [these][1] [two][2] approaches to collapsing a sidebar. I suspect something like that might make sense.

I'm okay with abandoning a hacky column implementation and updating the template to use current bootstrap conventions.

[1]: https://www.codeply.com/embed/zyKyFfeUcH
[2]: https://www.codeply.com/embed/sOaent8Azk

facelessuser:
@mitya57, there were some edge cases, how padding/margins was treated with sticky etc.; I witnessed this in chrome across platforms.  I couldn't use the same CSS to style it on Chrome across platforms which is ridiculous. As far as I know they are present in the most recent.  I'm not sure if you will run into them here.  I was trying to use it to sticky inline code numbers in code blocks, which worked, but I just couldn't get a single CSS to style it proper across chrome, little or less across all browsers.  It was such a hassle I dropped trying to use it.  I wasn't concerned that it didn't work in all browsers, I was concerned that it was treated differently across platforms and web browsers making it hard for me to style reliably with CSS and get an okay look in all browsers.

Depending on how you use it, maybe it won't be a problem , but in my case, it just wasn't worth the pain.

waylan:
@waylan I know, but the hanging expand button and the placement of menu do not look nice in my opinion.

I agree. I might like the second one if the sidebar stayed on the left rather than switching to the right. I see those more as a place to start to come up with something better. I'm really not keen on having the TOC hidden completely. On a small screen, linking to a section of a long document is even more important IMO.

waylan:
We should probably address the various outstanding issues with this theme while we're at it. That is, many of the issues in the [theme][0] tag. My plan is to include this and #1594 (which addresses the other half of the issues in the theme tag) as 1.1.

Also, as a reminder, the [mkdocs-bootstrap][1] and [mkdocs-bootswatch][2] themes inherit from this theme, so any updates here will need to be partially mirrored there.

[0]: https://github.com/mkdocs/mkdocs/issues?q=is%3Aopen+is%3Aissue+label%3ATheme
[1]: https://github.com/mkdocs/mkdocs-bootstrap
[2]: https://github.com/mkdocs/mkdocs-bootswatch

mitya57:
> I might like the second one if the sidebar stayed on the left rather than switching to the right.

Ok, I will continue my experiments when I have time, maybe tomorrow.

waylan:
For what its worth, I tested the current implementation (using sticky) on both Chrome/Windows and Chrome/Mac and didn't see any difference in behavior. However, I did not test any other browsers, just the problem @facelessuser reported encountered previously.

waylan:
FYI, I broke the theme tag up into multiple tags. All `mkdocs` theme related issues are now tagged [Theme-mkdocs](https://github.com/mkdocs/mkdocs/labels/Theme-mkdocs).

facelessuser:
Cool. I think what I saw was in a very narrow case with certain CSS, so if you don't see an issue, you're probably fine in this case.

mitya57:
I removed the last commit and pushed an alternative one (5458426bcac892c5) — collapsing the TOC when it does not fit in the side panel. How does that look like?

mitya57:
Also, for ease of testing from mobile devices, I deployed mkdocs documentation built with my changes to https://mitya57.me/mkdocs-bs4/.

waylan:
I think I merged your changes correctly. I like it. Although I made one small tweak to the icon for the expandable TOC.

mitya57:
You merged the changes correctly, thanks. Do you think it makes sense to try stylelint instead of csslint?

waylan:
I used the `fa-angle-down` as the icon, which is small and lightweight. However, if it is to small for touch devices, `fa-caret-down` might work better. I thought it looked too heavy, which is why I went with `fa-angle-down`, but target size is probably more important. Actual testing on a device is needed, so thanks for offering that. Let me know when you update.

waylan:
> Do you think it makes sense to try stylelint instead of csslint?

I'm indifferent. The only reason we went with csslint to begin with was because it seemed to be a widely used tool. Of course, any such change should probably be a separate PR from this one.

mitya57:
> Let me know when you update.

Done.

mitya57:
I have found a solution simpler than switching to stylelint: we just need to install parserlib from Git. I have created a new branch [csslint](https://github.com/mitya57/mkdocs/commits/csslint) and the Travis tests are passing there.

Feel free to include it here, or leave it for a separate merge request.

Speaking about other issues with theme-mkdocs tag, I don’t think this pull request fixes any of them, and they are less trivial to fix so I would leave them to somebody else.

waylan:
Testing on a small screen (iPhone 5S), the `fa-angle-down` icon is fine. so I'm leaving it. Also, I added a vertical flip to it when it toggles open/closed.

I addressed the failing csslint test by simply telling csslint to ignore those two lines. If/when csslint is updated, we can remove the ignore comments. Or if we switch to a different tool, we'll need to audit all csslint comments at that time.

@mitya57 thanks for spearheading the work on this. Unless you have something else you want to do with this, I'm considering the work done. However, I may not merge right away. I'm not quite ready to start working on 1.1 in earnest yet.

PR #1607, opened 2018-08-25
Use lunr.py for prebuilding indices

yeraydiazdiaz:
[Many moons ago](https://github.com/mkdocs/mkdocs/issues/859#issuecomment-251216390) there was a discussion that sparked my interest in creating [lunr.py](https://github.com/yeraydiazdiaz/lunr.py). Today I released v0.5.2 introducing multilanguage support and I feel it's time to bring it back to the source 🙂 

Changes include:

- Adding `lunr[languages]>=0.5.2` as a dependency, more on this later
- `prebuild_index` is now a string with two possible values `node` or `python`, directing the search plugin to use lunr.js or lunr.py respectively
- Update bundled lunr.js version to 2.3.2

I went for the more direct approach: requiring lunr + language support which in turn pulls NLTK as a dependency. This may not be desirable but it's the only way to guarantee complete support and it's roughly what we're doing by bundling lunr.js + lunr.languages when prebuilding using Node. I am however completely open to suggestions in this front.

I also added some timing information on the build process since, annoyingly, it takes about twice as long to build the index in Lunr.py that it is to build it in Lunr.js... 😞  I guess I have some profiling to do.

I am, of course, also open to suggestions on how to best define the `prebuild_index` setting, seemed to me users might want all three options.

Congrats on v1.0.0 btw 💯

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -43,5 +43,4 @@ copyright: Copyright &copy; 2014 <a href="https://twitter.com/_tomchristie">Tom 
 google_analytics: ['UA-27795084-5', 'mkdocs.org']
 
 plugins:
-    - search:
-        prebuild_index: python
+    - search
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -43,4 +43,5 @@ copyright: Copyright &copy; 2014 <a href="https://twitter.com/_tomchristie">Tom 
 google_analytics: ['UA-27795084-5', 'mkdocs.org']
 
 plugins:
-    - search
+    - search:
+        prebuild_index: python
--- a/mkdocs/contrib/search/__init__.py
+++ b/mkdocs/contrib/search/__init__.py
@@ -40,7 +40,7 @@ class SearchPlugin(BasePlugin):
     config_scheme = (
         ('lang', LangOption(default=['en'])),
         ('separator', config_options.Type(utils.string_types, default=r'[\s\-]+')),
-        ('prebuild_index', config_options.Type(utils.string_types, default='')),
+        ('prebuild_index', config_options.Choice((False, True, 'node', 'python'), default=False)),
     )
 
     def on_config(self, config, **kwargs):
--- a/mkdocs/contrib/search/__init__.py
+++ b/mkdocs/contrib/search/__init__.py
@@ -40,7 +40,7 @@ class SearchPlugin(BasePlugin):
     config_scheme = (
         ('lang', LangOption(default=['en'])),
         ('separator', config_options.Type(utils.string_types, default=r'[\s\-]+')),
-        ('prebuild_index', config_options.Type(bool, default=False)),
+        ('prebuild_index', config_options.Type(utils.string_types, default='')),
     )
 
     def on_config(self, config, **kwargs):
--- a/mkdocs/contrib/search/search_index.py
+++ b/mkdocs/contrib/search/search_index.py
@@ -7,6 +7,9 @@ import re
 import json
 import logging
 import subprocess
+
+from lunr import lunr
+
 from mkdocs import utils
 
 try:                                    # pragma: no cover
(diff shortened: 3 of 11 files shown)


mkdocs.yml
```diff
@@ -43,4 +43,5 @@ copyright: Copyright &copy; 2014 <a href="https://twitter.com/_tomchristie">Tom
 google_analytics: ['UA-27795084-5', 'mkdocs.org']
 
 plugins:
-    - search
+    - search:
+        prebuild_index: python
```
waylan: I realize having this enabled helps with testing, but I think perhaps we should not have this enabled for our own documentation. According to my testing (with the node implementation some months back), the size of our documentation does not justify the extra size to the index by prebuilding. That is, unless you can provide some clear benefits.

mkdocs/contrib/search/__init__.py
```diff
@@ -40,7 +40,7 @@ class SearchPlugin(BasePlugin):
     config_scheme = (
         ('lang', LangOption(default=['en'])),
         ('separator', config_options.Type(utils.string_types, default=r'[\s\-]+')),
-        ('prebuild_index', config_options.Type(bool, default=False)),
+        ('prebuild_index', config_options.Type(utils.string_types, default='')),
```
waylan: I understand the change here, but what about users who have `True` or `False` already set in their config? This will error as an invalid type. Seems to me `False` should be treated the same as an empty string and `True` should be treated as `'node'`.

mkdocs/contrib/search/search_index.py
```diff
@@ -105,7 +108,7 @@ def generate_search_index(self):
         }
         data = json.dumps(page_dicts, sort_keys=True, separators=(',', ':'))
 
-        if self.config['prebuild_index']:
+        if self.config['prebuild_index'] == 'node':
```
waylan: This should account for users who previously set the config value to `True`. There are two ways to do this:

1. Leave the user provided value in the config and account for it here with:

        if self.config['prebuild_index'] in ['node', True]:

2. Create a custom config validation class which transforms `True` and `False` values to `'node'` and `''` respectively.

Option 1 is obviously less work, but option 2 provides a way to provide clear error messages, including any deprecation warnings to encourage users to update their config files for a graceful deprecation.

Naturally, either way would need some additional tests to ensure the proper results. For option 1 that would mean replicating the existing tests with a set which pass in  `True` or `False` as config values. For option 2, we would simply need some tests for the custom config validation class to ensure it always resulted in a value containing the proper string.

mkdocs.yml
yeraydiazdiaz: absolutely, I included it to show how the `prebuild_index` looks like now, I'll remove it 👍

mkdocs/contrib/search/search_index.py
yeraydiazdiaz: I went with a [`Choice` config validation type](https://github.com/mkdocs/mkdocs/pull/1607/commits/d1ed542a396c7b6962c8cdd57d6ebaf588d8b2de) that accepts one of a set of predefined choices, seemed more flexible that a custom made one just for this case.

yeraydiazdiaz:
Thanks for the feedback @waylan, I addressed your comments.

Every version of Lunr.py targets a particular lunr.js version, in the case of 0.5.2 it's 2.3.2 which is also the latest. This version is added to the serialized index and upon loading it in the frontend if the version doesn't match a warning pops up in the console, so I decided we might as well update it.

mkdocs/tests/config/config_options_tests.py
```diff
@@ -80,6 +80,24 @@ def test_length(self):
                           option.validate, "Testing Long")
 
 
+class ChoiceTest(unittest.TestCase):
+
+    def test_valid_choice(self):
+        option = config_options.Choice(('yes', 'no'))
```
waylan: `yes` and `no` seem like good general strings to use for testing. However, we may never see them in actual use as the YAML parser will convert them to bool values (I think). This doesn't really require a change to the test but it might be less confusing to someone reading the source who doesn't know about YAMl's type handling if we use something else.

Not a necessary change, but might be nice to have.

waylan:
@yeraydiazdiaz if you rebase, resolve conflicts, and add to the release notes, I'll merge this.

yeraydiazdiaz:
@waylan updated now 👍

PR #1613, opened 2018-08-28
Warn on relative paths in nav.

### LINKED ISSUE #1604: mkdocs does not fail if in strict mode and if navigation table refers to non-existent file
If I add a reference in my table of contents to a non-existent markdown file, and I run mkdocs in strict mode (either by specifying `strict: true` in `mkdocs.yml` or by running `mkdocs build --strict`), I would expect it to fail. However, it does not. Prior versions of mkdocs (0.17) _would_ fail if the table of contents contained a reference to a non-existent file (even without strict mode).

1) Why would strict mode not check for non-existent files included in the table of contents? It doesn't make sense to leave that out.

2) Is there any possibility the strit behavior will be changed to be more like 0.17? Or am I stuck with this and should I sink time into writing my own function to enumerate all files and check to see if files listed in the TOC actually exist (which, ahem, mkdocs _was doing before just fine_)?

waylan:
Note that prior to 1.0 this was handled as part of the config 
validation. Now that it is moved out of config validation, we no longer 
have the option to collect all wartnings and then raise an error on 
strict mode. We now exist on the first error and never get to any later 
errors.

Closes #1604.

### DIFF
--- a/mkdocs/exceptions.py
+++ b/mkdocs/exceptions.py
@@ -11,4 +11,4 @@ class ConfigurationError(MkDocsException):
 
 
 class MarkdownNotFound(MkDocsException):
-    """A linked local Markdown file isn't found in the table of contents."""
+    """A linked local Markdown file isn't found in the documents."""
--- a/mkdocs/structure/nav.py
+++ b/mkdocs/structure/nav.py
@@ -4,7 +4,8 @@ from __future__ import unicode_literals
 import logging
 
 from mkdocs.structure.pages import Page
-from mkdocs.utils import string_types, nest_paths
+from mkdocs.utils import string_types, nest_paths, urlparse
+from mkdocs.exceptions import ConfigurationError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


waylan:
This should probably have some tests.

> We now exist on the first error and never get to any later errors.

I wonder if we could add a handler or filter to the logger which checks for any `WARNING` level messages and exits with a failing return code on strict mode. Then within the code, we could simply call `log.warning` as many times as we need and only exit after all warnings have accumulated.

As it stands now, strict mode's "fail on warnings" is hard-coded to the config validation only, not to MkDocs generally. It has always been that way, but today a lot of non-config validation stuff has been moved out of config validation, so strict mode currently applies to less than it did unless we check `config['strict']` at each location in the code and explicitly raise an error.

waylan:
Just added a logging filter which counts all `WARNING` logging messages and exits with return code `1` and a message if one or more `WARNING` logging messages are issued. Output looks like this:

```sh
$ mkdocs build --strict
INFO    -  Building documentation to directory: /mkdocs/site
WARNING -  A relative path to 'example.com' is included in the 'nav' configuration, which does not exist in the docs directory
WARNING -  A relative path to './relative/' is included in the 'nav' configuration, which does not exist in the docs directory

Exited with 2 warnings in strict mode.
$ echo $?
1
```

With verbose mode you get more info about other non-existent items in nav:

```
DEBUG   -  An external link to 'http://example.com' is included in the 'nav' configuration.
WARNING -  A relative path to 'example.com' is included in the 'nav' configuration, which does not exist in the docs directory
DEBUG   -  An absolute path to '/absolute/' is included in the 'nav' configuration, which presumably points to an external resource.
WARNING -  A relative path to './relative/' is included in the 'nav' configuration, which does not exist in the docs directory
```

But only the `WARNING` items will cause MkDocs to exit in strict mode.

Something I like about this is that `example.com` is considered a local file whereas `http://example.com` is recognized as an external link. Leaving off the scheme is a mistake some users are likely to make and the warning (strict mode or not) will help them work out the problem.

waylan:
Perhaps an important distinction to note: Warnings issued in config validation interrupt the build after config validation is complete--causing the build to not complete. Warnings issued during the build (after config validation passes) allow the build to complete and only exit with a failed return code.

PR #1642, opened 2018-09-27
Bump version to 1.1.dev

waylan:


docs/about/release-notes.md
```diff
@@ -21,6 +21,10 @@ The current and past members of the MkDocs team.
 * [@d0ugal](https://github.com/d0ugal/)
 * [@waylan](https://github.com/waylan/)
 
+** Version 1.1 (*in development*)
```
mitya57: Why `**` instead of `##`?

waylan: Whoops. Thanks for catching that.

Issue #1639, opened 2018-09-25
Use `rel=external` on external links


ppKrauss:
In any hipertextual redaction and final layout is important to distinguish external links and local-document (or local-hypertext) links... There are a way to show with different styles the local and external links?

(if not, it is a suggestion for issue)

-----

* External link (`http://external`)
* Local page link (`here`)
* Intra-link (`here#here`)

Most important to distinguish external from others... If possible, a little distinction between local-page and intra-link.

waylan:
Sorry but I don't follow. Is this a question, a bug report or a feature request?

ppKrauss:
Hi @waylan, well... Is a  **feature request**.

ppKrauss:
Request to mkdocs generate HTML with attribute `rel`  on links, `rel=external`  when the link is external. See https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

waylan:
I've recently given this some consideration myself. It seems like a reasonable request and expectation, especially as we are already determining whether each link is an internal or external link.

One way to accomplish that is by setting `rel=external` on external links.

Another option would be to replicate Sphinx, which sets `class="link external"` or `class="link internal"` on all links (IIRC).  A benefit of that solution would be that themes which clone Sphinx themes (such as `readthedocs`) already include CSS for styling internal and external links differently. However, that in itself isn't a reason to chose this option.

Personally, I'm inclined to go with `rel=external` as that is the more semantic and universally understood method for identifying external links.

waylan:
Markdown links are handled by the private Markdown [extension][1] so any changes to the Markdown output would happen there. Any other links (such as navigation) would be controlled by theme templates.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/structure/pages.py#L188

berot3:
Hi! Not sure, but I think this would make it easier to mark all external links in mkdocs as external (with a icon/symbol or so), right?

But at least there are workarounds: https://gitter.im/squidfunk/mkdocs-material?at=5d9dd20192920c36a13b7dd5

It would be great if this would be already handled by mkdocs of course.

ppKrauss:
Hi @berot3 , your comment make sense, but ... hum... This is my personal view:

* it is not a good practice to use regular expression (checking value-by-value about `http|https|ftp`) in a CSS selector.
* there are situations where we can express local site with different subdomain (ex. API), so it is not a good heuristic.

[...]

PR #1643, opened 2018-09-28
Set rel="external" on external links

ppKrauss:
Hi, please check if it is correct... and perhaps opportunity to some reuse.  The focus here was only to suggest a simple solution to the issue #1639, adding the attribute `rel="external"` in an `<a href>` tag that seems not local.

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -206,12 +206,15 @@ class _RelativePathTreeprocessor(Treeprocessor):
                 continue
 
             url = element.get(key)
-            new_url = self.path_to_url(url)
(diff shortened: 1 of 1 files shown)


ppKrauss:
> why not have path_to_url return a tuple of the url and a Boolean 

Yes, ideal, but the code need the original schema, not the returned (local can be a HTTP?), so the best is to use components as parameter.
> indicating whether the URL is internal or not?
It is all in components to generic test, but, yes, if no reuse can be reduced to specific, so, a boolean. 

> Also, the tests need to be updated.

You can delete this pull request and do yours, was only a suggestion, I have no way to test here.

Issue #1646, opened 2018-10-04
hightlight.js not loading with ReadTheDocs theme and http server


DDorch:
In the file  mkdocs/mkdocs/themes/readthedocs/base.html, the url used for loading hightlight.js are not complete: 

```html
{%- if config.theme.highlightjs %}
  <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script>
  {%- for lang in config.theme.hljs_languages %}
  <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/languages/{{lang}}.min.js"></script>
{%- endfor %}
```
The protocole is missing causing pages serves by http server to try loading to `http://cdnjs.cloudflare.com...` which is not correct.

The code should be changed to:
```html
{%- if config.theme.highlightjs %}
  <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js"></script>
  {%- for lang in config.theme.hljs_languages %}
  <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/languages/{{lang}}.min.js"></script>
{%- endfor %}
```

waylan:
I'm not seeing any problem. I've checked multiple browsers and they all make a `https` request. Which browser are you using and what error are you seeing (you may need to check your browsers dev tools for the error)?

waylan:
For the record I have tested this using the local dev server (`http://127.0.0.1:8000/`) on both Windows and Mac with Chrome, Firefox and Safari. I have tried both with and without local cache. I have never seen a failure. In fact, the request for `highlight.js` has always been an `https` request. I don't even see a redirect from `http` to https`.

geomunion:
I like this protocol-relative urls, but have a look at this:
https://stackoverflow.com/questions/28446314/why-use-protocol-relative-urls-at-all

If you visit your MkDocs site via HTTP, then assets from cdn are also requested via HTTP.
But because of HSTS (https://de.wikipedia.org/wiki/HTTP_Strict_Transport_Security) your Browser may choose HTTPS, if an asset was previously requested via HTTPS.

So - in this case - i prefer https:// urls

btw. : HTTP is still working - try:
`curl -vs "http://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/highlight.min.js" > /dev/null`

waylan:
[...]

PR #1653, opened 2018-10-12
Replace relative URL of CDNs by https mkdocs/mkdocs#1646

DDorch:


waylan:
The same changes should be made to the `mkdocs` theme.

DDorch:
Hi,
I've already checked that before the PR :)
The URLs are already in https in the MkDocs theme!

waylan:
So they are. I must have been looking at an older version or something.

Issue #1529, opened 2018-07-03
Use of yaml.load instead of yaml.safe_load


tonybaloney:
Hi,

I noticed that the yaml module is using load instead of safe_load. It's generally recommended to use safe_load as otherwise, attackers can run arbitrary code through configuration files.
See this OpenStack file 
https://security.openstack.org/guidelines/dg_avoid-dangerous-input-parsing-libraries.html

Happy to raise a PR and swap those with `yaml.safe_load` in https://github.com/mkdocs/mkdocs/blob/34ef3ca6d0390959080ce93a695361eea1649272/mkdocs/utils/__init__.py#L78

waylan:
Your reference states:

> You always want to use the safe functions to load input. 

I have to disagree with that blanket statement. Yes, many (most?) times that is true, But sometimes you might actually want/need the so-called "unsafe" option (otherwise, why would the library offer it). 

In any event, there are at least two questions we need to answer:

1. Is there a legitimate need for not using `safe_load` with MkDocs?

    For some users, yes, See [this comment][1] to #212 as an example. Although, as mentioned elsewhere in that discussion, we may not actually want to allow that. And it is likely possible to get the same functionality using a plugin rather than pointing to a Python object directly from the YAML config file.

    So, in the end, no there is no "need." In fact, we don't officially support doing that. But for some advanced users it is "nice to have."

2. As MkDocs is a ***static** site generator*, what danger, if any, really exists?

    Again this was discussed in #212 and we don't see it as a high concern. As a reminder, the "server" that we include with MkDocs is a dev server and not intended for use in a server open to the world. All our users ever upload to a server is a bunch of flat files. What exactly would the attack vector be here?

    Worst case, an attacker adds some code calling option into the config file and that code adds some evil JavaScript code to the generated site. But if the attacker can do that, then they can just add the JavaScript code into the source files for the site directly. No need to use the config file. At most, using the config file adds a level of indirection if the attacker is trying to hide their activities, but things are not actually any more or less safe than if we use `safe_load`.

    Yes, I realize that the attacker could simply use this as a method to get arbitrary code to run on the  the developer's machine (when she runs `mkdocs build`). But even if we did use `safe_load`, we still support plugins. So any plugin referenced in the config will call code which could be from an unknown source. How is that any different that not using `safe_load`?

[1]: https://github.com/mkdocs/mkdocs/issues/212#issuecomment-115404162

tonybaloney:
[...]

PR #1666, opened 2018-10-27
update pyyaml to fix CVE-2017-18342

vijaykramesh:
In PyYAML before 4.1, the yaml.load() API could execute arbitrary code. In other words, yaml.safe_load is not used.  See https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-18342

PyYAML released 4.1 but then pulled it from PyPi.  See https://github.com/yaml/pyyaml/issues/207 for some discussion.  https://github.com/yaml/pyyaml/tree/release/4.2 is the ongoing release branch, but nothing has been committed since July when 4.2b4 was released to PyPi

I have a project using my fork, with PyYaml updated to 4.2b4 and everything seems to be working fine. 
 Additionally the travis tests passed (and my project that includes mkdocs no longer triggers a vulnerability alert due to the unsafe version of PyYAML).

### DIFF
--- a/requirements/project-min.txt
+++ b/requirements/project-min.txt
@@ -2,6 +2,6 @@ click==3.3
 Jinja2==2.7.1
 livereload==2.5.1
 Markdown==2.5
-PyYAML==3.13
+PyYAML==4.2b4
 tornado==4.1
 mdx_gh_links>=0.2
--- a/requirements/project.txt
+++ b/requirements/project.txt
@@ -2,6 +2,6 @@ click>=3.3
 Jinja2>=2.7.1
 livereload>=2.5.1
 Markdown>=2.5
-PyYAML>=3.13
+PyYAML>=4.2b4
 tornado>=5.0
 mdx_gh_links>=0.2
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
         'livereload>=2.5.1',
         'lunr[languages]>=0.5.2',
         'Markdown>=2.3.1',
-        'PyYAML>=3.10',
+        'PyYAML>=4.2b4',
         'tornado>=5.0'
     ],
     python_requires='>=2.7.9,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',


facelessuser:
MkDocs uses the "unsafe" load on purpose: https://github.com/mkdocs/mkdocs/issues/1529. As MkDocs is used to pre-render the pages for a static site (you shouldn't be using this to dynmacially render pages on a live server), this isn't really a problem. See the provided link for more info.

I wonder though, as people bring this up from time to time, if it would make sense to just provide a setting to use the safe_loader to ease people's mind.  With that said, you could still load unsafe code with a malicious Python Markdown extension or a malicous MkDocs extension...yeah, I don't think it really makes that much of a difference either way...

If people have access to your shell to add in bad dynamic object loading in your yaml file, they have access to add bad plugins and all sorts of things.

vijaykramesh:
Thanks for the fast response!  I looked through #1529 (sorry I missed it before submitting this) and I have the same scenario as https://github.com/mkdocs/mkdocs/issues/1529#issuecomment-402304217 (CI runs mkdocs, and we run a dev-only docs server inside a VPC).  But as a part of the CI build we run some vulnerability detection so at least having a way to call `safe_load` from here will make that happier.  I'll close this for now and will report back if that works.  

It also sounds like the 4.2 release of pyyaml is stalled with a bunch of debate over this same change there (see https://github.com/yaml/pyyaml/issues/193, it's why they pulled 4.1) so probably not a good idea to hitch mkdocs to the current 4.2 release branch.

waylan:
@facelessuser covered this quite well. However I have a few comments:

> we run a dev-only docs server inside a VPC

This is not a supported use of the dev server and therefore it is not a valid reason to make any changes. If you want to run a server in a VPC (or anywhere except a single user dev box), then you should be using `mkdocs build` and serve the built site with your own third party server. In that scenario, the YAML parser is in no way connected to the server, which should reduce any security concerns.

Of course, that still leaves the `build` "insecure", but as previously stated, no more so than plugins do. And in fact, using `safe_load` would disable a feature: the ability to intentionally load Python code. In fact some users want that as a feature. For example, see Python-Markdown/markdown@6f463273 were the version of the documented library is retrieved and included in the documentation by the theme template. Some might suggest attempting to make it more secure by implementing our own scaled down variant which would only accept variables which point to strings. The problem is that anyone can create a subclass which includes code in the `__str__` method, so you can not really eliminate code execution unless you completely disable any ability to include anything.

> as a part of the CI build we run some vulnerability detection

Then you should be telling the vulnerability detection tool to skip the check for YAML safe loading, just like you can tell a linter to skip a rule, etc. This is a known, conscious decision made with the understanding of the concerns at issue. Namely, requiring such a restriction does not make MkDocs any more secure, but in fact, breaks supported features.

PR #1673, opened 2018-11-02
Account for encoded URLs.

### LINKED ISSUE #1670: Links with hex encoded characters do not work (specifically %20)
For the app Vespucci (see [vespucci.io](https://vespucci.io/)) we use mkdocs to create the website and a pegdown based system to use the same markdown files for on device  documentation.

For various reasons the file names of the individual documents contain spaces and it would be difficult to avoid this. Inter document links work with no issue on device by encoding the spaces as %20, however this does not work with mkdocs because the logic in https://github.com/mkdocs/mkdocs/blob/master/mkdocs/structure/pages.py line 217 does not decode the path before checking if the corresponding file exists. 

Naturally the same issue exists for other hex encoded characters and for using "+" as a space replacement. 

PS: yes, I know it works if one does not encode the spaces, however that is not "valid" markdown and causes issues with other markdown parsers, for example pegdown.

waylan:
Encode URLs of all internal documents (to account for spaces in 
filenames, etc). Then unencode Markdown links to internal pages before 
confirming existence, etc. Fixes #1670.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -56,6 +56,7 @@ your global navigation uses more than one level, things will likely be broken.
 
 ### Other Changes and Additions to Version 1.1
 
+* Bugfix: Account for encoded URLs (#1670).
 * Bugfix: Ensure theme files do not override `docs_dir` files (#1671).
 * Bugfix: Do not normalize URL fragments (#1655).
 * Add canonical tag to `readthedocs` theme (#1669).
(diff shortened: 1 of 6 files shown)


waylan:
I'm not 100% sure that this is the correct fix for PY27. I wonder if there may be some filenames with non-ASCII characters which this will break. My inclination is to suggest that users who *need* support for such filenames should be using PY3.

PR #1685, opened 2018-11-21
docs(plugins.md): outdated docs

zhaoterryy:
Docs are outdated: site_navigation -> files in almost all of the events.


waylan:
Thanks for this. We [updated some of the plugin docs][1] when that change was made. Not sure why we missed this.

[1]: https://github.com/mkdocs/mkdocs/pull/1504/files#diff-d65996b966b539b8298e7f90a78d1303

Issue #1696, opened 2018-12-06
Default yaml loader is unsafe


JulesDT:
Hello !

I think there is an issue in the yaml_loader method in `/utils/__init__.py`. The default loader used to load templates is unsafe and might lead to an RCE.

https://github.com/mkdocs/mkdocs/blob/cc03fd05876ad7ff14dfeff55832dc0dd7f0bf06/mkdocs/utils/__init__.py#L62

This should rather be the SafeLoader by default to avoid someone loading a malicious template to execute arbitrary code.

JulesDT:
I made a PR to fix this issue (https://github.com/mkdocs/mkdocs/pull/1697), as well as a small PoC of this vulnerability available here: https://github.com/JulesDT/mkdocs-theme-rce-poc

waylan:
This is a duplicate of #1529

PR #1697, opened 2018-12-06
use SafeLoader instead of the default Pyyaml loader

JulesDT:
This PR is intended to fix https://github.com/mkdocs/mkdocs/issues/1696


waylan:
This is rejected for the reasons stated in #1529.

Issue #1152, opened 2017-03-08
site_dir contains folders (and subfolders) from doc_dir though not specified in mkdocs.yml


Tset-Noitamotua:
After doing `mkdocs build` I expected to find in the site_dir only the rendered .html files in according folder structure as defined by `pages:` in **mkdocs.yml** + some mkdocs specific assets. Instead I find there every file and folder/subfolder from docs_dir.

Am I doing something wrong?

I am building test documentation for several projects which are all in my local workspace. So I'm kind of using mkdocs not as supposed to be used to build doculmentation per single project but instead for test docu only but for multiple projects at once. But the only relevant raw `.md` files are in `project_folder/src/test/robotframework/TESTS` (subsubsubfolder of each project)
e.g. `ugly_app/src/test/robotframework/TESTS/test_doc.md`. 

My docs_dir tree looks akin like:
```shell
│   ├───ugly_app
│   │   └───src
│   │       ├───main
│   │       │   ├───java
│   │       │   │   └───com
│   │       │   │       └───mycompany
│   │       │   │           ├───...
│   │       │   │           │   └───...
│   │       │   │           │       └───...
│   │       │   │           └───....
│   │       │   │               └───...
│   │       │   │                   ├───...
│   │       │   │                   ├───...
│   │       │   │                   │   ├───...
│   │       │   │                   │   │  
│   │       │   │                   │   └───..
│   │       │   │                   ├───...
│   │       │   │                   ├───...
│   │       │   │                   │   └───...
│   │       │   │                   │       └───...
│   │       │   ├───...             ...
│   │       │   │   ├───...
│   │       │   │   │   └───..
│   │       │   │   │       └───...
│   │       │   │   ├───...
│   │       │   │   ├───...
│   │       │   │   └───.
│   │       │   └───...
│   │       │       ├───..
│   │       │       └───.
│   │       └───test
│   │           ├───java
│   │           │   └───com
│   │           │       └───mycompany
│   │           │           └───...
│   │           │               └──...
│   │           │                   ├───...
|   |           |                   ...
│   │           └───robotframework
│   │               ├───...
│   │               ├───...
│   │               │   └───...
│   │               ├───TESTS
│   │               ...
│   │               
│   ├───SCArchive_20111111Test
│   │   ├───robotframework
│   │   │   ├───...
│   │   │   ├───...
│   │   │   └───TESTS
│   │   │       └───...
│   │   └───...
... ...
```

[...]

PR #1699, opened 2018-12-11
Remove template exclusion

soffes:
To quote @waylan from #1152:

> There is no "exclude" of any kind. It was never anticipated that non-documentation would be in the docs_dir... And quite frankly, why would anyone actually want that?

In my case, I have a directory of documentation called `templates` that I'd really like to add. From what I can tell, a `templates` directory inside of `docs_dir` isn't referenced anywhere. Is this safe to remove?

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -60,6 +60,7 @@ your global navigation uses more than one level, things will likely be broken.
 * Bugfix: Ensure theme files do not override `docs_dir` files (#1671).
 * Bugfix: Do not normalize URL fragments (#1655).
 * Add canonical tag to `readthedocs` theme (#1669).
+* No longer ignore a directory named `templates` inside of `docs_dir`. (#1699)
 
 ## Version 1.0.4 (2018-09-07)
 
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -133,7 +133,7 @@ def _build_extra_template(template_name, files, config, nav):
 
     log.debug("Building extra template: {}".format(template_name))
 
-    file = files.get_file_from_path(template_name)
+    file = files.get_extra_template_file_from_path(template_name)
     if file is None:
         log.warn("Template skipped: '{}' not found in docs_dir.".format(template_name))
         return
--- a/mkdocs/structure/files.py
+++ b/mkdocs/structure/files.py
@@ -218,7 +218,7 @@ class File(object):
 def get_files(config):
     """ Walk the `docs_dir` and return a Files collection. """
     files = []
-    exclude = ['.*', '/templates']
+    exclude = ['.*']
 
     for source_dir, dirnames, filenames in os.walk(config['docs_dir'], followlinks=True):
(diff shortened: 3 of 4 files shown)


waylan:
Hmm, it appears that first appeared in ad72f4c88, which has since been quashed with various other commits before being merged into master. I'm not sure why @tomchristie felt the need to add that and there's no clues in the commit message. AFAIK we don't do anything special with any `templates` dir within the `docs_dir`. If a user wants to make custom templates available, they should be using `theme.custom_dir` to point to a dir. However, the expectation is that that dir would not be in `docs_dir`.

Given the above, my inclination would be to accept this. However, if doing so breaks existing users sites, that would be a regression, which gives me pause. Additionally, we have at least [one test][1] which explicitly tests this behavior, which suggests that the behavior is intentional. I added that test, but don't specifically recall why I thought it necessary and, again, no helpful commit messages exist.

In any event, at the least, the [failing test][2] would need to be updated before this could be accepted. If you update the test, I'll take another look, but I haven't made a decision one way or the other yet.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/structure/file_tests.py#L560-L576
[2]: https://travis-ci.org/mkdocs/mkdocs/jobs/466605323#L504

waylan:
I figured out why the `templates` directory is excluded. The [extra_templates][1] config setting can point to templates within the `docs_dir`. I believe the idea was that one could avoid raw `extra_templates` from being copied to the site by keeping them in a `templates` directory.

That said, the exclusion of a `templates` dir is not documented anywhere. And it is reasonable to expect that users may want to have a publicly available `templates` dir. Perhaps the directory should be renamed. Or, users could always just use a dot-prefixed name (ex. `.templates/`) which gets excluded anyway.

The one concern I have is that there may be some existing users who are depending on the existing behavior. Any change will cause their "hidden" templates to suddenly become "unhidden." At the very least we would need to add a comment to the release notes warning about the change.

[1]: https://www.mkdocs.org/user-guide/configuration/#extra_templates

waylan:
On further reflection, I think perhaps we should exclude all files listed in `extra_templates`. That would provide a backward compatible solution as any existing users won't have their templates suddenly appearing and the `templates` dir has no special meaning. It also means that users can use whatever names and/or directories they want to store their `extra_templates`. 

One concern is if a user has a template which is inherited from another, the inherited template would not be listed in `extra_templates`. But a rename to be dot-prefixed would address that concern.

soffes:
@waylan do you (or another maintainer/volunteer) have time to make that change? I'm pretty out of my element here. I'm extremely new to Python.

waylan:
@soffes I'll take a look at this when I have some time.

soffes:
I'm going to close this since I don't have the knowledge or time to continue pursuing it.

Feel free to branch off of this and continue if you'd like this change. Thanks again for your time reviewing it!

waylan:
If this is closed, I'll forget that it exists when I have time to look at it. And I can't reopen it as you apparently deleted your branch which this was based on.

PR #1703, opened 2018-12-17
Overhaul events

orlnub123:
The current event implementation is a bit limited. The idea of having events revolve around an item is cool; you pass an item to run_event, it loops through all the available event handlers, calls them with the item and if they return anything it reassigns the item to the return value, finally returning the item's final iteration, possibly completely different from the one passed in. Unfortunately this doesn't accommodate events that don't revolve around an item, which are only used as callbacks, such as pre_build and post_build, or even events that are for creating a new item, such as page_read_source.

I've changed `run_event` by making `item` optional (defaulting to `None`). If an item isn't provided it doesn't get passed to the event handlers. This doesn't mean that the event handlers can't make their own items to be returned, the items they create just don't get passed to the next handlers. This allows both callback-based and creation-based events to coincide.

### DIFF
--- a/mkdocs/tests/plugin_tests.py
+++ b/mkdocs/tests/plugin_tests.py
@@ -27,6 +27,14 @@ class DummyPlugin(plugins.BasePlugin):
         """ do nothing (return None) to not modify item. """
         return None
 
+    def on_page_read_source(self, **kwargs):
+        """ create new source by prepending `foo` config value to 'source'. """
+        return '{} {}'.format(self.config['foo'], 'source')
+
+    def on_pre_build(self, **kwargs):
+        """ do nothing (return None). """
+        return None
+
 
 class TestPluginClass(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


mkdocs/tests/plugin_tests.py
```diff
@@ -20,13 +20,21 @@ class DummyPlugin(plugins.BasePlugin):
     )
 
     def on_pre_page(self, content, **kwargs):
```
orlnub123: This is unrelated to the PR but would it be OK to change this event to on_page_content to more closely match its docstring?

waylan:
Interesting. Thank you for your contribution. 

Could you provide a use case which wasn't possible before this change? AFAICT, all this does is formalize the behavior we already had. It doesn't actually add any new ability.

Also, are any updates to the documentation required? If so, they would need to be included before this would get accepted.

PR #1707, opened 2018-12-18
Update installation requirements

James-Quigley:
Add more specific installation requirements when installing manually via python and pip for version 1.0.0+

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -67,8 +67,8 @@ Python 2.7.2
 $ pip --version
 pip 1.5.2
 ```
-
-MkDocs supports Python versions 2.7, 3.4, 3.5, 3.6, 3.7 and pypy.
+MkDocs v1.0.0+ requires Python versions 2.7.9 or 3.4+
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -67,8 +67,8 @@ Python 2.7.2
 $ pip --version
 pip 1.5.2
 ```
-
-MkDocs supports Python versions 2.7, 3.4, 3.5, 3.6, 3.7 and pypy.
+MkDocs v1.0.0+ requires Python versions 2.7.9 or 3.4+
+MkDocs <v1.0.0 supports Python versions 2.7, 3.4, 3.5, 3.6, 3.7 and pypy.
```
waylan: I have multiple issues with this change:

1. `<v1.0.0` is not inclusive. It would be more correct to use `<=`. So let's just leave the `+`.
2. `2.7.9` is not strictly correct. `>=2.7.9 & <3.0` is strictly correct, but not really appropriate for prose. Perhaps `2.7.9+`? The trick is making it clear the `>` only applies to the "patch" number. The fact that any legible solution leaves ambiguity is probably why the problem exists int he first place.
3. `pypy` should not be removed from the list. The other concerns may be debatable, this one is not.

PR #1726, opened 2019-01-19
feat(pages): allow metadata edit_uri & repo_url to overide edit_url

paulrbr-fl:
In the context of merging multiple source code repositories to build a single documentation, I find it useful to be able to override the global `edit_uri` and `repo_url` config for each pages.

WDYT?


waylan:
Interesting. I hadn't considered anyone wanting to do that. As a general rule, it seems like a terrible usability issue. Changing the value hundreds of times for hundreds of pages is not something most people would want to do.

My suggestion would be to address this at the template level. Something like:

```django
{% is page.meta.edit_url %}{{ page.meta.edit_url }}{% else %}{{ page.edit_url }}{% endif %}
```

Note that I used `edit_url` not `edit_uri`. There is no autobuilding of the URL. And as each page is defined separately, there is no need for that. The actual URL for a given page can be included in the metadata.

And with [theme customization][1], you can simply [override the `repo` block][3] (see the [`mkdocs` theme][2] for example) with no changes necessary to the MkDocs.

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/themes/mkdocs/base.html#L137-L167
[2]: https://www.mkdocs.org/user-guide/styling-your-docs/#using-the-theme-custom_dir
[3]: https://www.mkdocs.org/user-guide/styling-your-docs/#overriding-template-blocks

waylan:
Another solution might be to use a [plugin][1] to alter the `edit_url` defined on each `page` object. Any of the `on_page_markdown`, `on_page_content` or `on_page_context` events should work to alter the `page` object by reassigning the `page_edit_url` to the value of `page.meta.edit_url`.

[1]: https://www.mkdocs.org/user-guide/plugins/

paulrbr-fl:
Thanks for the feedback @waylan. Indeed a plugin seems like a better way to manage this.

psamim:
@paulrbr-fl Hi, How did you solve this eventually? Did you write a plugin for it?

paulrbr-fl:
> @paulrbr-fl Hi, How did you solve this eventually? Did you write a plugin for it?

Hi there @psamim. I finally wrote a 3 line of code plugin: https://gitlab.com/paulrbr/mkdocs-edit-url/

It has worked very well until now!

waylan:
@paulrbr-fl feel free to add your extension to the [wiki][1] so others can find it.

[1]: https://github.com/mkdocs/mkdocs/wiki/MkDocs-Plugins

PR #1739, opened 2019-02-06
Fixed default link in new page

philippec-ls:
Just using "https://mkdocs.org" yields a certificate error because that is the GitHub wildcard certificate, so the domains don't match.


waylan:
I thought we changed all of those some time ago. Guess we missed a few. Thanks.

waylan:
The failing test was not introduced here, so I'm ignoring it for this PR. Presumably an update to flake8 is causing it to catch something it missed before...

PR #1756, opened 2019-02-21
Add a second level of 'strict'.

waylan:
This is an experimental solution to #1755 to see if it makes sense. The plan is to add a new logging level between `info` and `warning` named `notice`. Then `--strict --strict` would increase the severity and the build would exit/fail if any `notice` messages were logged. A single `--strict` would  maintain its current behavior and only exit/fail on `warning`, but not `notice`.

So far I have only refactored the log counter to count all levels, not just warnings.

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -30,7 +30,7 @@ class DuplicateFilter(object):
 
 log = logging.getLogger(__name__)
 log.addFilter(DuplicateFilter())
-log.addFilter(utils.warning_filter)
+log.addFilter(utils.log_counter)
 
 
 def get_context(nav, files, config, page=None, base_url=''):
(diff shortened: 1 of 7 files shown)


waylan:
While working on the [tests][1] for this, I was reminded that `log.setLevel` can be used to silence lower level messages. This could theoretically be used with a single counter (which counted all levels that are not blocked as a single value). However, we also need to support the `--verbose` flag which may have different requirements.

That said, it makes me wonder if we should instead use a second `logging.Handler` instance (in addition to the default) which only counts messages and set the level on that based on `strict`.

[1]: https://github.com/mkdocs/mkdocs/pull/1756/files#diff-9932f65fd1f8ba8b9753f321713f15beR542

waylan:
Closing this in favor of #2364.

Issue #1757, opened 2019-02-23
Image paths can break when deploying to gh-pages


devakker:
1. Let's say you want to store your images in an `img` folder inside the root of your docs
2. Embed an image like this: `![](/img/image.png)`
3. test in a local environment -> works fine
4. deploy to gh-pages -> missing image

The generated HTML will link to the image like this: `<p><img alt="" src="/img/image.png" /></p>`

But the image will actually be at `https://username.github.io/repo-name/img/image.png` and it won't show up.

This entire issue can be avoided (and most likely is by a lot of people) by using relative paths. Still worth considering if maybe the repo's name should automatically added before the image URL, if it's an absolute path. 

config: mkdocs 1.0.04, windows10

waylan:
> This entire issue can be avoided ... by using relative paths.

For consistent results the [documentation][1] actually points out that relative paths are required. You may or may not get things to work correctly with absolute paths and use them at your own risk. In fact, they only work if the mkDocs root and server root are the same directory. The reason for this is that relative paths are adjusted by MkDocs to ensure they are always relative to the page, However, absolute paths are not modified at all. In other words, absolute paths do not get adjusted based on the server environment.

We do not adjust absolute URLs as some users actually want/need that behavior. For example, some users may want to link to something hosted on the same sever, but outside of MkDocs. Suppose the MkDocs site was at `http://example.com/docs/` but they wanted to link to `http://example.com/foo/bar/`. They could simply include a link to `/foo/bar` and it would work correctly. However, if we treated the link as an internal link, we would either break the link or raise an error when trying to verify it, neither if which is desired. Therefor all internal links **must** be relative links to be verified and auto-adjusted as necessary.

Finally, note that if you are in an environment where the MkDocs root directory is a subdirectory of the server root, then you should set the [site_url][2] config setting to include the  subdirectory. For example:

```yaml
site_url: https://username.github.io/repo-name/
```

The local dev server will then use that subdirectory, which will help to ensure consistent behavior between the local dev server and your production server.

In conclusion, the problem is user error, not a bug. Two things should be done to correct the problem:

1. For all internal links, use relative URLs: `![](img/image.png)`
2. Properly define `site_url` to include the subdirectory so that it points at the MkDocs' root.

[1]: https://www.mkdocs.org/user-guide/writing-your-docs/#internal-links
[2]: https://www.mkdocs.org/user-guide/configuration/#site_url

devakker:
[...]

PR #1759, opened 2019-02-25
Warning about the usage of absolute paths in links

devakker:
Regarding #1757. An update to the docs to warn about the dangers of absolute paths in links.

I am not sure if you guys manually break lines or there is some tool to do it, so I just left them as is for now. I know it's not going to be merged like this, but I couldn't find anything on this in the contributing guidelines. Let me know what's the official way to go about this an I'll fix it.

### DIFF
--- a/docs/user-guide/writing-your-docs.md
+++ b/docs/user-guide/writing-your-docs.md
@@ -201,6 +201,9 @@ Please see the [project license](license.md) for further details.
 When the MkDocs build runs, these Markdown links will automatically be
 transformed into an HTML hyperlink to the appropriate HTML page.
 
+!!! warning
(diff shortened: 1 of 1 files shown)


waylan:
Most text editors have  "hard wrap" feature which will add the line beaks for you. Of course, the actual name differs but that should give you something to search for.

pferreir:
Maybe there could be as well a compile-time warning? This seems to be a common pitfall and I have myself lost time wondering what was wrong with my site only to realize that absolute paths fail randomly. I would be willing to help out if you think this is the best approach.

waylan:
@pferreir as explained [here][1], there are legitimate uses for absolute paths and there is no way for us to know which was intended. Therefore a warning/error would be inappropriate.

[1]: https://github.com/mkdocs/mkdocs/issues/1757#issuecomment-466698398

PR #1760, opened 2019-02-26
Added processing time debug printout.

karlssonjohan:
Some pages take long time to process so added processing time printout to aid in troubleshooting.


waylan:
This seems like a great candidate for a plugin.

PR #1771, opened 2019-03-23
Add: openbrowser as serve command's option to open browser automatically

peaceiris:
PR for #1772 

This pull request is ready for review. Could you confirm this?

How can I pass the CI checks?

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import logging
 import shutil
 import tempfile
+import webbrowser
 
 from os.path import isfile, join
 from mkdocs.commands.build import build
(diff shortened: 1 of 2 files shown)


mkdocs/commands/serve.py
```diff
@@ -120,6 +121,9 @@ def builder():
 
         host, port = config['dev_addr']
 
+        if openbrowser:
+            webbrowser.open(config['site_url'])
```
waylan: I have two concerns about this:

1. Why `site_url` and not `dev_addr`? I realize that when the dev server is used, we alter the `site_url` to match `dev_addr`, but is something were ever to go wrong with that, this would break things. Seems like we should just use `dev_addr` here.

2. Should't the browser be launched after the server is started, not before?

peaceiris:
This is a prototype to inquire about interest in this feature at #1772 

I got the feedback and understood that this feature is not necessary. If this was acceptable, we need a huge cost to manage, but no benefits.

I drop this PR not to waste time.

Thanks.

PR #1774, opened 2019-03-25
contributing.md: fix typos, add link

hfagerlund:
* contributing.md:52-53

"needs
to be pip installed ..." 
-> "needs
[pip](https://pypi.org/project/pip/) to be installed ..."

* contributing.md:53

"then the test suite can be ran for MkDocs but running ..." -> "then the test suite can be run for MkDocs by running  ..."

### DIFF
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -49,10 +49,16 @@ to the git repository.
 
 ## Running the tests
 
-To run the tests, it is recommended that you use [Tox]. This just needs
-[pip](https://pypi.org/project/pip/) to be installed and then the test suite
-can be run for MkDocs by running the command `tox` in the root of your MkDocs
-repository.
+To run the tests, it is recommended that you use
+[Tox](https://pypi.org/project/tox/).
+
+Install Tox using [pip](https://pypi.org/project/pip/) by running the
+following command:
+```
+pip install tox
+```
+Then the test suite can be run for MkDocs by running the command `tox` in the
+root of your MkDocs repository.
 
 It will attempt to run the tests against all of the Python versions we
 support. So don't be concerned if you are missing some and they fail. The rest
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -54,9 +54,11 @@ To run the tests, it is recommended that you use
 
 Install Tox using [pip](https://pypi.org/project/pip/) by running the
 following command:
+
 ```
 pip install tox
 ```
+
 Then the test suite can be run for MkDocs by running the command `tox` in the
 root of your MkDocs repository.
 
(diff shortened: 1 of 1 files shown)


docs/about/contributing.md
```diff
@@ -50,8 +50,9 @@ to the git repository.
 ## Running the tests
 
 To run the tests, it is recommended that you use [Tox]. This just needs
-to be pip installed and then the test suite can be ran for MkDocs but running
-the command `tox` in the root of your MkDocs repository.
+[pip](https://pypi.org/project/pip/) to be installed and then the test suite
```
waylan: "needs to be pip installed" is not an error. In other words, "`tox` needs to be installed using pip." Or "run the command `pip install tox`."

I suppose this could be worded more clearly, but this change is not correct as it alters the intended meaning.

hfagerlund: My mistake (I misread that). Fixed in 8fe2c33.

docs/about/contributing.md
```diff
@@ -49,9 +49,16 @@ to the git repository.
...
 
-To run the tests, it is recommended that you use [Tox]. This just needs
-to be pip installed and then the test suite can be ran for MkDocs but running
-the command `tox` in the root of your MkDocs repository.
+To run the tests, it is recommended that you use
+[Tox](https://pypi.org/project/tox/).
+
+Install Tox using [pip](https://pypi.org/project/pip/) by running the
+following command:
+```
```
waylan: As reported by the [markdownlint][1] test, the fenced code block should be surrounded by blank lines and have a language defined (use 'bash')

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/511133946#L510

docs/about/contributing.md
```diff
@@ -49,9 +49,16 @@ to the git repository.
 
 ## Running the tests
 
-To run the tests, it is recommended that you use [Tox]. This just needs
-to be pip installed and then the test suite can be ran for MkDocs but running
-the command `tox` in the root of your MkDocs repository.
+To run the tests, it is recommended that you use
+[Tox](https://pypi.org/project/tox/).
+
+Install Tox using [pip](https://pypi.org/project/pip/) by running the
```
waylan: Please follow the pattern for links used elsewhere in the document. All links are reference style links with the URLs grouped together at the bottom of the document. In fact, there is already a link for `tox` there. No need to repeat the URL again. Also, why are we using pypi links? The projects themselves don't advertise those links. Lets use the links which the projects themselves use.

hfagerlund: Fixed in ccdd882 (updated link to `pip`).

waylan:
@hfagerlund thanks for your contribution and for working through the issues with me.

hfagerlund:
Thanks @waylan for all the pointers (good catches) and your patience.

PR #1778, opened 2019-04-10
Add theme localization

anlau:
_&lt;edit&gt;_

### Note from project maintainers:

There are multiple aspects to supporting multiple languages in MkDocs, each of which is linked to from #211, which is the primary issue and ties all of the various aspects together. However, it is best to discuss each aspect in its related issue. 

This issue/PR addresses theme localization (l10n). The templates need to be updated to support translations of the next/previous navigation, page footer, etc.  Volunteers are needed to complete the work as outlined [below](#issuecomment-559211318).

_&lt;/edit&gt;_

Goal: be able to use translations from PyBabel to localize a theme when built.

Before moving forward, I would like some feedback. There's a bit more work to do (mostly add tests). Second commit will be removed/or bonified (note: .mo files need to be generated with `pybabel compile -d locales`). I think translations could be done in a separate PR.

Related with [this comment](https://github.com/mkdocs/mkdocs/issues/211#issuecomment-466420384).

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -415,6 +415,7 @@ class Theme(BaseConfigOption):
 
     Validate that the theme exists and build Theme instance.
     """
+    name = 'theme'
 
     def __init__(self, default=None):
         super(Theme, self).__init__()
@@ -447,7 +448,8 @@ class Theme(BaseConfigOption):
         theme_config = config[key_name]
 
         if not theme_config['name'] and 'custom_dir' not in theme_config:
-            raise ValidationError("At least one of 'theme.name' or 'theme.custom_dir' must be defined.")
+            raise ValidationError("At least one of '{name}.name' or '{name}.custom_dir' must be defined."
+                                  .format(name=self.name))
 
         # Ensure custom_dir is an absolute path
         if 'custom_dir' in theme_config and not os.path.isabs(theme_config['custom_dir']):
@@ -455,8 +457,11 @@ class Theme(BaseConfigOption):
             theme_config['custom_dir'] = os.path.join(config_dir, theme_config['custom_dir'])
 
         if 'custom_dir' in theme_config and not os.path.isdir(theme_config['custom_dir']):
-            raise ValidationError("The path set in {name}.custom_dir ('{path}') does not exist.".
-                                  format(path=theme_config['custom_dir'], name=self.name))
+            raise ValidationError("The path set in '{name}.custom_dir' ('{path}') does not exist."
+                                  .format(path=theme_config['custom_dir'], name=self.name))
+
+        if 'locale' in theme_config and not isinstance(theme_config['locale'], utils.string_types):
+            raise ValidationError("'{name}.locale' must be a string.".format(name=self.name))
 
         config[key_name] = theme.Theme(**theme_config)
 
--- a/mkdocs/structure/files.py
+++ b/mkdocs/structure/files.py
@@ -67,6 +67,8 @@ class Files(object):
         """ Retrieve static files from Jinja environment and add to collection. """
         def filter(name):
             patterns = ['.*', '*.py', '*.pyc', '*.html', '*readme*', 'mkdocs_theme.yml']
+            # Exclude translation files
+            patterns.append('locales/*')
             patterns.extend('*{0}'.format(x) for x in utils.markdown_extensions)
             patterns.extend(config['theme'].static_templates)
             for pattern in patterns:
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,5 @@
 include README.md
 include LICENSE
-recursive-include mkdocs *.ico *.js *.css *.png *.html *.eot *.svg *.ttf *.woff *.woff2 *.xml *.mustache *mkdocs_theme.yml
+recursive-include mkdocs *.ico *.js *.css *.png *.html *.eot *.svg *.ttf *.woff *.woff2 *.xml *.mustache *mkdocs_theme.yml *.mo
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -14,6 +14,7 @@ import jinja2
 from mkdocs import utils
 from mkdocs.structure.files import get_files
 from mkdocs.structure.nav import get_navigation
+from mkdocs.localization import install_translations
 import mkdocs
 
 
@@ -262,6 +263,9 @@ def build(config, live_server=False, dirty=False):
     files = get_files(config)
     env = config['theme'].get_env()
     files.add_files_from_theme(env, config)
+
+    if 'locale' in config['theme']:
+        install_translations(env, config)
 
     # Run `files` plugin events.
     files = config['plugins'].run_event('files', files, config=config)
--- a/mkdocs/contrib/search/templates/search/main.js
+++ b/mkdocs/contrib/search/templates/search/main.js
@@ -37,7 +37,11 @@ function displayResults (results) {
       search_results.insertAdjacentHTML('beforeend', html);
     }
   } else {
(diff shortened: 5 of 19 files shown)


waylan:
It's great to see someone working on this. However, I think theme localization could probably be either a built feature or a third party plugin. 

I don't intend to add any more builtin extensions. The only reason search is a plugin is because it used to be a builtin feature which some people did not want to use. So we broke it out into a plugin, but included it by default to maintain backward compatibility while providing users a mechanism to disable it. l10n is very different. First of all, it has not yet been supported so we break nothing by not including it by default. However, I am concerned that if the templates contain translation strings (such as `{{ _('nav.previous') }}`), then we will always need l10n. In that case, it should be a built-in feature rather than a plugin.

To be clear, this should only cover translations of the templates themselves, not translation of any page content. Translation of page content should only be provided by a third party plugin.

anlau:
Thanks for your feedback!

I did pretty much the same but built-in. See last update.

mkdocs/structure/files.py
```diff
@@ -67,6 +67,8 @@ def add_files_from_theme(self, env, config):
         """ Retrieve static files from Jinja environment and add to collection. """
         def filter(name):
             patterns = ['.*', '*.py', '*.pyc', '*.html', '*readme*', 'mkdocs_theme.yml']
+            # Exclude translation files
+            patterns.extend(['*.pot', '*.po', '*.mo'])
```
waylan: I wonder if it makes more sense to exclude the `locales` dir instead: `locales/*`. 

And while I'm thinking of it, these file types need to be added to `MANIFEST.in` so they get included in the package.

waylan:
For reference purposes, this PR is addressing step 1 of [this comment][1] on #211.

Also, so that we don't forget about it, it might make sense for `plugings.search.lang` to default to `theme.locale` if it is set.

[1]: https://github.com/mkdocs/mkdocs/issues/211#issuecomment-466420384

mkdocs/config/config_options.py
```diff
@@ -458,6 +458,9 @@ def post_validation(self, config, key_name):
             raise ValidationError("The path set in {name}.custom_dir ('{path}') does not exist.".
                                   format(path=theme_config['custom_dir'], name=self.name))
 
+        if 'locale' in theme_config and not isinstance(theme_config['locale'], utils.string_types):
+            raise ValidationError("'theme.locale' must be a string.")
```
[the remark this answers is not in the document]
anlau: It doesn't work, even for the line you mentioned. It raises an AttributError (it seems to be used correctly with `FilesystemObject` and `Dir`).

mkdocs/structure/files.py
anlau: On the same subject: which files should I leave behind? `.pot` and `.mo` files are ignored by git.

What I did to generate `.mo`:
1. Create a babel config file with content:
    ```
    [jinja2: **.html]
    extensions=jinja2.ext.i18n
    ```
2. `pybabel extract -F locales/babel.cfg -o locales/messages.pot .`
3. `pybabel init -d locales -l en -i locales/messages.pot`
4. `pybabel compile -d locales`

anlau:
Beside documentation and your point about `plugings.search.lang`, there's a small snippet of English text in `mkdocs/contrib/search/templates/search/main.js` ("No results found") I'm not sure how to localize. Should it be moved in a html file passed through Jinja?

mkdocs/config/config_options.py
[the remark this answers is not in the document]
waylan: Then this is a bug that needs fixed. Definitely do not leave `theme`.

waylan:
> there's a small snippet of English text in `mkdocs/contrib/search/templates/search/main.js` ("No results found") I'm not sure how to localize. Should it be moved in a html file passed through Jinja?

I suppose that should be defined in the template. Then, for backward compatibility, the script should only redefine it if it is not defined in the template.

mkdocs/structure/files.py
waylan: The `.pot` and `.mo` files are ignored by git because they are listed in the [.gitignore][1] file. The list of files there is just the standard list for Python projects. Personally I'm not familiar enough with the various translation files types to know which should be ignored and which shouldn't. But given these are in the list of standard file types to ignore, my assumption is that they should be. Presumably they would be generated files? We likely need to add a command (or few) to the tox file to generate the translations before running the tests.

[1]: https://github.com/mkdocs/mkdocs/blob/master/.gitignore#L48-L50

mkdocs/config/config_options.py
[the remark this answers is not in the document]
anlau: I did change for `theme` but do you think `key_name` would be more appropriated?

mkdocs/structure/files.py
anlau: That's what I meant by ignored! I'm not that much familiar with translation files either but:

- `.pot` file is the output of 2 and probably can be generated from a `.po` file
- `.po` files are the output of 3 (they are then manually modified and can be updated with `update` command, see [here](http://babel.pocoo.org/en/latest/cmdline.html))
- `.mo` files are the output of 4

I modified the failing test so it doesn't depend on a `.mo` file. Note: `.mo` files would have to be generated before packaging (manually or preferably automatically).

waylan: Any files which need generated before packaging should also be generated before testing. Perhaps [here][1]?

When I have some more time, I'm going to need to take a closer look at this to figure out how to proceed.

[1]; https://github.com/mkdocs/mkdocs/blob/master/tox.ini#L15

waylan:
I finally had an opportunity to take a closer look at this. Looks good. However, the work is not done yet. At a minimum we need to do the following work:

1. [ ] Update the `readthedocs` theme to support translations (as has already been done with the `mkdocs` theme).
2. [ ] List `babel` in the requirements files (in addition to setup.py).
3. [ ] Document everything. We need to document things for users, theme devs, and translators; presumably separately.
4. [ ] Work out the development workflow and possibly develop scripts/sub-commands to automate it.

On the last point, it appears that the following pieces are required to maintaining translations:

1. Each theme should have a PO template, which is the "extracted" file `babel` generates when it scans the source. Basically it is a PO file with the translations all blank (this file is missing from the PR at this time). Any time a theme dev updates a theme in such a way that the translatable strings are changed, the PO template needs to be updated. We should probably provide a script/subcommand for theme devs to use for this. I'm not sure where this PO template should be stored. Maybe under `locale` in the theme? The PO Template file would never be included in a release, but would remain in the repo.

2. Each theme should also have a collection of translations, one per supported locale. Each translation is composed of the following pieces:

    1. A PO file, which is where the translations are maintained. To start a new locale, a translator would copy the PO template to a new locale subdirectory and then define the translations in the new PO file. After an update to the PO Template (by the theme dev as described above), a translator would diff the PO template against the PO file to determine which translations need to be added/updated and then make those updates/additions. We might want to automate both of those. Or explore third-party services that automate this for us.

    2. A MO file is a compiled PO file and is used by MkDocs to build the site. There are multiple possible approaches here.

        a. We could require the translator to (re)generate the MO file before pushing up a PR.
        b. We could automate (re)generation of a MO file when a PO file is committed/merged.
        c. The MkDocs' release manager could (re)generate MO files when preparing a release.
        d. We could generate a MO file on each `build` (and test run?) of MkDocs (current behavior of this PR, I think).

        In the first three of those options (a, b, c), the PO files would not be included in the release as we would have pregenerated MO files which would be included. However in the last option (d), we would include the PO files in the release, but not any MO files as they would be generated on the fly. Given the small size of the PO files, I'm not terribly worried about saving time by pregenerating the MO files. I think the solution we go with will depend on which is easier to implement/maintain.

TheOneWithTheBraid:
@anlau Are you still working on this?

anlau:
Not at the moment.

waylan:
Closing this as it was completed in #2299.

PR #1782, opened 2019-04-17
Update Travis config for default Python 2.7=>3.6

waylan:
Travis has changed their default Python version from 2.7 to 3.6 on April 16, 2019 in anticipation of PY27 support being dropped in 2020. The `basepython` setting isn't nessecary as the tests will run by default in whichever version tox is called from. As we can be sure that version exists on the platform, it is best to use that version and avoids the need for future changes.

This avoids test failures like the [this](https://travis-ci.org/mkdocs/mkdocs/jobs/521064818) and [this](https://travis-ci.org/mkdocs/mkdocs/jobs/521064816).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ matrix:
     - env: TOXENV=markdown-lint
       before_install: npm install -g markdownlint-cli
     - env: TOXENV=linkchecker
+      python: '2.7'
     - env: TOXENV=jshint
       before_install: npm install -g jshint
     - env: TOXENV=csslint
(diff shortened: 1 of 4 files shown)


waylan:
In relation to LinkChecker, note that by specifying `python2.7` in the Travis config, we get Python 2.7.14, but if we don't specify the Python version in the Travis config, `basepython = python2.7` in the tox config results in 2.7.6, which fails as we only support 2.7.9+. But we need to keep the tox config so we can do local tox runs. As long as we specific 2.7 in both config files then everything works. 

Can't wait for all Python 2 support to be a thing of the past.

Issue #722, opened 2015-09-08
Error while executing gh-deploy


ThomasArdal:
I've successfully deployed a MkDocs site using the gh-deploy command. When I try to deploy some additional changes to my master branch, I get the following error:

```
c:\docs>mkdocs gh-deploy --clean
INFO    -  Cleaning site directory
INFO    -  Building documentation to directory: c:\docs\site
INFO    -  Copying 'c:\docs\site' to 'gh-pages' branch and pushing to GitHub.
Traceback (most recent call last):
  File "C:\Python34\lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python34\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "c:\Python34\Scripts\mkdocs.exe\__main__.py", line 9, in <module>
  File "C:\Python34\lib\site-packages\click\core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "C:\Python34\lib\site-packages\click\core.py", line 644, in main
    rv = self.invoke(ctx)
  File "C:\Python34\lib\site-packages\click\core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Python34\lib\site-packages\click\core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Python34\lib\site-packages\click\core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "C:\Python34\lib\site-packages\mkdocs\cli.py", line 186, in gh_deploy_command
    gh_deploy.gh_deploy(config, message=message)
  File "C:\Python34\lib\site-packages\mkdocs\gh_deploy.py", line 69, in gh_deploy
    remote_branch)
  File "C:\Python34\lib\site-packages\mkdocs\utils\ghp_import.py", line 163, in ghp_import
    if not try_rebase(remote, branch):
  File "C:\Python34\lib\site-packages\mkdocs\utils\ghp_import.py", line 78, in try_rebase
    if sp.call(cmd) != 0:
  File "C:\Python34\lib\subprocess.py", line 537, in call
    with Popen(*popenargs, **kwargs) as p:
  File "C:\Python34\lib\subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "C:\Python34\lib\subprocess.py", line 1086, in _execute_child
    args = list2cmdline(args)
[...]

PR #1797, opened 2019-05-19
Provide sensible error message when git not found

onelivesleft:
Currently if you try to run `gh-deploy` without git installed you get a fairly unhelpful backtrace, that will lead you to the thread for #722 if you google it.  This adds a more helpful error message.

### DIFF
--- a/mkdocs/commands/gh_deploy.py
+++ b/mkdocs/commands/gh_deploy.py
@@ -14,8 +14,12 @@ default_message = """Deployed {sha} with MkDocs version: {version}"""
 
 
 def _is_cwd_git_repo():
-    proc = subprocess.Popen(['git', 'rev-parse', '--is-inside-work-tree'],
+    try:
+        proc = subprocess.Popen(['git', 'rev-parse', '--is-inside-work-tree'],
                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    except FileNotFoundError:
+        log.error("Could not find git - is it installed / on path?")
+        raise SystemExit(1)
     proc.communicate()
     return proc.wait() == 0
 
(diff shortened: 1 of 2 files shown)


mkdocs/commands/gh_deploy.py
```diff
@@ -14,8 +14,12 @@
 
 
 def _is_cwd_git_repo():
-    proc = subprocess.Popen(['git', 'rev-parse', '--is-inside-work-tree'],
+    try:
+        proc = subprocess.Popen(['git', 'rev-parse', '--is-inside-work-tree'],
                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    except FileNotFoundError:
+        log.error("Could not find git - is it installed / on path?")
```
waylan: I understand the purpose of the slash in the error message, but it could be confusing to those who actually need it. Perhaps we should reword the message. A few options:

* "... installed properly"
* "... installed and on your path"

Also, a note should be added to the release notes.

With those two changes, I'll merge this.

waylan:
Looks good. Add a note to the release notes, and I'll merge this.

mkdocs/commands/gh_deploy.py
```diff
@@ -14,8 +14,12 @@
 
 
 def _is_cwd_git_repo():
-    proc = subprocess.Popen(['git', 'rev-parse', '--is-inside-work-tree'],
+    try:
+        proc = subprocess.Popen(['git', 'rev-parse', '--is-inside-work-tree'],
                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
waylan: Flake8 is [complaining][1] about the indentation here.

You either need to indent line 19 so it lines up with the first character and the opening `(` on line 18, or break it up like this:

```python
proc = subprocess.Popen(
    ['git', 'rev-parse', '--is-inside-work-tree'],
    stdout=subprocess.PIPE, 
    stderr=subprocess.PIPE
)
```

[1]: https://travis-ci.org/mkdocs/mkdocs/jobs/617874568#L513

PR #1840, opened 2019-08-12
Docs: Clarify navigation structure

### LINKED ISSUE #1837: Title in markdown file used for navigation -> Clarify markdown style
Hey,

then navigation will use the title given in a markdown file, if no title was given in the »nav« option. See https://github.com/mkdocs/mkdocs/issues/316

The title in the document needs to have this schema:

    # H1 Title

However markdown supports this schema as well:

    H1 Title
    ======

mkdocs only accepts the first schema. I would expect it to work with the second schema as well.

Idea 1: Parse the second schema as well.
Idea 2: Clarify the accepted markdown style in the [documentation](https://www.mkdocs.org/user-guide/configuration/#nav).

pixelbrackets:
Clarify how the navigation structure is build & how to overwrite it.
Clarifiy how to overwrite page titles and what markdown styles
for headlines are supported to overwrite a page title.

Closes #1837

### DIFF
--- a/docs/user-guide/writing-your-docs.md
+++ b/docs/user-guide/writing-your-docs.md
@@ -85,17 +85,26 @@ If both an `index.md` file and a `README.md` file are found in the same
 directory, then the `index.md` file is used and the `README.md` file is
 ignored.
 
-### Configure Pages and Navigation
+### Navigation
+
+MkDocs can generate a navigation for you. It is based on all Markdown files
+in the [documentation directory](configuration.md#docs_dir). Each navigation
+level is sorted alphanumerically by the page title. Index files however
+will always be listed first within a sub-section.
+
+The page title is based on meta data in the linked Markdown file, a level 1
+headline on the first line of the document body or the filename
+(see [Meta Data](writing-your-docs.md#meta-data)). Since `README.md` files are
+treated like index files as described above, these files will be named »index«.
+
+#### Configure Pages and Navigation
+
+It is possible to manually define your navigation configuration if you would
+like your navigation menu sorted differently or use other page titles.
 
 The [nav](configuration.md#nav) configuration setting in your `mkdocs.yml` file
 defines which pages are included in the global site navigation menu as well as
-the structure of that menu. If not provided, the navigation will be
-automatically created by discovering all the Markdown files in the
-[documentation directory](configuration.md#docs_dir). An automatically created
-navigation configuration will always be sorted alphanumerically by file name
-(except that index files will always be listed first within a sub-section). You
-will need to manually define your navigation configuration if you would like
-your navigation menu sorted differently.
+the structure of that menu.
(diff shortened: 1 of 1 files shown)


docs/user-guide/writing-your-docs.md
```diff
@@ -365,11 +374,14 @@ specific page. The following keys are supported:
     1. A title defined in the [nav] configuration setting for a document.
     2. A title defined in the `title` meta-data key of a document.
     3. A level 1 Markdown header on the first line of the document body.
+       Please note that [Setext-style] headers are not supported.
     4. The filename of a document.
 
     Upon finding a title for a page, MkDoc does not continue checking any
     additional sources in the above list.
 
+[Setext-style]: https://github.com/mkdocs/mkdocs/issues/1826#issuecomment-511078194
```
waylan: Let's link to the actual definition of Setext-style headers: https://daringfireball.net/projects/markdown/syntax#header

pixelbrackets:
Ah, I understand what you mean. And yes, this seems to have be related to “issue” that you know the code and I dont. ;)

From my beginner perspective: I start the project, read the docs, and try to create a *minimal* configuration file. Aha, »nav« is not required, so ignore it… start my first build… its working… oh, it even generates a navigation for me… awesome!

Regarding this PR I now stumble upon the sentences “By default `nav` will contain an alphanumerically sorted, nested list of all the Markdown files found within the `docs_dir` and its sub-directories. If none are found it will be `[]` (an empty list).”. Without knowledge of the code: I can imagine that the code always works with `nav` and has a fallback if its empty. As a beginner I dont know about this sequence and honestly don't care. 

Thats why I reversed the assumption. Right now it mixes “you may define a `nav`… or dont… then this sorting is used… you may override the sorting by defining a `nav` however”. Instead I tried to say “You can let MkDocs autogenerate a navigation for you and your are done … if you want to modify anything in this behaviour you can do it by defining a `nav`, and here is how“.

PS: Since this seems to be a different issue I should probably split this PR.

waylan:
@pixelbrackets thanks for the explanation. I wrote my response assuming you were trying to address some other (unknown to me) issue. It is true that the code actually has a default of auto-generating the nav. In other words, the conclusions to drew were correct. However, we want to give users the impression that that is a fallback, not the normal use case.

Yes, when you are getting started you don't need to create a nav object in your config. However, we encourage everyone to do so. Even the [Getting Started Guide][1] includes [instructions][2] for defining a nav item in the config.

If you want to discuss this further, please split this out into a separate PR.

[1]: https://www.mkdocs.org/#getting-started
[2]: https://www.mkdocs.org/#adding-pages

pixelbrackets:
I understand your point. Yepp, I'll split this.

waylan:
Closing in favor of #1843.

PR #1850, opened 2019-08-22
Expose builder to the 'on_serve' hook

bih:
This change allows us to add additional files/folders to the `watch` command - the use case specifically here is for our mkdocs plugin which adds support for building documentation in monorepos with multiple `docs/` folders.

The alternative proposal I have is to expose a lambda called `watch` which wraps in `builder` automatically, but I'm happy to adjust as per your feedback.

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -53,7 +53,7 @@ def _livereload(host, port, config, builder, site_dir):
         server.watch(d, builder)
 
     # Run `serve` plugin events.
-    server = config['plugins'].run_event('serve', server, config=config)
+    server = config['plugins'].run_event('serve', server, config=config, builder=builder)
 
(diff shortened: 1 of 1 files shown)


waylan:
> This change allows us to add additional files/folders to the `watch` command

This is one of the intended use cases for the `on_serve` hook. I'm curious why you closed this. Note that the AppVeyor tests are known to be failing and I'm mostly ignoring them until I have time to work out the cause.

It occurs to me that back when the Plugin API was first created, the `pages` collection was part of the `config`. However, in the large refactor shortly before 1.0 was released, we separated the `nav` from `pages` as we no longer require the `nav` to contain all pages. It never occurred to me that we broke `on_serve` or we would have updated the hook as well. I'm not convinced passing `builder` to the hook is the right fix, but we certainly want to provide access to `files` and/or `pages` in some manner or another.

PR #1864, opened 2019-09-11
Remove unneeded IE8 polyfills

coliff:
Bootstrap 4.x doesn't work with IE8 at all.
IE8 marketshare is now just 0.1% anyway (https://caniuse.com/usage-table)

### DIFF
--- a/mkdocs/themes/mkdocs/base.html
+++ b/mkdocs/themes/mkdocs/base.html
@@ -29,11 +29,6 @@
       {%- endblock %}
 
       {%- block libs %}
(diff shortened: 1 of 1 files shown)


waylan:
> Bootstrap 4.x doesn't work with IE8 at all.

That seems like a reasonable reason to do this. Thanks.

waylan:
BTW, the holdup on the next release is described in #1865. @coliff I appreciate your work to update the theme, but for the work to see a release, we need to resolve that issue, which is directly tied to the theme. My time is limited to reviewing PR's right now so I don't know when I'll have time to get to it. Any help would be greatly appreciated.

PR #1878, opened 2019-10-08
Ensure theme files do not override docs_dir files on Windows

### LINKED ISSUE #1876: Theme files override docs_dir files on Windows
As investigated and discussed in https://github.com/squidfunk/mkdocs-material/issues/1282, custom files in docs_dir are overriden by theme files. 

For example, if you use the default MkDocs theme and create a custom `favicon.ico` in `docs/img`, the custom icon does not appear in the output. It is overwritten by the theme's own favicon.

The issue was already brought up in https://github.com/mkdocs/mkdocs/issues/1671 and fixed by https://github.com/mkdocs/mkdocs/pull/1672.

However, the fix (https://github.com/mkdocs/mkdocs/pull/1672) **only works on Unix systems (including Mac). It does not work on Windows.**

Windows handles path names differently, so I suppose the issue must be caused by that. See especially the use of `os.path` in `files.py` (https://github.com/mkdocs/mkdocs/commit/f8ac3c7ee).

I had a similar issue in mkdocs-exclude, see here: https://github.com/apenwarr/mkdocs-exclude/issues/2

wilhelmer:
Fixes #1876.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -62,6 +62,7 @@ your global navigation uses more than one level, things will likely be broken.
 * Bugfix: Ensure theme files do not override `docs_dir` files (#1671).
 * Bugfix: Do not normalize URL fragments (#1655).
 * Bugfix: Skip external URLs in sitemap.xml (#1742).
+* Bugfix: Ensure theme files do not override docs_dir files on Windows (#1876)
 * Add canonical tag to `readthedocs` theme (#1669).
 
 ## Version 1.0.4 (2018-09-07)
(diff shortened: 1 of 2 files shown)


waylan:
Looks good. If you add a note to the release notes, I'll merge this. 

Thanks.

wilhelmer:
Note added. The changes are on a `master` branch though, I saw that was a problem with my PR on python-markdown, too. So I'm supposed to create a new branch on my fork for each PR? Should I do that now or can you handle this? Sorry, didn't know, I'm fairly new to Git.

waylan:
Yes, in the future you should create a new branch for each PR. In this case its not a problem as I don;t need to make any updates or changes myself. On the other PR it was a problem because I needed to push some changes to your PR.

However, even with this PR, after I merge it you will have a conflict when you try to update your fork. You'll have to do a force push. But that is only an inconvenience for you. It won't effect by ability to merge it.

PR #1890, opened 2019-10-19
Fix GitHub project pages link

JohanBrorson:
Update the `GitHub project pages` link in the docs to an existing article. The current link takes you to a page that doesn't exit.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -353,7 +353,7 @@ the MkDocs IRC channel `#mkdocs` on freenode.
 [nav]: user-guide/configuration/#nav
 [discussion group]: https://groups.google.com/forum/#!forum/mkdocs
 [GitHub issues]: https://github.com/mkdocs/mkdocs/issues
-[GitHub project pages]: https://help.github.com/articles/creating-project-pages-manually/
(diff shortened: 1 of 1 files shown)


waylan:
Thanks for this. The problem is that the new page does not provide the information we need to point people to. If GitHub removed the info we need, that is unfortunate. Not sure what we can do about that.

JohanBrorson:
@waylan yea, I wasn't sure what information was expected on the page. I made some more digging and it looks like at some point http://help.github.com/articles/creating-project-pages-manually/ started to redirect to https://help.github.com/articles/creating-project-pages-using-the-command-line/ which now redirects to https://help.github.com/en/github/working-with-github-pages/creating-a-github-pages-site

If that page doesn't provide the needed information maybe the link should just be removed. What do you think?

waylan:
Yeah, the command line specific stuff is what we are looking for. It seems that has been removed. I seem to remember there being a public repo for the help site in the past, which would allow us to see what changes they made in their commit history, but I can't find it. I also checked the old URLs on archive.org, but there's nothing there either.

Issue #1865, opened 2019-09-11
Fix `mkdocs` theme inheritance


waylan:
In #1597 we updated Bootstrap from 3.0.3 to 4.1.3. However, when we did so, we renamed some files (`css/bootstrap-custom.min.css` => `css/bootstrap.min.css` and `js/bootstrap-3.0.3.min.js` => `js/bootstrap.min.js`). The problem is that other themes inherit from the `mkdocs` theme with the old file names. The renaming in #1597 breaks those themes. 

On the one hand, the new names make more sense. On the other hand, we promised forward compatibility with version 1 and there is no easy way (that I'm aware of) to gracefully deprecate the old filenames. We need to resolve this conflict before releasing version 1.1. In other words, this is a blocking issue to releasing version 1.1.

I see a few possibilities:

1. Revert the filenames. Yes, that means a file named `bootstrap-3.0.3.min.js` would actually contain a different version of Bootstrap, but it would be the easy fix."
2. Leave the change in-place and document it for third party themes. In practice, this probably would not be too disruptive as the only themes I'm aware of which make use of inheritance are [mkdocs-bootswatch][1] and [mkdocs-bootstrap][2], both of which are maintained by the MkDocs team.
3. Some other solution I haven't thought of.

If we go with option 1, then a PR to this repo would address the issue. If we go with option 2, then we would need a PR updating the release notes here as well as a PR each for mkdocs-bootswatch and mkdocs-bootstrap.

[1]: https://github.com/mkdocs/mkdocs-bootswatch
[2]: https://github.com/mkdocs/mkdocs-bootstrap

waylan:
For an example of how a theme inherits from the `mkdocs` theme, see mkdocs/mkdocs-bootstrap#22. In that PR, all common files between the two themes were removed from the child theme and the [mkdocs_theme.yml][1] config file was updated to indicate that the theme inherits from `mkdocs`. The only files remaining are the ones which override/replace the behavior in the parent theme.

The problem exists in the [main.html][2] template where `styles` and `libs` blocks are defined which override the blocks from the `mkdocs` theme. In each block, the old filenames are used. But in the current codebase, this would result in 404s.

[1]: https://github.com/mkdocs/mkdocs-bootstrap/pull/22/files#diff-1148036bbce8b114082a6d03e4e3cea4
[2]: https://github.com/mkdocs/mkdocs-bootstrap/pull/22/files#diff-76ffc1343ec2342f4ca2cdaa96038bc7

[...]

PR #1897, opened 2019-10-26
Theme-Inheritance

Ajwerth:
Updated docs, added a note to the styling section about bootstrap filename changes.

### DIFF
--- a/docs/user-guide/custom-themes.md
+++ b/docs/user-guide/custom-themes.md
@@ -680,7 +680,7 @@ Bootswatch theme].
 !!! Warning
     As of version 1.1 Bootstrap has been updated from 3.0.3 to 4.1.3. After this update 2 filenames were changed to avoid confusion. 
 
-    [MkDocs Bootstrap theme](https://github.com/mkdocs/mkdocs-bootswatch), [MkDocs Bootswatch theme](https://github.com/mkdocs/mkdocs-bootstrap) both inherit directly from mkdocs, if you are using either of these themes and are having issues with bootstrap you may need to change the filenames as described below.
+    [MkDocs Bootstrap theme](https://github.com/mkdocs/mkdocs-bootswatch), [MkDocs Bootswatch theme](https://github.com/mkdocs/mkdocs-bootstrap) both inherit directly from mkdocs, if you are using either of these themes and are having issues with bootstrap you may need to change the filenames or paths to these files as described below.
 
     Filename Updates:
         css/bootstrap-custom.min.css => css/bootstrap.min.css
--- a/docs/user-guide/custom-themes.md
+++ b/docs/user-guide/custom-themes.md
@@ -673,18 +673,18 @@ Bootswatch theme].
     [configuration] being defined, and they can then take advantage of the
     [custom_dir] to make tweaks to your theme to better suit their needs.
 
-!!! Warning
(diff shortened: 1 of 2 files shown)


Ajwerth:
#1865

docs/user-guide/custom-themes.md
```diff
@@ -677,6 +677,15 @@ Bootswatch theme].
 [MkDocs Bootstrap theme]: https://mkdocs.github.io/mkdocs-bootstrap/
 [MkDocs Bootswatch theme]: https://mkdocs.github.io/mkdocs-bootswatch/
 
+!!! Warning
```
froschdesign: I see a problem in this warning because it is only helpful to a fraction of the users. Any user who starts with a new template does not need this warning.

I would suggest to create a separate migration guide for this type of information.

Ajwerth: Thats a good point, I will be solving these issues in the 2 child themes. 

I'll remove this warning from, custom-themes.md, and remove the paragraph about the bootstrap themes from styling-your-docs.md.

Or do you think it doesn't make sense to mention the filename changes in these docs at all?

waylan: Yeah, lets remove this altogether. This sort of information goes in the release notes, not here.

In addition this is information which should be directed at authors of themes which inherit from the `mkdocs` theme, not users of those themes. We don't expect the users to modify the theme. Instead we expect the theme author to update the theme and the users to update to the new version.

docs/user-guide/styling-your-docs.md
```diff
@@ -28,6 +28,15 @@ The default theme, which was built as a custom [Bootstrap] theme, supports most
...
 
+!!! Warning
+   As of version 1.1 Bootstrap has been updated from 3.0.3 to 4.1.3. After this update 2 filenames were changed to avoid confusion. 
+   
+   Currently there are 2 themes which make use of inheritence from mkdocks [MkDocs Bootstrap theme](https://github.com/mkdocs/mkdocs-bootstrap) and [MkDocs Bootswatch theme](https://github.com/mkdocs/mkdocs-bootswatch),if you are using either of these and are having issues with bootstrap you may need to change the filenames paths to these files as described below.
+
+   Filename Updates:
+   css/bootstrap-custom.min.css => css/bootstrap.min.css
+   js/bootstrap-3.0.3.min.js => js/bootstrap.min.js
+
```
waylan: Again, this information should not be included here but in the release notes and directed at theme authors, not users.

docs/user-guide/custom-themes.md
Ajwerth: Ah yeah ok that makes sense, I'll take this out.

Ajwerth: Also is updating the release notes something I can do? Or should I just close this PR and make the updates to the themes that inherit?

waylan: Add a "Backward Incompatible Changes" sub-section to the 1.1 section of [docs/about/release-notes.md][1]. See the [similar section][2] for version 1.0 for an example. You can either update this PR or close it and open a new one, whichever is easier.

[1]: https://github.com/mkdocs/mkdocs/blob/master/docs/about/release-notes.md
[2]: https://github.com/mkdocs/mkdocs/blob/master/docs/about/release-notes.md#backward-incompatible-changes

Ajwerth: I'll just open a new one.

Issue #1905, opened 2019-11-15
Support Subressource Integrity


nbraud:
Subressource Integrity is a standard feature of HTML, that lets a webpage specify a hash when loading a ressource (say, a script, a CSS stylesheet, ...).

Currently, `mkdocs`' default themes load ressources from `cdnjs.cloudflare.com` without specifying their hash, so both CDNJS and Cloudflare (or anyone successfully impersonating them) could inject evil content there.

Please consider:
- adding support for Subressource Integrity;
- enabling it by default;
- letting users specify their own preferred CDN for loading 3rd-party javascript.

waylan:
A PR is welcome to add the appropriate hashes to the `<link>` and `<script>` tags in the existing themes.

As for letting users specify their own CDN, you can already do this by either [creating your own theme][0], or [customizing][4] an existing one. Additionally, an individual theme could offer the option through [theme configuration settings][3]. There is nothing that needs to change in the MkDocs core to do that.

Finally, for the record, here is the [W3C Spec][1]. Also see [Mozilla's overview][2].

[0]: https://www.mkdocs.org/user-guide/custom-themes/
[1]: https://www.w3.org/TR/SRI/
[2]: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
[3]: https://www.mkdocs.org/user-guide/custom-themes/#theme-configuration
[4]: https://www.mkdocs.org/user-guide/styling-your-docs/#overriding-template-blocks

waylan:
Upon further review, and the extensive discussion in #1906, this is being punted to be implemented via a third party theme and/or plugin. Initially I was expecting a PR to simply add a few hashes to existing URLs, but on further review it is clear that it is much more complicated than that -- especially if we need to update the hashes with every future change. However, the solution proposed in #1906 demonstrates that an entire working solution can be contained within a theme.

Therefore, a custom theme or a fork of an existing theme could implement a fully working solution. Alternatively, a third party plugin could be developed to compliment a theme. The higher security should not be necessary for the average site containing static documentation. Any organizations which need the higher security (due to internal policies) should also have the resources to provide their own solution.

limakzi:
@waylan As I see, this issue is pretty old.
Indeed, #1906 demonstrates that entire solution can be contained inside a theme.
Said that, I see a lot of examples, where `extra_javascript` is necessary, for example `KaTeX`, `MathAjax` support and it has nothing to do with a `theme`, per se.

[...]

PR #1906, opened 2019-11-15
Use subressource integrity in default themes

nbraud:
- [x] Added a datafile w/ the hashes of all assets for `highlight.js`, plus script for generating it.
- [x] Added macros `cdnjs.{url,load}` that allow using that data in templates.
- [x] Updated the default themes to take advantage of it.


nbraud:
@waylan This was a bit more complicated than you suggested, since:
- both themes use `highlight.js`, and the specific files they use depends on user configuration;
- `highlight.js` ships too many files for it to be reasonably handled by hand.

nbraud:
@astronouth7303 just taught me that, starting in Py3.7, [`importlib.resources`] is a better way to do it, but I guess that won't work for `mkdocs`' supported platforms.

[`importlib.resources`]: https://docs.python.org/3/library/importlib.html#module-importlib.resources

waylan:
Sorry, but at a glance this seems way too complicated. I'm not really interested in maintaining it. Adding some hashes to a few tags is one thing, but this is something completely different.

I hadn't considered the issue with highlight.js, but even if I was okay with the JavaScript stuff, this **must** be completely implemented in the themes. There should be no Python code changes.

Also, it occurs to me that highlight.js should be addressing this itself. Why are we adding on to their lib? I would suggest taking that upstream which would remove any need for us to address it here.

nbraud:
@waylan How would you suggest addressing it upstream?
The hashes must be provided in the `<script>` and `<link rel="stylesheet">` elements, so it's not obvious to me it's possible to do it upstream at all...

nbraud:
PS: I don't understand what you said re: “even if was okay with the JavaScript stuff”.
I didn't add or modify any Javascript, so all “the Javascript stuff” is already present in themes...

waylan:
Sorry, I apparently responded to quickly. In my response to #1905 I explicitly stated that this should all be possible without any changes to the MkDocs core. My understanding was that the entire patch would be contained to the themes. Therefore, when I read your summary I assumed you were using JavaScript within the themes. When I looked at the actual changes, once I saw Python changes, I stopped my review and responded. I never noticed that my assumption was wrong.

I still maintain my position that this must be contained to the themes, without any changes implemented in Python.

astronouth7303:
Given that cryptography has to happen, I don't think it's reasonable to implement strong SRI support with no python changes. The alternative would be to just modify the themes to use locked URLs (ie, exact version URLs that aren't supposed to be modified) and precompute the SRI data and embed it in the theme.

The pro of this minimal approach is that it becomes a few-lines change. The con is that updates require redoing that work every time the JS needs to be touched.

In fairness, this PR isn't just "Add SRI", it includes a bunch of tooling to refer to libraries by name/version, not URL. I would argue that as it stands, it's at a weird in-between for features; it currently: computes sri, and assembles a predictable URL from the given components (which you have to know them all anyways).

I see three good options for the level of functionality implementing SRI should provide:

1. No SRI computation provided, just modify themes to add SRI data in the markup (see above)
2. Basic SRI support: Add functions for to add SRI given a URL (eg, `sri_link(href)`/`sri_script(src)` or just `with_integrity(url)` inside of a tag)
3. Full CDNJS library: Given a library name (`jquery`, `highlight.js`), do all the API calls to figure out what all the library entails, and produce the markup to pull in all the pieces (eg `cdnjs_scripts(lib)`/`cdnjs_sheets(lib)`)

Unfortunately, I'm not sure CDNJS's data gives us enough information on what of the provided files are (1) Acually necessary, (2) May be referenced for additional functionality, or (3) Should not be used under normal/production use. Given this, I don't think Option 3 is feasible.

If I'm reading the thread above and the code correctly, @waylan wants option 1, and what @nbraud implemented was half-way between options 2 and 3.

If you want to encourage easy theme development, I say that option 2 is the better option: You can encourage best practices by reducing the onus of implementing said practices. If you'd rather minimize frills and leave that complexity in the hands of theme developers, go with option 1.

nbraud:
@waylan I found a way to implement this entirely within themes:
instead of generating a JSON file, a Jinja2 template file is generated (with the same data), and the `cdnjs_url` and `cdnjs_load` helpers are now Jinja2 macros.

There is still a new Python script, `themes/mkdocs/cdnjs.py` (`utils/cdnjs.py` in the previous version of this PR), which generates the `cdnjs_hashes.j2` data file.  Like previously, it is not used at runtime (in fact, it isn't even included in the Python packaging, `MANIFEST.in` excludes it), but only for maintaining the theme.

nbraud:
PS: I have a [minimalistic preview](https://deploy-preview-9--condescending-bhaskara-d5a97a.netlify.com/) setup with the changes from this PR.

nbraud:
Looks like the tests fail on Windows, presumably because symlinks work differently there: I symlinked `themes/mkdocs/cdnjs*` into `themes/readthedocs/` to avoid having multiple copies of the same code.

Is there another mechanism to share those templates between the 2 themes?

nbraud:
I just push a bunch of minor improvements; the most notable one is that `cdnjs.load(library, asset)` is now a thing, no need to specify the version number again (since it's specified in `cdnjs_libraries.json`)

waylan:
So I finally had the time to actually look at this to try to understand what the changes were actually doing. My quick glances earlier were confusing me because I was expecting that a few hashes would be assigned to some links, but the actual changes looked much more complicated than that.

In any event, @astronouth7303 spelled it out well. Thank you. As @astronouth7303 noted:

> If I'm reading the thread above and the code correctly, @waylan wants option 1, and what @nbraud implemented was half-way between options 2 and 3.

Exactly. I am only interested in option 1:

> 1. No SRI computation provided, just modify themes to add SRI data in the markup.

The computation is just one more thing to maintain. If someone wants to take the time to manually add the hashes to each link/script, I'll consider that. But anything more complex is not something I'm interested in (there should be no `cdnjs.py` file included or called anywhere). That said, the current approach is interesting in that it offers the computation completely within the theme, demonstrating that it is possible. I'm inclined to suggest that we close this PR as-is and save it as a reference point for any third-party (or future first party) themes. Then a new PR could be opened which implements option 1.

@squidfunk, as the developer of the most popular third-party theme, do you have any input on this?

squidfunk:
@waylan In general, I'm a big fan of the new security headers (HSTS, CSP, etc.) as well as tag-level security features like CORS and SRI. At some point there were two issues in Material that proposed to implement support for stricter CSP ([#858](https://github.com/squidfunk/mkdocs-material/issues/858)) and ([#859](https://github.com/squidfunk/mkdocs-material/issues/859)), but it turned out that we would loose some flexibility in regard to configuration. On the other hand, for [another project](https://github.com/squidfunk/terraform-aws-cognito-auth) of mine which has some stricter security requirements, I've implemented most of what's suggested by [Mozilla Observatory](https://observatory.mozilla.org/analyze/terraform-aws-cognito-auth.play.squidfunk.com).

Looking at the history of issues of the Material theme and questions that people have, I think implementing SRI or other similar security features would induce more problems that it would solve, as it would introduce additional barriers for people that want to change (i.e. hack) stuff. Many of the people using Material/MkDocs are not savvy in frontend development. That's one of the main reasons they use it - they get something that "just works" with minimal effort. I haven't reviewed the PR, but it would mean that dropping something into `extra_javascripts` would always mean to include the security hash, or call some post-processing logic which does that, correct?

I don't want to overcomplicate things and I think one of the reasons why the theme is so popular is that it "just works". In the end, we're talking about technical documentation (that's at least what this project major aim is) and not MFA banking authentication. If somebody wants to tighten the requirements, they can always extend or fork the theme and add SRI on their own.

My 2 cents.

nbraud:
> I am only interested in option 1:
> 
> > 1. No SRI computation provided, just modify themes to add SRI data in the markup.
> 
> The computation is just one more thing to maintain. If someone wants to take the time to manually add the hashes to each link/script, I'll consider that. But anything more complex is not something I'm interested in (there should be no `cdnjs.py` file included or called anywhere).

Does that mean you would consider the exact same PR, with the `cdnjs.py` file removed?

That sounds strictly-worse, from a theme's maintainability perspective: `highlight.js` alone ships 259 files, every single one of which is potentially used by the theme (depending on `config.theme.hljs_style` and `hljs_languages`).

If the issue is that you do not want the `cdnjs.py` script to exist in the repository (or be expected to maintain it), I could maintain and package it separately, and it wouldn't be any of `mkdocs`' concern if some theme maintainers happen to use that tool to manage part of their templates' data.

nbraud:
>  I haven't reviewed the PR, but it would mean that dropping something into extra_javascripts would always mean to include the security hash, or call some post-processing logic which does that, correct?

@squidfunk  No; the PR only adds hashes for the ressources that the `mkdocs` and `readthedocs` themes load from `cdnjs.cloudlfare.com`.  There's nothing forcing themes to implement SRI, nor is there anything requiring users to specify SRI hashes for `extra_javascript` (or even the ability to do so if they wanted).

TL;DR: if a user adds a `<script>` tag (through `extra_javascripts` or by modifying a theme), there won't be any hash (any `integrity` attribute) unless they go out of their way to add one.

(As a quick aside: it would be fairly easy to automatically add hashes for `extra_javascript`, since those files are available when generating the templates, but that would require a Python helper that does the hash computation, and @waylan made it clear it wouldn't be considered for inclusion in `mkdocs`)

squidfunk:
@waylan Isn't this a great use case for a plugin that might be integrated downstream in themes? The plugin could calculate the checksums automatically, ideally after building the site, injecting the respective attributes in the built HTML and keeping a JSON file with the computed checksums which could be checked into VCS, similar to a `package-lock-json` file.

I don't know if the current architecture supports it, but it's a better way to implement it outside of MkDocs, as it's not a "killer" feature that most people will want to use. Probably only a very small subset.

nbraud:
> @waylan Isn't this a great use case for a plugin that might be integrated downstream in themes? The plugin could calculate the checksums automatically, ideally after building the site [...]

Is that a reply to my aside about `extra_javascripts`?  'cause for the current proposal, doing it in the theme makes IMO more sense:
- All users of the theme automatically get the benefits. In particular, if a theme pulls in resources from an external location like a CDN, users may not be aware of it (and so cannot make an informed decision whether to trust that CDN); SRI mitigates that trust issue, as it ensures that the resources are the ones the theme's author intended.
- There's no need to compute those hashes when rendering templates, since the theme knows exactly which resources it can potentially use.
- The process of fetching external resources to compute hashes is somewhat slow (half a minute for the ~300 assets in `highlight.js`, over my home connection). This isn't a problem if it's a thing the theme's maintainer does once, but not-so-much when it's done on potentially every build.

waylan:
> In the end, we're talking about technical documentation (that's at least what this project's major aim is) and not MFA banking authentication. If somebody wants to tighten the requirements, they can always extend or fork the theme and add SRI on their own.

Exactly. MkDocs target use-case is generating a static site of technical documentation. Users of the site would not be providing personal info, passwords or the like to such a site. This is not anything with high security needs. If an organization with higher security requirements is using MkDocs, then they should have the resources to meet those requirements on their own. Whether they fork an existing theme or build their own from scratch is their choice.

> Does that mean you would consider the exact same PR, with the `cdnjs.py` file removed?

Not exactly, Also remove the list of hashes, Each hash should be hardcoded into the template along with the URL. In other words, the only change I was willing to consider is one which adds the hashes directly to the `link` and `script` tags. However, I don't think I am even willing to consider that anymore (see below).

> That sounds strictly-worse, from a theme's maintainability perspective: `highlight.js` alone ships 259 files, every single one of which is potentially used by the theme (depending on `config.theme.hljs_style` and `hljs_languages`).

I agree. Perhaps it would make more sense to add SRI to a theme which did not use `highlight.js`.  I think this need would best be filled by a third party theme. As is demonstrated in this PR, the entire implementation can be contained within the theme. And if the theme dev is motivated by the need for SRI, then they can ensure the theme maintains that standard. I can't guarantee every future submission I review for the existing themes will.

> Isn't this a great use case for a plugin that might be integrated downstream in themes? The plugin could calculate the checksums automatically, ideally after building the site, injecting the respective attributes in the built HTML and keeping a JSON file with the computed checksums which could be checked into VCS, similar to a `package-lock.json` file.

I hadn't considered that, but yes I suppose a third-party plugin could be useful as well. Plugins have access to the Jinja environment through the `on_env` event and could supplement a theme. There are also the three template events which can alter a template or its context. 

Another possible plugin approach would be to implement something similar to how translation works in templates. A dummy do-nothing function could be defined which would wrap `link` and `script` tags defined in templates. Then a plugin could replace that function with one which computed and inserted hashes into those tags. Of course, for this to work, the dummy do-nothing function would need to be added to the core, along with the calls to it in the templates.

Whichever solution is used, it seems like this can start out as a third party solution. If there are any roadblocks to a solution, I am willing to consider minimal changes to the core to make those solutions possible. However, the solution needs to be implemented outside of MkDocs. itself.

Given the above, I am closing this.

Issue #1919, opened 2019-12-07
Unexpected behaviour with page.is_homepage


StarfallProjects:
Starting a new site and rolling my own theme. Came across some slightly odd behaviour.

Mkdocs version 1.0.4
Python version 3.7.1

**Expected:**
`page.is_homepage` evaluates to True on the home (index.md) of the site, and false on all other pages.

**Actual:**
`page.is_homepage` evaluates to True on the home (index.md), and on any other index.md that is included in the nav object without nesting.

**Examples:**

The unexpected result:

```
nav:
  - Home: index.md  <--- page.is_homepage evaluates to True
  - About: about.md <--- page.is_homepage evaluates to False
  - Projects: projects/index.md <--- page.is_homepage evaluates to True
```

Changing the filename causes it to evaluate to false:

```
nav:
  - Home: index.md  <--- page.is_homepage evaluates to True
  - About: about.md <--- page.is_homepage evaluates to False
  - Projects: projects/test.md <--- page.is_homepage evaluates to False
```

If I tweak it a bit, so that the sections are nested, then it evaluates to false as I'd expect:

```
nav:
  - About: 
      - About: about.md <--- page.is_homepage evaluates to False
  - Projects: 
       - Project home: projects/index.md <--- page.is_homepage evaluates to False
```

This feels like a bug - especially as simply changing the markdown file name causes the behaviour to change.

waylan:
Yes, that's a bug. Thanks for the report.

https://github.com/mkdocs/mkdocs/blob/master/mkdocs/structure/pages.py#L94-96

We are only checking that the page is an index page at the root of the `nav` and not looking at the path at all. However, we do not require the `nav` to match the file structure. We simply need to add a check that the path is at the `doc_dir` root.

PR #1920, opened 2019-12-07
page.is_homepage evaluates to True only for index.md at route of docs_dir

StarfallProjects:
Fixes: https://github.com/mkdocs/mkdocs/issues/1919

Note to reviewers/maintainers: I ran the tests (my version of Python is 3.7) and it looked like they passed, and I tried it on the site where I saw the original issue, and it looked like it fixed it. But I am not a Python dev (or really a developer at all), so I apologize in advance if this isn't a good solution - bit of a learning exercise for me here!

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -93,7 +93,7 @@ class Page(object):
 
     @property
     def is_homepage(self):
-        return self.is_top_level and self.is_index
+        return self.is_top_level and self.is_index and self.file.url == '.'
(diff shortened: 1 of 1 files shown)


waylan:
@StarfallProjects that looks good. I added tests and updated the release notes in #1921. If you had used a branch other than master, I could have easily pushed to additions here. However, in this case, it was easier to just create a new PR. Therefore I am closing this. Still using your commit though. So thank you.

Issue #1926, opened 2019-12-11
Drop 2.7 Support?


waylan:
Python will officially drop support for PY27 on January 1, 2020 (as [documented][0]), which is less than 3 weeks away. Presumably we should drop support as well.

I had intended for the next release (1.1.0) to maintain support for PY27, but given the long delay in getting it out, I don't think it makes sense anymore. I expect the 1.1.x series to remain current for some time and it doesn't make sense to continue to support PY27 past the next release.

Additionally, in #1797 we introduced a [check for `FileNotFoundError`][1] and that error does not exist in 27 (it is PY3 only). Therefore, the "deploy" command will no longer run on PY27. We missed it in the tests because calls to the system are mocked in the tests. The [Appveyor Flake8 test][2] caught it, but I ignored that because I hadn't realized that a previous [error with Appveyor][3] which I not understood seems to have resolved itself. The Travis tests were switched to PY3 a while ago in #1782.

As the code doesn't run in PY27 anyway, we should probably run [asottile/pyupgrade][4] (with the command `pyupgrade --py3-plus`) on the code to convert everything to PY3 specific syntax. We also need to audit any `TODO` comments for any additional changes previously noted.

[0]: https://devguide.python.org/#status-of-python-branches
[1]: https://github.com/mkdocs/mkdocs/pull/1797/files#diff-e3013b3fd93785be7022b4a3ba2733a7R23
[2]: https://ci.appveyor.com/project/d0ugal/mkdocs/builds/29168409/job/47buab7k19x368pe#L62
[3]: https://ci.appveyor.com/project/d0ugal/mkdocs/builds/27953652/job/vox091savneqrw94
[4]: https://github.com/asottile/pyupgrade

PR #1936, opened 2019-12-17
Drop Python 2.7.

waylan:
So far, this is step 1 in the process. Let's make sure the tests are working properly and then we can move on to code changes. Ref #1926.


waylan:
The failing tests are because some of the new stuff isn't supported in PY34. However, that has had support dropped even earlier than PY26. I'll address that for MkDocs in a separate PR as I'd like to keep each as a separate commit but I don't want to squash all of the commits in this PR's history for later reference.

PR #1938, opened 2019-12-20
Drop Python 3.4 and add Python 3.8.

waylan:
Python dropped support for 3.4 months ago.


waylan:
The failing MarkdownLint test is unrelated to the changes here.

Issue #1629, opened 2018-09-14
Refactor Docs for Devs and Users


waylan:
The documentation could be reorganized to better separate the info for users from the info for developers (themes and plugins). I'm thinking the Nav might look something like this (actual page titles and order subject to change):

```
- Home
- User Guide:
    - Getting Started
    - Installation
    - File Layout
    - Writing Markdown
    - Configuration
    - Themes
- Developer Guide:
    - Themes
    - Plugins
    - API
- About:
    - Release Notes
    - Contributing
    - License
```

`Getting Started` and `Installation` have been broken out into their own pages. The `Home` page should be a simple introduction with prominate links to the new pages (and perhaps a few others).

Note that the `Developer Guide` and the `User Guide` each contain a `Themes` page. Each would contain theme related documentation specific to the needs of theme develpers and users respectively. Currently the two are mixed with pieces of each intermingled, which makes it hard for devs to find what they need and can be overwhelming for users.

In the `Developer Guide`, the `Themes` and `Plugins` pages would contain mostly prose regarding their respective topics, while the `API` page would document the various classes, methods, and functions in the MkDocs' code base which plugin devs might need to use. For example, plugins can define their own config, but we currently have no documentation about the config classes and how to use them. We also define page and nav objects for theme devs (attributes which might be called from a template), but not for plugin devs (all public attributes and methods of the class). Many of the `utils` functions could be useful to plugin devs as well.

Much of the work consists simply of rearranging the existing documentation with minor edits to make the text fit in it's new context. However, the `API` is currently non-existent. I see two possible paths:

1. Author the documentation from stratch in Markdown.
2. Build a plugin which extracts documentation from the source code.

Option two would certainly be more userful long term (and it provides a new feature for other users), but it would be more work as much of the source code has very limited documentation in the comments.

Whichever way we go, the rest of the work could be started now. The `API` can always be added later.

squidfunk:
Quoting your comment in #1871 and moving the discussion over here:

> I suspect that proposed change would help immensely with the MkDocs vs. theme problem. If only I had the time to do the work... Please feel free to provide any feedback on the idea in that issue rather than here.

The proposed changes may be a good start reducing any potential template-related issues that are opened here and related to non-default templates like for example `mkdocs-material`. I haven't checked how often this happens though.

[...]

PR #1950, opened 2019-12-29
Links to the MkDocs Plugins wiki page, a few extra links regarding user guide, themes and plugins

twardoch:
Since I’ve refactored and extended the MkDocs Plugins wiki page ( https://github.com/mkdocs/mkdocs/wiki/MkDocs-Plugins ), I thought it’d be a good idea to improve its visibility by adding a link from the docs. I’m also adding a few more links to better connect the introduction with the user guide.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -18,9 +18,9 @@ Amazon S3, or [anywhere][deploy] else you choose.
 
 ### Great themes available
 
-There's a stack of good looking themes available for MkDocs. Choose between
+There's a stack of good looking [themes] available for MkDocs. Choose between
 the built in themes: [mkdocs] and [readthedocs], select one of the 3rd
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -343,10 +343,14 @@ you're done. For specific instructions on a number of common hosts, see the
 To get help with MkDocs, please use the [discussion group], [GitHub issues] or
 the MkDocs IRC channel `#mkdocs` on freenode.
 
+[guide]: user-guide/
```
waylan: This results in a `404 Not Found` error as there is no index page for any subdirs in the docs. Let's just remove the link for now. This can be addressed more fully when we get to #1629.

Otherwise this looks good. Thanks.

twardoch: Fixed in the PR

Issue #1972, opened 2020-02-03
Insert something in mkdocs.yml


uliska:
Is it possible to have some code automatically inserted into the configuration file while building?

I have a set of sites that are to be configured nearly identically, except for the colors of the Material theme. Rather than having to manually *maintain* the identical portions of the `mkdocs.yml` file I would like to retrieve the data directly from a central file.

When plugins start to work, the config file has already been read, isn't it?

I can (and probably will) write a script to update the "child" config files, which I can run after changing the "master" file, but a "live" solution would be cleaner.

waylan:
We are simply using standard YAML parsing with the [PyYAML][1] library. YAML is defined by a very strict [spec][2] and does not offer any method for including partial files. Therefore we do not either.

However, we do offer a Plugin API, which includes an [`on_config`][3] event. That event is intended to offer plugins an opportunity to alter the config after the config file is parsed and loaded, but before the build process begins. That would be the best way to implement something like this.

[1]: https://pyyaml.org/wiki/PyYAML
[2]: https://yaml.org/spec/
[3]: https://www.mkdocs.org/user-guide/plugins/#on_config

uliska:
Thank you for the explanation. Probably this event would be a good starting point, but I realized that my needs go further than that. I ended up with a pretty powerful "meta" tool that allows the creation of a multi-book site with a template based approach and even merging the search indexes of the resulting sub-books. I will use this for the documentation of a library with multiple packages, where documentation for each package is maintained in a separate MkDocs sub-book (and Git repository).

I will announce it when it is ready to be reviewed and considered as a possible citizen in the MkDocs ecosystem. What would be the appropriate channel to do so (more like starting a discussion than making an announcement)?

waylan:
If you would like any work you do to be listed, there are a number of publicly editable pages on the [wiki][1] for that purpose.

For making announcements or having discussions, the [mailing list][2] would be the appropriate place to do that.

Note that both are listed in the [README.md][3] file.

[1]: https://github.com/mkdocs/mkdocs/wiki
[2]: https://groups.google.com/forum/#!forum/mkdocs
[3]: https://github.com/mkdocs/mkdocs#mkdocs

PR #1978, opened 2020-02-08
Add link to README to contributing.md

uliska:
In https://github.com/mkdocs/mkdocs/issues/1972#issuecomment-582952604
I was pointed to the README, but I hadn't looked for contact information there.
I think looking on the website is a reasonable way, and pointing to the README
from the "Contributing" page looks like a good help to me.

### DIFF
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -10,6 +10,10 @@ ways, a few examples are:
 - Documentation improvements
 - Bug reports and patch reviews
 
+For information about available communication channels please refer to the
+[README](https://github.com/mkdocs/mkdocs/blob/master/README.md) file in our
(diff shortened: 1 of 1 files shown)


waylan:
The general premise here is good. However I wonder if perhaps a better solution would be to include the information directly in the contributing guide itself. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

uliska:
I don't know the project good enough to tell. But I would strongly suggest to keep the actual information in a single location.
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

oprypin:
I agree with the latter comment. Thank you for the contribution.

PR #1982, opened 2020-02-11
Fix link to virtualenv user guide (in contributing section of the docs)

maxalbert:
Thanks a lot for MkDocs and to everyone who has contributed to it!

Minor issue: the link to the virtualenv user guide in the [contributing](https://www.mkdocs.org/about/contributing/#installing-for-development) section of the docs is broken.

This PR fixes that.

### DIFF
--- a/docs/about/contributing.md
+++ b/docs/about/contributing.md
@@ -65,7 +65,7 @@ Once you are happy with your changes or you are ready for some feedback, push
 it to your fork and send a pull request. For a change to be accepted it will
 most likely need to have tests and documentation if it is a new feature.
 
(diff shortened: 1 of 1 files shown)


maxalbert:
Hmm, the Travis CI build fails but this seems unrelated to the PR?

waylan:
Yeah, it would appear that there is an issue with either `tox` or `virtualenv`. Specifically, it appears that an error occurs when tox attempts to create a virtualenv. Strangely, the error only occurs for some versions of Python, but it seems to occur consistently for the same versions.

In any event, I accepted this PR as the failure is clearly not related to this change.

maxalbert:
Thanks!

PR #2007, opened 2020-02-26
Fix a minimum dependency description of Markdown at the release note

thombashi:
The release note describes:

>* Update minimum dependancy to Markdown 3.0.1.

However, the actual minimum dependency of Markdown package at mkdocs 1.1 is 3.2.1.

https://github.com/mkdocs/mkdocs/blob/eddeeab066cb071dbe426878a85de715df1b93e7/requirements/project.txt#L4


waylan:
Thanks for catching this and the fix. Normally we don't deploy docs without a release, but in this case I did as this is the only change since the release and it is fixing a misleading error in the release notes.

PR #2015, opened 2020-03-05
Release a public docker image on docker hub

lvjp:
I saw few non-official mkdocs docker image and I think that it can be a good idea to have one.

For the moment, I have made a simple start :
- Automatically built image on Docker Hub's infrastructure
- Image built on each git tag
- [label schema](http://label-schema.org/) compliant

I used my personal account to produce a poc :
https://hub.docker.com/r/lvjp/mkdocs

The only setting I have done is to add this building rule :

 Name | Value 
------ | ---------
Source type | Tag
Source | /^[0-9.]+.*$/
Docker tag | {sourceref}
Build context | /docker
Autobuild | On
Build Caching | Off

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,22 @@
+FROM python:3.8.2-alpine3.11
+
+ARG BUILD_DATE
+ARG SOURCE_BRANCH
+ARG SOURCE_COMMIT
+
+RUN pip install mkdocs=="${SOURCE_BRANCH}"
+
+# Default exposed port by 'mkdocs serve' command
+EXPOSE 8000
+
+ENTRYPOINT [ "/usr/local/bin/mkdocs" ]
+CMD [ "--help" ]
+
+LABEL \
+    org.label-schema.build-date="${BUILD_DATE}" \
+    org.label-schema.name="mkdocs" \
+    org.label-schema.description="Project documentation with Markdown." \
+    org.label-schema.vcs-url="https://github.com/mkdocs/mkdocs" \
+    org.label-schema.vcs-ref="${SOURCE_COMMIT}" \
+    org.label-schema.version="${SOURCE_BRANCH}" \
+    org.label-schema.schema-version="1.0"
--- a/docker/hooks/build
+++ b/docker/hooks/build
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+docker build \
+	--build-arg BUILD_DATE="$(date -u +'%Y-%m-%dT%H:%M:%SZ')" \
+	--build-arg SOURCE_BRANCH="${SOURCE_BRANCH}" \
+	--build-arg SOURCE_COMMIT="${SOURCE_COMMIT}" \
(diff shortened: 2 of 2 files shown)


waylan:
What is the use case here? It appears that this is installing MkDocs from the git HEAD rather than a release (from PyPI). The only need for that is for development of MkDocs itself. And that should not be necessary as we have a tox config.<sup>1</sup> For anything else, users should be crafting their own custom configs anyway.  For example, if you are using a docker image to do test builds of your documentation, it seems to me that you should be using a production release of MkDocs which requires nothing more than `pip install mkdocs` and `mkdocs build`. However, the setup (gathering your source files) is going to be custom to your project, so a custom dockerfile would be required anyway. 

- - -

<sup>1</sup> Tox is light-weight compared to docker as it isolates the environment through a Python virtualenv. However, I generally don't  even bother using tox for development, but let the CI server (of which we have configs set up) run the tests once I push a PR. The point is that for development of MkDocs, there is no need for a docker image.

lvjp:
The use case is for person who want to use mkdocs in their CI pipeline. With the publication of a docker image it can be almost used out of the box.

For example, this GiLab CI pipiline will run mkdocs and push generated file to GitLab pages :
```yaml
pages:
  image: mkdocs/mkdocs:1.11
  scripts:
    - cd docs
    - mkdocs build --strict --site-dir ../public 
  artifacts:                                                                                   
    paths:                                                                                     
      - public 
```
As you can see in this use case, you do not need to embed the configuration inside the docker image. In other words, the final user do need to create own docker image.


I think that there is a misunderstanding about `installing from the HEAD version`.
This is maybe because my explanation was very too short...
With the settings I gave you, the jobs will be triggered only when new tag is pushed. That's it, a new docker image will be created when a new release (git tag) will be created.
If you want to have a full control on when a new image is created/release you can do the work inside the Travis pipeline and usage of docker hub deploy token.

waylan:
> the jobs will be triggered only when new tag is pushed.

Wait. So you want a build of your documentation to run every time MkDocs pushes a tag? I can't image any time I would need/want that behavior. Therefore, this seems like a custom use case that is not for everyone.

Regardless, I have given this some more thought and the MkDocs project itself has no use for a docker file. Adding one would only be more maintenance with no benefit for us. There is no reason why a third party can't maintain a docker file. In fact, we don't maintain any other system specific packages, so why would docker be any different?

lvjp:
I want a new image of mkdocs built from this repo each time a new tag is pushed.
This can permit the end-user to use it in CI pipeline / local test....

At first glance, I did not catch the fact that you do not maintain any other system specific packages.
Docker can be different because :
- End-user can try/execute it without install it to his lovely computer
- It can be easily used (docker pull) inside a CI/CD pipeline
- Docker image file can be run without think about the distribution the user have installed
- Build can be reproducible (also for bug report)

oxr463:
See also: https://gitlab.com/lramage/mkdocs-ci-pipeline

lvjp:
Finally, I build my [own image](https://gitlab.com/lvjplabs/docker-mkdocs) and publish it at [Docker Hub](https://hub.docker.com/r/lvjplabs/mkdocs).

Issue #2014, opened 2020-03-02
Not possible to search for fewer than 3 characters.


jimconner:
One of our mkdocs sites contains details about IBM MQ, but currently our users are telling us that they cannot find any documentation because when they search for 'MQ' it comes back saying 'No Results Found'. 

There appears to be a hardcoded minimum search length configured in the doSearch() function within `contrib/search/templates/search/main.js`. If the search term query.length is >2 then the search is done, but if not then empty results are displayed instead. This is misleading - If the search is too short then a message should be displayed to indicate that is the reason why no results are available.

```
function doSearch () {
  var query = document.getElementById('mkdocs-search-query').value;
  if (query.length > 2) {
    if (!window.Worker) {
      displayResults(search(query));
    } else {
      searchWorker.postMessage({query: query});
    }
  } else {
    // Clear results for short queries
    displayResults([]);
  }
}
```

It would be really great to have a config parameter so that the minimum search length could be configured, but even without that, having a useful message appear to let users know that their search was too short would be really helpful for us - it would save us from having to answer the phone a few times a week to point people in the right direction.

waylan:
As I understand it, in most cases a search of less than 3 characters is not going to result in useful results, thus the default. However, there are some specific cases where a 2 character search could be useful. But those cases are relatively few and not a high priority for us. That being the case, we would be willing to consider a PR which added a config option to the search plugin, but otherwise are not likely to get to this feature anytime soon.

jimconner:
Thanks Waylan - I'll see if I can put together a respectable looking PR tomorrow.

jimconner:
[Pull request 2018](https://github.com/mkdocs/mkdocs/pull/2018) now raised which changes the default message displayed if the search term was too short (but leaves the 'No results message' alone if there were actually no results returned).

The PR doesn't address the idea of making the minimum search length configurable, but it will be sufficient for our needs to stop our users getting confused.

waylan:
As a reminder, a significant reason for the character limit is for the "live update" feature which some themes make use of. As an example, the `mkdocs` theme does not have a "search page" but instead provides a "search modal" which is available on all pages (you can find an example on https://mkdocs.org). As the user types, the search results are updated for each key press. Searches with no results which return a result would be extremely confusing to users. 

In contrast, other themes, such as the `readthedocs` theme, have a search page, where the user types out their search terms and then submits. In that case, it might  make more sense to not limit character count.

[...]

PR #2018, opened 2020-03-09
Support for minimum search length configuration parameter.

jimconner:
As mentioned in [issue 2014](https://github.com/mkdocs/mkdocs/issues/2014), searching for any term of less than 3 characters will give the user back a message saying 'No results found' - This is misleading as results may have been present, but due to a hardcoded limit on minimum search length the search was not actually performed.

This PR implements an additional config parameter `min_search_length` for the search plugin.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -25,6 +25,7 @@ The current and past members of the MkDocs team.
 
 Bugfix: Ensure wheel is Python 3 only.
 Bugfix: Clean up `dev_addr` validation and disallow `0.0.0.0`.
+Add support for `min_search_length` parameter for search plugin (#2014).
 
 ## Version 1.1 (2020-02-22)
 
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -506,8 +506,9 @@ plugins:
 
 An integer value that defines the minimum length for a search query. By default
 searches shorter than 3 chars in length are ignored as search result quality with
-short search terms is poor. However, for some use cases (e.g. searching for 'MQ') 
-it may be preferable to set a shorter limit.
+short search terms is poor. However, for some use cases (such as documentation
+about Message Queues which might generate searches for 'MQ') it may be preferable
+to set a shorter limit.
 
 ```yaml
 plugins:
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -502,6 +502,21 @@ plugins:
 
   **default**: `'[\s\-]+'`
 
+##### **min_search_length**
+
+An integer value that defines the minimum length for a search query. By default
+searches shorter than 3 chars in length are ignored as search result quality with
+short search terms is poor. However, for some use cases (e.g. searching for 'MQ') 
+it may be preferable to set a shorter limit.
+
+```yaml
+plugins:
+    - search:
+        min_search_length: 2
+```
+
+  **default**: 3
+
 ##### **lang**
 
 A list of languages to use when building the search index as identified by their
--- a/mkdocs/contrib/search/__init__.py
+++ b/mkdocs/contrib/search/__init__.py
@@ -36,6 +36,7 @@ class SearchPlugin(BasePlugin):
     config_scheme = (
         ('lang', LangOption(default=['en'])),
         ('separator', config_options.Type(str, default=r'[\s\-]+')),
+        ('min_search_length', config_options.Type(int, default=3)),
         ('prebuild_index', config_options.Choice((False, True, 'node', 'python'), default=False)),
     )
 
--- a/mkdocs/contrib/search/templates/search/main.js
+++ b/mkdocs/contrib/search/templates/search/main.js
@@ -43,7 +43,7 @@ function displayResults (results) {
 
 function doSearch () {
   var query = document.getElementById('mkdocs-search-query').value;
-  if (query.length > 2) {
+  if (query.length > min_search_length) {
     if (!window.Worker) {
       displayResults(search(query));
     } else {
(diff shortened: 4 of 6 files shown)


waylan:
I am not comfortable with the message being a "search result". If any such message is being displayed, then it should be in the same format as the "no results" message, which is distinctly different from a search result.

jimconner:
I have updated this PR so that it will now accept a min_search_parameter define within the plugin->search block in mkdocs.yml.

I struggled to find a way to get the config values visible within main.js, so I cut 'n' pasted the bits from worker.js that allows it to fetch the config values. Hopefully this is an acceptable approach. I'm more than happy to make tweaks if you can suggest a better approach to take.

jimconner:
I see that this is failing to pass the tests in Travis - I'll have a go at fixing that up now.

waylan:
> I struggled to find a way to get the config values visible within main.js, so I cut 'n' pasted the bits from worker.js that allows it to fetch the config values. Hopefully this is an acceptable approach.

It is not as it completely negates the benefits of using a worker. Now the json file needs to be fetched twice, once in main and once in the worker. Some users have reported thousands of files which result in huge indexes for their site. Anything except the worker trying to handle that will make their pages unusable.

Also, I find using your product as the example in our documentation to be highly inappropriate.

jimconner:
I think there is a misunderstanding here. RabbitMQ is not our product.. just an open-source technology we use that causes our users to search for 'MQ' when they're trying to find out how to use it.

Any pointers on how I can get config values within worker.js would be greatly appreciated. This is my first dabbling with javascript, so I'm mostly trying to guess what the right approach is here. Thanks.

waylan:
> Any pointers on how I can get config values within main.js would be greatly appreciated.

You have two options:

1. Include the setting in the json file and have the worker pass it back to the main thread. For example, add a call to `postMessage` right after `init` which passes back config settings:

    https://github.com/mkdocs/mkdocs/blob/2b92209d12bfed050e35eeee2e5073074939727c/mkdocs/contrib/search/templates/search/worker.js#L118-L127

    This would be helpful for other possible future features as well. We haven't done it yet because we haven't yet added a feature which required it.

    As a reminder, we also need to provide a fallback for browsers which do not support WebWorkers here:

    https://github.com/mkdocs/mkdocs/blob/2b92209d12bfed050e35eeee2e5073074939727c/mkdocs/contrib/search/templates/search/main.js#L79-L90

2. Set a variable in the main thread via the template like we do here:

    https://github.com/mkdocs/mkdocs/blob/2b92209d12bfed050e35eeee2e5073074939727c/mkdocs/themes/mkdocs/base.html#L184-L187

    The trick here is that we only want to do that when the search plugin is enabled. And of course, that would add additional requirements for third party themes to support. Therefore, I would prefer the first option to this.

jimconner:
Thank you for the pointers Waylan - I'll see if I can get PR attempt 3 arranged tomorrow.

jimconner:
Getting close on this now I think. I've got the postMessage mechanism working for passing the config back from worker to main, fixed the failing test, added new one for min_search_length and rebased it to tidy up the commit history. Currently trying to find a way of disabling web workers in either chrome or safari so that I can work on the fallback mechanism. (My google skills are failing me with finding ways to disable WebWorkers)

jimconner:
Just about at the end of my working day, so I'll finish off with a question. As I can't work out how to disable WebWorkers in Chrome, I tried forcing the code in main.js to follow the other path...
```
//if (!window.Worker) {
if (true) {
  console.log('Web Worker API not supported');
  // load index in main thread
```

This gives me errors because within the doSearch() function within main.js it attempts `searchWorker.postMessage({query: query});` which cannot succeed because searchWorker is only instantiated in the event that we're using WebWorkers. 

I'm very probably doing something wrong (since this is my first dabbling with javascript), but I thought I should check with you as I don't understand how this could ever work. Is there something else that would be instantiating a searchWorker when WebWorker API is not available ?

waylan:
> I tried forcing the code in main.js to follow the other path

My recollection is that I did the same thing when developing the initial implementation. However, there have been some changes by others who are more experienced with JS than me since then, so it's possible that won't work anymore. I'm not sure.

waylan:
I just took another look. You need to modify both lines 47 and 81 of `main.js`. Both lines test for `if (!window.Worker)`. Temporarily changing both to `if (true)` should allow you to test the fallback.

jimconner:
Thank you for steering me in the right direction once again Waylan - I'll spend some time on this tomorrow to see if I can get it across the line.

jimconner:
I've given the alternate code path a test this morning (having changed both to `if (true)` and everything appears to work correctly without requiring any further changes, so I believe that this PR is properly ready for review now. Please let me know of any required changes and I'll (try to) take care of them. Thanks.

waylan:
Looks good to me. This can be merged after a note is added to the release notes. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jimconner:
Is adding the note to the release notes something that I need to do? I took a quick look just now but it wasn't obvious where to put in items for upcoming releases.

waylan:
Yes, you can add a one-liner here which pints to the original issue.:

https://github.com/mkdocs/mkdocs/blob/653ba22230bc318f9e240f9cf5248c2f41fff410/docs/about/release-notes.md#L24-L27

Note that that was added in a few commits after you started working on this. You may need to rebase with upstream to see those lines in the file.

jimconner:
Rebased against upstream/master and added comment into release notes as requested.

waylan:
The note in the release notes needs to point to the original issue (#2014). Like this:

    Add support for `min_search_length` parameter for search plugin (#2014).

I would just made the change myself, but it appears you have disabled the ability for us to make changes to your PR and I'm not particularly interesting in forking it and creating another PR.

As an aside, I realize the other two items listed in the current release don't point to issues, but no issues had been opened in each case. This however, does have an issue, so we need to point to it. You can find plenty of examples in previous releases.

jimconner:
No idea why the permissions to edit PRs are disabled - Someone else must have set that on our gituib org. I've just force pushed  the fixed release notes with the change that was required.

Thank you so much for all your help and patience getting me across the line with this. Cheers!!

PR #2022, opened 2020-03-11
 Disallow 0.0.0.0 dev_addr.

waylan:
MkDocs has never supported using the included server for production.
However, some users have been using it that way. This would prevent that.

Still needs proper tests. However, a quick test on the command line produces these results:

```
λ mkdocs serve -a 0.0.0.0:8000
INFO    -  Building documentation...
ERROR   -  Config value: 'dev_addr'. Error: The MkDocs' server is intended for 
development purposes only. Therefore, '0.0.0.0' is not a supported IP address.
Please use a third party production-ready server instead.

Aborted with 1 Configuration Errors!
λ mkdocs serve -a 277.0.0.0:8000
INFO    -  Building documentation...
ERROR   -  Config value: 'dev_addr'. Error: '277.0.0.0' does not appear to be an
IPv4 or IPv6 address

Aborted with 1 Configuration Errors!
```

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -21,6 +21,11 @@ The current and past members of the MkDocs team.
 * [@d0ugal](https://github.com/d0ugal/)
 * [@waylan](https://github.com/waylan/)
 
+## Version 1.1.1 (in development)
+
+Bugfix: Ensure wheel is Python 3 only.
+Bugfix: Clean up `dev_addr` validation and disallow `0.0.0.0`.
+
 ## Version 1.1 (2020-02-22)
 
 ### Major Additions to Version 1.1
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -1,6 +1,7 @@
 import os
 from collections import Sequence, namedtuple
 from urllib.parse import urlparse
+import ipaddress
 import markdown
 
 from mkdocs import utils, theme, plugins
@@ -232,6 +233,11 @@ class IpAddress(OptionallyRequired):
             host, port = value.rsplit(':', 1)
         except Exception:
             raise ValidationError("Must be a string of format 'IP:PORT'")
+
+        try:
+            ipaddress.ip_address(host)
+        except ValueError as e:
+            raise ValidationError(e)
 
         try:
             port = int(port)
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -244,6 +244,15 @@ class IpAddress(OptionallyRequired):
 
         return Address(host, port)
 
+    def post_validation(self, config, key_name):
+        host = config[key_name].host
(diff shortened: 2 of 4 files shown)


waylan:
How nice. We apparently have been using invalid IP addresses in our tests. Well, now we can clean that up too.

waylan:
So `ipaddress_ip_address` must not include the square brackets when validating an IPv6 address. However, the user should supply them as are required in a URL (which includes a port). Need to add some additional logic for that.

waylan:
This is getting weirder. 

```
$ mkdocs serve -a [::1]:8000
INFO    -  Building documentation...
INFO    -  Documentation built in 1.22 seconds
INFO    -  Serving on http://[::1]:8000

[Errno 8] nodename nor servname provided, or not known
$ mkdocs serve -a ::1:8000
INFO    -  Building documentation...
INFO    -  Documentation built in 1.18 seconds
INFO    -  Serving on http://::1:8000
INFO    -  Start watching changes
INFO    -  Start detecting changes
```

`[::1]:8000` causes the server to error out, but `::1:8000` works fine. However, with the server running at `::1:8000`, from the browser, `http://[::1]:8000` works, but `http://::1:8000` fails.

hason:
@waylan This change complicated local development with docker containers. Why do you care what IP address user can get there? It's user responsibility and decision.

dschmidt:
> @waylan This change complicates local development with docker containers. Why do you care what IP address user can get there? It's user responsibility and decision.

Complicates? Do you have a workaround already?

waylan:
> Why do you care what IP address user can get there?

As has been explained many times, we do not support using our server for anything except on the local device. Using `0.0.0.0` allows people to use the server outside of what we support and leads to a constant stream of support requests to for help in addressing problems with their setup. By disallowing `0.0.0.0` and displaying a message explaining it is not supported, the idea was that this would lessen the misguided requests from users.

cirego:
> @waylan This change complicates local development with docker containers. Why do you care what IP address user can get there? It's user responsibility and decision.

I am also a user negatively impacted by this decision because I run `mkdocs serve` inside of Docker containers for my local development workflow. This simplifies our workflow by reducing the number of dependencies on the developer system. I then have a wrapper script around `docker run` that uses `--mount` to expose my source files into the container, which the server will then see changes live. I get the best of both worlds: edit the files on my local machine while using Docker for managing dependencies.

The only workaround that I have right now is to use host mode (`--network host`) when running the server. The problem with this solution is that:

1. The container now has the ability to query services running on my local host and I have lost network isolation for the code running in the container.
2. I cannot remap ports from inside the container to be different ports outside my container. This prevents, for example, the ability to run multiple development servers on the same host.

For production, I have a workflow that builds the static site and hosts via Nginx.

cirego:
Of course, the other obvious workaround is `pip install mkdocs==1.1`.

waylan:
The conversation regarding this change is in #2108. Please follow up there.

PR #2029, opened 2020-03-21
fix #2027 code colors

### LINKED ISSUE #2027: readthedocs theme: .rst-content code color defined twice, all code is black
After upgrading from mkdocs 1.0.4 to 1.1 all code blocks (including those inline) became black instead of still having the well known orange color. Browser inspector shows there are duplicated rules for `.rst-content code` color - `#000` and `#E74C3C`. Is this is a bug? In my opinion previous colors looked much better and online readthedocs documents seem to still use orange.

wodny:
Some CSS overrides. See images for Sphinx, mkdocs 1.0.4 and mkdocs 1.1 (without/with PR) comparison.

## sphinx-2.4.4+rtd-0.4.3

sphinx-2 4 4+rtd-0 4 3

## mkdocs-1.0.4

mkdocs-1 0 4

## mkdocs-1.1

mkdocs-1 1

## mkdocs-1.1+PR

mkdocs-1 1+PR

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -26,6 +26,7 @@ The current and past members of the MkDocs team.
 Bugfix: Ensure wheel is Python 3 only.
 Bugfix: Clean up `dev_addr` validation and disallow `0.0.0.0`.
 Add support for `min_search_length` parameter for search plugin (#2014).
+Bugfix: Code colors (#2027).
 
 ## Version 1.1 (2020-02-22)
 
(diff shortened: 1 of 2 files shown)


waylan:
Look good. I'll merge this if you could add a comment to the release notes, here:

https://github.com/mkdocs/mkdocs/blob/2505a907f4ef6cc4492031faef06192f9eddf0bc/docs/about/release-notes.md#L24-L28

PR #2030, opened 2020-03-25
3.8 is supported; 2.7 and 3.4 are not anymore

ScriptAutomate:
Updated installation information to include Python 3.8 support, and dropped 2.7/3.4 support.

- 2.7 is EOL and no longer supported by MkDocs: #1936 
- 3.4 is EOL and no longer supported by MkDocs: #1938 
- 3.8 supported by MkDocs: #1938


waylan:
Good catch. Thanks.

PR #2053, opened 2020-04-07
Adds --page-type option when deploying using gh-deploy

mfjackson:
I recently ran into an issue trying to deploy my docs to a personal (i.e. user) site with the `gh-deploy` command where the docs would deploy to the URL `http://<user>.github.io/<user>.github.io`. After looking at the [mkdocs docs](https://www.mkdocs.org/user-guide/deploying-your-docs/#organization-and-user-pages) surrounding deployment I found two things: (1) this was expected behavior because I was essentially following the "project" deployment steps and (2) the docs and UI for deploying user and organization pages was confusing.

As a result, I created a `--page-type` option that can be used in a straightforward fashion during the deployment stage, without having to worry about changing directories or having two local repos. This is accompanied by an update to the "Organization and User Pages" section in the documentation and has been tested using `tox`.

Please let me know if you see any issues with implementing this. And thank you so much for continuing to maintain this great library!

### DIFF
--- a/docs/user-guide/deploying-your-docs.md
+++ b/docs/user-guide/deploying-your-docs.md
@@ -43,34 +43,23 @@ files locally.
 
 ### Organization and User Pages
 
-User and Organization Pages sites are not tied to a specific project, and the
-site files are deployed to the `master` branch in a dedicated repository named
-with the GitHub account name. Therefore, you need working copies of two
-repositories on our local system. For example, consider the following file
-structure:
+GitHub Pages also allows users to create [User and Organization Pages sites](https://help.github.com/en/github/working-with-github-pages/about-github-pages#types-of-github-pages-sites)
+that are not tied to a specific project. Rather, the site files are deployed to the
+`master` branch in a dedicated repository named with the GitHub account name,
+i.e. `<user>.github.io` or `<org>.github.io`.
 
-```no-highlight
-my-project/
-    mkdocs.yml
-    docs/
-orgname.github.io/
-```
-
-After making and verifying updates to your project you need to change
-directories to the `orgname.github.io` repository and call the
-`mkdocs gh-deploy` command from there:
+After making and verifying updates to a dedicated user or organization repository on
+its `master` branch, call `mkdocs gh-deploy` in the command line and specify the site
+type with the `--page-type` option. Inputs for this option include 'user', 'org', or
+'project'. For example, to create a user-specific site enter:
 
 ```sh
-cd ../orgname.github.io/
-mkdocs gh-deploy --config-file ../my-project/mkdocs.yml --remote-branch master
+mkdocs gh-deploy --remote-branch master --page-type user
 ```
 
-Note that you need to explicitly point to the `mkdocs.yml` configuration file as
-it is no longer in the current working directory. You also need to inform the
-deploy script to commit to the `master` branch. You may override the default
-with the [remote_branch] configuration setting, but if you forget to change
-directories before running the deploy script, it will commit to the `master`
-branch of your project, which you probably don't want.
+Note that you need to inform the deploy script to commit to the `master` branch rather
+than the default `gh-pages` branch. The command above should result in your docs being
+deployed to `https://<user.github.io>`.
 
 Be aware that you will not be able to review the built site before it is pushed
 to GitHub. Therefore, you may want to verify any changes you make to the docs
--- a/docs/user-guide/deploying-your-docs.md
+++ b/docs/user-guide/deploying-your-docs.md
@@ -43,23 +43,24 @@ files locally.
 
 ### Organization and User Pages
 
-GitHub Pages also allows users to create [User and Organization Pages sites](https://help.github.com/en/github/working-with-github-pages/about-github-pages#types-of-github-pages-sites)
-that are not tied to a specific project. Rather, the site files are deployed to the
-`master` branch in a dedicated repository named with the GitHub account name,
-i.e. `<user>.github.io` or `<org>.github.io`.
+GitHub Pages also allows users to create User and Organization Pages sites
+that are not tied to a specific project. Rather, the site files are deployed to
+the `master` branch in a dedicated repository named with the GitHub account
+name, i.e. `<user>.github.io` or `<org>.github.io`.
 
-After making and verifying updates to a dedicated user or organization repository on
-its `master` branch, call `mkdocs gh-deploy` in the command line and specify the site
-type with the `--page-type` option. Inputs for this option include 'user', 'org', or
-'project'. For example, to create a user-specific site enter:
+After making and verifying updates to a dedicated user or organization
+repository on its `master` branch, call `mkdocs gh-deploy` in the
+command line and specify the site type with the `--page-type` option.
+Inputs for this option include 'user', 'org', or'project'. For example,
+to create a user-specific site enter:
 
 ```sh
 mkdocs gh-deploy --remote-branch master --page-type user
 ```
 
-Note that you need to inform the deploy script to commit to the `master` branch rather
-than the default `gh-pages` branch. The command above should result in your docs being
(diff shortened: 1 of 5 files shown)


docs/user-guide/deploying-your-docs.md
```diff
@@ -43,34 +43,24 @@ files locally.
...
+After making and verifying updates to a dedicated user or organization
+repository on its `master` branch, call `mkdocs gh-deploy` in the
+command line and specify the site type with the `--page-type` option.
+Inputs for this option include 'user', 'org', or'project'. For example,
+to create a user-specific site enter:
 
 ```sh
-cd ../orgname.github.io/
-mkdocs gh-deploy --config-file ../my-project/mkdocs.yml --remote-branch master
+mkdocs gh-deploy --remote-branch master --page-type user
```
waylan, line 58: If you're not in the working tree of the `user` or `org` repo, then how does it know which repo to push to? I don't see where that is defined or explained anywhere.

mkdocs/__main__.py
```diff
@@ -18,6 +18,11 @@
 
 log = logging.getLogger(__name__)
 
+# Disable the warning that Click displays (as of Click version 5.0) when users
+# use unicode_literals in Python 2.
+# See http://click.pocoo.org/dev/python3/#unicode-literals for more details.
+click.disable_unicode_literals_warning = True
+
```
waylan, line 25: You appear to not be working with the latest in master. These lines were removed in a recent commit (e092584c). They should not be here. I would recommend rebasing against the latest in master.

mkdocs/commands/gh_deploy.py
```diff
@@ -126,9 +127,15 @@ def gh_deploy(config, message=None, force=False, ignore_version=False):
         if host is None:
             # This could be a GitHub Enterprise deployment.
             log.info('Your documentation should be available shortly.')
+        if page_type == 'user' or page_type == "org":
+            username, repo = path.split('/', 1)
```
waylan, line 131: A bunch of tests at `tests.gh_deploy_tests` are failing because `path` can be `None` here. You need to account for that.

docs/user-guide/deploying-your-docs.md
```diff
@@ -43,34 +43,24 @@ files locally.
...
 
-Note that you need to explicitly point to the `mkdocs.yml` configuration file as
-it is no longer in the current working directory. You also need to inform the
-deploy script to commit to the `master` branch. You may override the default
-with the [remote_branch] configuration setting, but if you forget to change
-directories before running the deploy script, it will commit to the `master`
-branch of your project, which you probably don't want.
+Note that you need to inform the deploy script to commit to the `master` branch
+rather than the default `gh-pages` branch. The command above should result in
+your docs being deployed to `https://<user.github.io>`.
```
waylan, line 63: I'm wondering if it would make sense for the default branch to be `master` when `page-type` is `user` or `org` (but remain `gh-pages` for `project`). Of course, that would make things more complicated, but it would be less error prone for users.

mkdocs/__main__.py
```diff
@@ -162,17 +168,21 @@ def build_command(clean, **kwargs):
...
-def gh_deploy_command(clean, message, remote_branch, remote_name, force, ignore_version, **kwargs):
+def gh_deploy_command(clean, message, remote_branch, remote_name, page_type, force, ignore_version, **kwargs):
     """Deploy your documentation to GitHub Pages"""
     try:
         cfg = config.load_config(
             remote_branch=remote_branch,
             remote_name=remote_name,
+            page_type=page_type,
             **kwargs
         )
```
waylan, line 187: Because you added the `page_type` argument here, a bunch of tests in `tests.cli_tests` are failing. They are all mock tests which verify that `load_config` was called with specific arguments. Each test needs to be updated by having the `page_type` argument added to it.

mkdocs/commands/gh_deploy.py
```diff
@@ -126,9 +127,15 @@ def gh_deploy(config, message=None, force=False, ignore_version=False):
...
+                repo = repo[:-len('.git')]
+            url = 'https://{}.github.io/'.format(username)
+            log.info('Your documentation should shortly be available at: ' + url)
         else:
             username, repo = path.split('/', 1)
             if repo.endswith('.git'):
                 repo = repo[:-len('.git')]
             url = 'https://{}.github.io/{}/'.format(username, repo)
-            log.info('Your documentation should shortly be available at: ' + url)
+            log.info('Your documentation should shortly be available at: ' + url)
```
waylan, line 141: For some reason the newline at the end of this file was removed, which is resulting in a flake8 error.

mfjackson:
@waylan thanks for the quick and thorough response! I'll try to flush out the documentation a bit more and work on making the appropriate fixes.

As for the idea, I think the implementation is certainly up for debate, but in terms of user friendliness, I feel like it helps to define clear roles from a user perspective (i.e. am I making docs for myself, an org, or a project?) when designing the CLI.

mfjackson:
After thinking about the design a bit more, the two repository design still seems like the way to go for `user` and `org` pages. I'm going to update the docs accordingly. However, I believe this PR improves the `gh-deploy` process in three ways:

1. It allows users and organizations to publish to https://<user>.github.io rather than https://<user>.github.io/<user_docs_repo>.
2. By @waylan's suggestion, it automatically deploys user and organization pages from the `master` branch, as is required by Github.
3. It let's the user explicitly state what type of project they are deploying while keeping the default user (`project`) and deployment behavior unchanged.

waylan:
First of all, this needs to be rebased against master before it can be merged. However, I'm not sure it will be anyway.

The only benefit this provides is that instead of typing `--remote-branch master` I can type `--page-type org`. That's not an improvement at all. The only help that potentially provides is for new users who don't understand how GitHub Pages works. But even then, it still requires switching to a separate directory. I'd rather just leave this as-is.

The problem is that `page-type` introduces some `magic` whereas explicitly setting `remote-branch` makes it clear what I am doing as a user. 

Now if the Pages repo and the "type" were defined in settings and a call to the deploy command from the project directory (without any options) resulted in a deployment directly to the Pages repo, then that would be interesting (this is also what I thought you were attempting to do).

mkdocs/commands/gh_deploy.py
```diff
@@ -126,6 +132,10 @@ def gh_deploy(config, message=None, force=False, ignore_version=False):
         if host is None:
             # This could be a GitHub Enterprise deployment.
             log.info('Your documentation should be available shortly.')
+        elif page_type == 'user' or page_type == 'org':
+            username, repo = path.split('/', 1)
+            url = 'https://{}.github.io/'.format(username)
+            log.info('Your documentation should shortly be available at: ' + url)
```
mfjackson, line 138: @waylan I understand what you're saying and I'll close the PR if it won't be merged. However, I believe these lines add value to `gh-deploy`, as I couldn't seem to get my own personal pages to deploy without being deployed to the URL https://mfjackson.github.io/mfjackson.github.io, which is not what I wanted. Is this the expected behavior or was I doing something wrong on my end?

I instead wanted them to be deployed to the URL https://mfjackson.github.io, which is what the `--page-type` option allows a user to specify. Initially I thought you could do this from the project repo, but I couldn't find a way to implement it without nuking the markdown files themselves.

waylan: If that's all you're concerned about, then all we really need to do is fix the bug where the url is generated. But that url is only used in a message output to the user. It has no bearing on the actual deployment. And that's my point. This is a major change in the user facing interface only to fix a bug in an `info` message. Seems like more than is needed.

Now, if you can demonstrate an actual issue with the deployment. Show that without this change, things get deployed to GitHub incorrectly. If you can demonstrate that, then my view may be different.

oprypin:
I also think this should not be merged.

> The only benefit this provides is that instead of typing `--remote-branch master` I can type `--page-type org`

PR #2081, opened 2020-04-25
Use README.md as index.html also if use_directory_urls is false

oprypin:
Right now, just like any *foo/index.md*, any *foo/README.md* becomes *foo/index.html*, but that works only if `use_directory_urls` is on.
But the `use_directory_urls` feature is for changing *foo/bar.md* into *foo/bar/index.html* and should not prevent the former from working.

I'm hoping that this is more of an oversight (due to the source code for the two things being very related) than an intentional exclusion.
So I'm sending a small pull request to fix that and also unify the code paths. But feel free to discuss this as an issue in case it's not as agreeable as I thought.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -23,6 +23,7 @@ The current and past members of the MkDocs team.
 
 ## Version 1.1.1 (in development)
 
+* Bugfix: Use README.md as index.html even if use_directory_urls is false.
 * Bugfix: Ignore links which start with a backslash (#1680).
 * Bugfix: Pass `builder` to the `on_serve` event so that it can be passed to
   `server.watch` by plugins (#1952).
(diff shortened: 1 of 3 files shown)


oprypin:
Best viewed [with "hide whitespace changes"](https://github.com/mkdocs/mkdocs/pull/2081/files?diff=split&w=1)

waylan:
I just searched through the old discussions about this and can't find any reason for the current behavior of only working when `use_directory_urls` is `True`. Additionally, the documentation makes no suggestion of a difference in behavior here. Seems like this was an oversight. Thanks for the report and the work.

waylan:
One thing is missing. We need a note added to the release notes. Here: 

https://github.com/mkdocs/mkdocs/blob/3bada392d2430e31edd55d67b59d60899c22ae11/docs/about/release-notes.md#L24-L33

oprypin:
That is done BTW.

Perhaps you could mention that the convention is to add the release notes line as the first item in the list under "*(in development)*". That is not trivial to deduce :)
This detail could also be added [into CONTRIBUTING.md](https://www.mkdocs.org/about/contributing/#submitting-pull-requests).

PR #2089, opened 2020-05-01
Ignore IntelliJ IDE conf settings

pawamoy:


waylan:
Tool specific items like this should go in the user's [global][1] `.gitignore` file, not in the project file. Not everyone uses the same IDE and there are more IDEs that we could ever account for. In fact, if you include the lines specific to your Idea in your global file, then you only need to do so once rather than in every individual project.

[1]: https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

pawamoy:
Oh great, didn't know that, thanks and sorry for the noise

PR #2095, opened 2020-05-06
Fix markdownlint test

waylan:
Just experimenting trying to understand why the markdownlint test is failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
     # Default Python
     - env: TOXENV=flake8
     - env: TOXENV=markdown-lint
-      language: node_js
+      #language: node_js
       node_js:
         - 12
       before_install: npm install -g markdownlint-cli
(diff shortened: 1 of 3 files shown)


waylan:
So Travis is using Node.js version v8.9.1, but `markdownlint-cli` dropped support for version 8 in 0.22.0 (see igorshubovych/markdownlint-cli#93). Apparently, Node doesn't prevent itself from installing incompatible versions as markdownlint-cli version 0.23.0 is installed.

We could force an older version with `npm install -g markdownlint-cli@0.21.0` but I'd rather using a newer version of Node. The Travis [docs][1] only list version 12 through 14, so I don't know why we are getting 8.

[1]: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions

waylan:
Sigh. So if I set the language to node.js I can get Travis to use a more recent version, but then we run into weird Python errors and we need Python to run tox. I guess I could run the test directly from Travis rather than using tox, but that would be inconsistent with the rest of our tests.

PR #2100, opened 2020-05-09
Make compressed sitemap deterministic

ofek:
Our deployment step for https://datadoghq.dev/integrations-core/ should only trigger when a change occurs in the built site. However, every build was different because gzip includes the timestamp in its header.

This fixes it by setting the field to zero.

cc @waylan since you were involved with https://github.com/mkdocs/mkdocs/pull/1130

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -23,6 +23,8 @@ The current and past members of the MkDocs team.
 
 ## Version 1.1.1 (in development)
 
+* Bugfix: Allow compressed sitemap to be deterministic by supporting the
+  `SOURCE_DATE_EPOCH` environment variable (#2100).
 * Bugfix: Use README.md as index.html even if use_directory_urls is false (#2081).
 * Bugfix: Ignore links which start with a backslash (#1680).
 * Bugfix: Pass `builder` to the `on_serve` event so that it can be passed to
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -1,5 +1,3 @@
-from datetime import datetime
-from calendar import timegm
 import logging
 import os
 import gzip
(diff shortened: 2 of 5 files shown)


waylan:
Since #1010 (4 years ago) we support the `SOURCE_DATE_EPOCH` environment variable for this very purpose. The value of that variable is used to set the [build_date_utc][1] variable in the template context. It seems to me that the same should be used for the gzip timestamp. After all, depending in which theme you use, you may need it set for deterministic builds anyway.

That said, I'm curious if perhaps the gzip lib already makes use of the env variable. After all, our users who use the variable have not complained about the sitemap.

Finally, I see that the docs for the `build_date_utc` template context variable does not make mention of `SOURCE_DATE_EPOCH`. That should probably be updated.

[1]: https://www.mkdocs.org/user-guide/custom-themes/#mkdocs_version

ofek:
Interesting. Where does that date get rendered? Removing the archive makes our builds reproducible without setting that environment variable.

waylan:
In the builtin themes, the `build_date_utc` gets included in an HTML comment on the homepage only.

`mkdocs` theme:

https://github.com/mkdocs/mkdocs/blob/a4eb4eb42be5b2b7f401f51baf8c0863f54fe63d/mkdocs/themes/mkdocs/base.html#L197-L204

`readthedocs` theme:

https://github.com/mkdocs/mkdocs/blob/a4eb4eb42be5b2b7f401f51baf8c0863f54fe63d/mkdocs/themes/readthedocs/base.html#L156-L163

ofek:
Ah I see, we minify the html so we have no comments.

ofek:
@waylan Addressed! FYI a new version of flake8 was released, hence the errors.

waylan:
> FYI a new version of flake8 was released, hence the errors.

Sigh. Thanks for the heads-up. We can ignore that here as it is unrelated to this change.

However, we do have a [failing test][1] in your code. 

Also, we need a note added to the release notes (under [version 1.1.1][2]).

Otherwise, this looks great. Thanks for the work on this.

[1]: https://travis-ci.org/github/mkdocs/mkdocs/jobs/685992944#L268
[2]: https://github.com/mkdocs/mkdocs/blob/master/docs/about/release-notes.md#version-111-in-development

ofek:
Done, thank you! Any idea when 1.1.1 will be released?

waylan:
> Any idea when 1.1.1 will be released?

I expect very soon.

waylan:
FYI, version 1.1.1 has been released.

ofek:
Thank you!!!

Issue #2082, opened 2020-04-26
Error reporting from plugins, and log.error


oprypin:
I've been wondering how to properly report warnings and errors that happen during operation of my plugin.

I found (not through documentation) that perhaps it should be this:

https://github.com/mkdocs/mkdocs/blob/44f3ae212d85a60750769df0e9fa8022274166c8/mkdocs/structure/files.py#L7-L11 https://github.com/mkdocs/mkdocs/blob/44f3ae212d85a60750769df0e9fa8022274166c8/mkdocs/structure/files.py#L241

(with the *very* subtle issue that mkdocs configures only `mkdocs.*` loggers but the plugin would not fall under that and would hence not follow the `--verbose` option and such.)

What's nice is that warnings automatically cause the generation to fail in MkDocs `strict` mode.

But I was surprised that `log.error` does *not* cause a failure **at all** even in `strict` mode (so I guess it looks for *warning* specifically), though I could argue that it should cause a failure even in non-strict mode. But I guess it's controversial to change the default behavior, and there's something to be said about errors causing an immediate crash versus being able to wait until the end and then cause a failure.

MkDocs itself seems to always employ a combination of `log.error` + `raise` (sometimes `raise SystemExit`). But I don't think a plugin can just `raise` anything like that, it'll cause a stack trace to show up, which MkDocs makes sure to avoid in its own handling, but a plugin can't prevent that, other than intentionally raising `SystemExit`, which is not great. It's *also* impossible to implement an error that should not be allowed to go unnoticed but also doesn't have to cause an immediate crash (config validation errors are the one example where that *does* work).

So to sum up, the use case of "log an error now and make sure to fail at the end" seems impossible to do from a plugin, and the use case of "log an error now and fail now without a stack trace" is kinda ugly.
And either way, error reporting for plugins is under-documented. All of the nuance above is certainly not obvious.

pawamoy:
I agree. 

At first I was using `mkdocs.utils.log` because I thought it was designed for this. In fact, it's just the utils logger. Each MkDocs module has its own. So I changed my code to get the loggers myself, with `log = logging.getLogger(__name__)`, and now my loggers are not initialized like MkDocs' loggers., and so do not follow the verbose option. I don't feel like duplicating initialization code in all plugins would be a good idea. So I will go back to `from mkdocs.utils import log`.

[...]

PR #2103, opened 2020-05-11
Add plugin error exception and event

pawamoy:
This PR adds a `PluginError` exception to the `mkdocs.exceptions` module to allow plugin writers to cleanly abort a build without showing a traceback.

It also adds an `on_build_error` event that is triggered by such `PluginError` exceptions, that we catch in the `build` command of the `mkdocs.commands.build` module (most of the diff in this module is just indentation in between the try/except, I could not reduce the noise). This event allows to clean things up if needed, just before aborting the build.

I wondered if I should not rename the event `on_plugin_error`, but to me it makes more sense to call it a build error, because it could be triggered in other places as well. Also, if you think the event doesn't have its place in this PR, I'll gladly remove it so we can discuss it a bit more before opening (or not) another PR for it.

Related: #2082 (the issue also asks for a logging error counter, as well as a `get_logger` utility function, so this PR does not resolve it entirely).

TODO:

- [x] add docs: I documented the `on_build_error` event, but not the `PluginError` since I couldn't find an existing page documenting MkDocs exceptions. It's mentioned in the `on_build_error` documentation though.
- [x] add tests (any advice?)


docs/user-guide/plugins.md
```diff
@@ -229,6 +229,15 @@ entire site.
     Parameters:
     : __config:__ global configuration object
 
+##### on_build_error
+
+:   The `build_error` event is called after an error happened during the build.
+    It is triggered by plugins by raising a `mkdocs.exceptions.PluginError`.
+    Use this event to clean things up before the build is aborted.
```
waylan, line 236: This is not exactly accurate. Saying that the "event is called after an error happened" implies that it is called immediately after. Instead it is called at the end of the build process. I realize that technically, the error short circuits the build process, but I think this could be a little clearer. 

For example: "the event is only called at the end of the build process when a `PluginError` has been raised during the build." I realize that's a little rough, but I think it is a better general direction.

pawamoy: But *it is* called immediately after :smile: 

Indeed, raising a `PluginError` will interrupt every scopes of execution until it is caught. We wrapped the whole `build` function's body into a `try`/`except PluginError`, and we immediately trigger the `build_error` event when such an exception is caught. The build process actually never ends as it is interrupted by the exception. Now it seems that you know that, based on your second comment about short-circuiting the build process, so I'm not sure what you first meant :confused: 

Maybe you meant that it is *physically* called at the end of the build process, as in, based on line numbers? But do plugin developers want/need to know that?

Lastly, your example would make me think that the error is somehow recorded, and that we check at the end of the build process if there was an error. But this is not what happens, as we do not record it, but rather catch it and immediately process it. It would also make me think that I, as a plugin developer, could raise multiple `PluginError`, letting the build continue, and have it exit with a code greater than 0. This would be incorrect again (except for the exit code).

I'm just not sure to understand what you mean :slightly_smiling_face:

pawamoy:
Thank you very much for your review! I will add the `Handling Errors` section to the documentation, and I'll happily let you suggest improvements over the text because English is not my native language and I'm not the best technical writer :slightly_smiling_face:

docs/user-guide/plugins.md
waylan: > Maybe you meant that it is physically called at the end of the build process,

Yes, that is what I mean. The way you worded it, that was not clear. I'm asking that that be made clear.

pawamoy: OK, I'll try to rephrase then :slightly_smiling_face:

pawamoy:
- I tried to be more clear in the description of the event.
- I added `BuildError`, and `PluginError` is a subclass of it
- Both types are caught with `except BuildError` in the build command
- I added a test in `build_tests.py`, because plugins tests were not actually "building" and this was required to check if the exception were correctly caught and handled

mkdocs/commands/build.py
```diff
@@ -232,75 +233,84 @@ def _build_page(page, config, files, nav, env, dirty=False):
...
-        raise SystemExit('\nExited with {} warnings in strict mode.'.format(utils.warning_filter.count))
+        log.info('Documentation built in %.2f seconds', time() - start)
 
-    log.info('Documentation built in %.2f seconds', time() - start)
+    except BuildError as error:
+        # Run `build_error` plugin events.
+        config['plugins'].run_event('build_error', error=error)
+        if error.reraise:
+            raise error
+        raise SystemExit('\nAborted build due to error.')
```
waylan, line 313: The purpose here is to hide the unhelpful (to the user) traceback, not to completely obscure the actual error. We should be passing on the error message to `SystemExit` like we do in the [build command][1].

[1]: https://github.com/mkdocs/mkdocs/blob/master/mkdocs/__main__.py#L153-L155

mkdocs/tests/build_tests.py
```diff
@@ -475,6 +477,55 @@ def test_copy_theme_files(self, site_dir, docs_dir):
         self.assertPathNotExists(site_dir, 'content.html')
         self.assertPathNotExists(site_dir, 'main.html')
 
+    def test_build_error_exception(self):
```
waylan, line 480: I understand that this test takes the form of similar tests in `build_tests.py`. However, it is a test related to the plugin API and that is where one would expect to find it. Lets move the test to `plugin_tests.py`.

mkdocs/commands/build.py
pawamoy: Done!

mkdocs/tests/build_tests.py
pawamoy: Sure, moved into `plugin_tests.py` :slightly_smiling_face:

pawamoy:
> Note, I have not carefully reviewed the documentation. I'll do a more thorough review and make my own edits when I have the time.

Of course, no problem :slightly_smiling_face:

waylan:
I reviewed the docs and made some adjustments in 6568cc4. @pawamoy what you had was a great start. Its always easier to work off of something than to start with nothing. 

Now that I've more familiar with the docs, I have a few thoughts:

1. Perhaps we should be catching `MkDocsException` instead, which is more inclusive than `BuildError`.
2. I keep wondering why we need the `reraise` argument. Just let a non-MkDocs specific error bubble up. Then I remember that only the MkDocs specific errors result in the `on_build_error` event being called. I wonder if we need to make that more clear in the docs.

waylan:
Also, if you see anything in the docs that I missed, please feel free to make the changes.

pawamoy:
Hey @waylan, your changes to the docs are great, thanks a lot :slightly_smiling_face: 

You raised two good points indeed.

> Perhaps we should be catching MkDocsException instead, which is more inclusive than BuildError.

I guess it makes sense to only catch `BuildError` because it happens in the `build` command. `ConfigurationError` are caught somewhere else for example. However if you feel like we should catch `MkDocsException`, I think it wouldn't hurt, so I'm fine with it.

> I keep wondering why we need the reraise argument. Just let a non-MkDocs specific error bubble up.

True.

> Then I remember that only the MkDocs specific errors result in the on_build_error event being called. I wonder if we need to make that more clear in the docs.

Yes we should tell that in the docs as it could be confusing ("error raised but event not called?"). Maybe we should call the event for any error then? And re-raise only if it's not a `BuildError` (subclass /instance /subclass instance) error.

Artnoc1:
> This PR adds a `PluginError` exception to the `mkdocs.exceptions` module to allow plugin writers to cleanly abort a build without showing a traceback.
> 
> It also adds an `on_build_error` event that is triggered by such `PluginError` exceptions, that we catch in the `build` command of the `mkdocs.commands.build` module (most of the diff in this module is just indentation in between the try/except, I could not reduce the noise). This event allows to clean things up if needed, just before aborting the build.
> 
> I wondered if I should not rename the event `on_plugin_error`, but to me it makes more sense to call it a build error, because it could be triggered in other places as well. Also, if you think the event doesn't have its place in this PR, I'll gladly remove it so we can discuss it a bit more before opening (or not) another PR for it.
> 
> Related: #2082 (the issue also asks for a logging error counter, as well as a `get_logger` utility function, so this PR does not resolve it entirely).
> 
> TODO:
> 
>     * [x]  add docs: I documented the `on_build_error` event, but not the `PluginError` since I couldn't find an existing page documenting MkDocs exceptions. It's mentioned in the `on_build_error` documentation though.
> 
>     * [x]  add tests (any advice?)

pawamoy:
Hello all. I implemented what was said in the previous comments and suggestions:

> Yes we should tell that in the docs as it could be confusing ("error raised but event not called?"). Maybe we should call the event for any error then? And re-raise only if it's not a BuildError (subclass / instance / subclass instance) error.

The `on_build_error` event is now triggered for any exception caught by MkDocs. If this exception is an instance of `BuildError` (true for instances of `PluginError` as well), then MkDocs will raise a SystemExit with the error message, and will not print any traceback. Any other exception is raised again.

The docs were updated accordingly (they were greatly simplified by this change!).

oprypin:
FYI the whitespace change can be hidden from the diff during review.
https://github.com/mkdocs/mkdocs/pull/2103/files?diff=unified&w=1

Thanks for the PR, I'm excited about it :)

pawamoy:
Hi @waylan, thank you for putting this in the 1.2 milestone :slightly_smiling_face: !

Is there anything I need to update (I see the label is still here)? Happy to help in any way.
/cc reviewers @Stanzilla @Artnoc1

waylan:
I just need to find the time to properly review this.

pawamoy:
Noted, thank you very much!

waylan:
I finally found some time to do a final review. This looks good! 👍 

@pawamoy if you rebase and resolve the the conflict, I'll merge this.

oprypin:
Shouldn't a merge commit be preferred over a rebase and force-push? So much easier to follow the review history then

waylan:
@oprypin you make a good point. In any event, the conflict needs to be resolved.

pawamoy:
Thank you for the review @waylan! I've resolved the conflicts, it should be good now.

waylan:
Oh sorry, I just realized this is missing an addition to the release notes.

pawamoy:
Should I take care of it? Not sure if I should add it in "Major additions" or "Other changes"?

waylan:
Yes, this is adding a new feature so we should highlight it in the Major additions section.

pawamoy:
Just added the release notes. As always, feel free to change the wording or even the whole paragraphs!

PR #2111, opened 2020-05-14
Normalize IP addresses and change error to warning

### LINKED ISSUE #2108: Change to prevent `0.0.0.0` makes containerized usage painful
The change in 1.1.1 to prevent the usage of `0.0.0.0` makes using MkDocs in a container painful. We have a local development environment used by our developers which spins up a documentation repository that they can live-edit, with MkDocs hosted inside a Docker container. This is now significantly more difficult.

waylan:
All IP addresses passed to `dev_addr` are now normalized.
`127.000.000.001` is normalized to `127.0.0.1`.

With apologies to docker container users, the address `0.0.0.0`
no longer raises an error but a warning instead. Apparently,
docker containers need to use that address, even in dev environments.

Closes #2108.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -21,6 +21,10 @@ The current and past members of the MkDocs team.
 * [@d0ugal](https://github.com/d0ugal/)
 * [@waylan](https://github.com/waylan/)
 
+## Version 1.1.2 (Under development)
+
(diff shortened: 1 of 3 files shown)


btisdall:
Thanks for this @waylan all your effort maintaining makedocs.

peng-cheng-li:
Not able run mkdoc server [::]:8000？

PR #2116, opened 2020-05-26
Fix relative links

MyrikLD:
It solves cases when the path starts from the project root. The file still must be in `docs_dir`, but links like `[image](../docs/image.png)` and `[image](docs/image.png)` now works. It makes it possible to use README.md files with relative links.

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -165,7 +165,7 @@ class Page:
         """
 
         extensions = [
-            _RelativePathExtension(self.file, files)
+            _RelativePathExtension(self.file, files, config)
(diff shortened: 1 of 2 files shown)


waylan:
So, this adds the ability to resolve links that are not proper relative links? Why would we do that?

oprypin:
Seems like the use case is having a file `README.md` which links to `docs/image.png` - correctly from its point of view, and *then* having a symlink `docs/README.md` &rarr; `../README.md`, in which `docs/image.png` becomes "wrong" - `docs/docs/image.png`.

However, this particular way of solving this would be called a hack, so I think this should not be merged.

PR #2193, opened 2020-09-22
Remove sidebar when printing

timvink:
This is without the sidebar:

image

This is the old view:

image


timvink:
Fixed the `csslint` issue & added a release note!

PR #2203, opened 2020-10-12
Fix _get_by_type for Section

lukasgeiter:
I noticed that the `_get_by_type` function doesn't work properly when used to find nested `Section` items. Since the `if` condition is true, it would not actually traverse the children of sections.

This is currently not causing any problems because MkDocs only uses the function with `Page` and `Link`. But I thought it would be a good idea to fix it anyway, in case that changes in the future.


waylan:
This needs a note added to the release notes as well as a test which fails before the change is made and passes after.

PR #2206, opened 2020-10-13
Allow mkdocs.yaml when '--config' is not passed

hwmrocker:
We wanted to give our colleagues to possibility to build docs automatically when they put a `mkdocs.yml` in the docs folder of their repository. But it happened often, that they used `.yaml` instead the `.yml`.

When this PR is accepted, it will look first look for `mkdocs.yml` when `--config` is not provided. If this file is not present, it will try to find `mkdocs.yaml`.
When both are not present, it behaves like before.
When both are present, it behaves like before.
If only the `.yaml` version it present, this one will be used as the config file.

### DIFF
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -143,7 +143,12 @@ def _open_config_file(config_file):
 
     # Default to the standard config filename.
     if config_file is None:
-        config_file = os.path.abspath('mkdocs.yml')
+        for possible_file_ending in ('yml', 'yaml'):
+            config_file = os.path.abspath('mkdocs.{}'.format(possible_file_ending))
+            if os.path.exists(config_file):
+                break
+        else:
+            config_file = os.path.abspath('mkdocs.yml')
 
     # If closed file descriptor, get file path to reopen later.
     if hasattr(config_file, 'closed') and config_file.closed:
--- a/mkdocs/tests/config/base_tests.py
+++ b/mkdocs/tests/config/base_tests.py
@@ -131,7 +131,6 @@ class ConfigBaseTests(unittest.TestCase):
             temp_dir.cleanup()
             os.chdir(old_dir)
 
-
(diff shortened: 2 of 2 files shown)


hwmrocker:
If you like this PR I would appreciate when you add the `hacktoberfest-accepted` lable to this PR. As you might know hacktoberfest is now opt-in only.

waylan:
Interesting patch. While I understand why some might want this, I'm not sure I like the idea. The file extension is clearly documented. So just use the correct extension. I'll have to think about this.

In any event, at a minimum, this would need a note added to the release notes.

hwmrocker:
Of course, I added a line to the release notes. Hope this is the correct way to do so.

waylan:
There is a conflict you need to resolve. You should have rebased against master prior to adding the note. Then you could have added the note to the new (under development) section of the release notes which has recently been added.

hwmrocker:
I resolved the conflict already.

OrkoHunter:
> The file extension is clearly documented. So just use the correct extension.

Agreed that the documentation is clear enough. :) The problem in our case (https://github.com/backstage/backstage) is that most of the other YAML files have `.yaml` extension. And so when users go ahead and create `mkdocs.yml`, they look at other existing YAML files in the project and use `mkdocs.yaml`. The problem often takes long time to debug. It is just a costly mistake, and `mkdocs` can help solve this by allowing both file extensions! 🙏

oprypin:
Thanks for the contribution.
I force-pushed only to drop the commits https://github.com/mkdocs/mkdocs/commit/26f8310343462fab43618519f3b7771a571e3f4f and https://github.com/mkdocs/mkdocs/commit/fa4c23e30a476d3c016590309f4fb4428adba93e.
I will merge this but only with edits in https://github.com/mkdocs/mkdocs/pull/2478
Feel free to review my edits.

PR #2209, opened 2020-10-18
Import ABC from collections.abc explicitly.

tirkarthi:
This fixes below warning for Python 3.10 compatibility. 

```
  /root/py39-venv/lib/python3.9/site-packages/mkdocs/config/config_options.py:2: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    from collections import Sequence, namedtuple
```

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -1,5 +1,6 @@
 import os
-from collections import Sequence, namedtuple
+from collections import namedtuple
+from collections.abc import Sequence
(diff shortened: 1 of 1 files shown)


graingert:
can we get a release of this on pypi.org? it's blocking uvicorn CI supporting python 3.10

waylan:
See #2440.

Issue #2104, opened 2020-05-12
Broken link in deploying documentation


seekM:
https://www.mkdocs.org/#deploying links to [GitHub project pages](https://help.github.com/articles/creating-project-pages-manually/) which doesn't work anymore. I'm not sure what the correct links is now -- maybe https://pages.github.com/?

waylan:
Thanks for the report. GitHub is constantly refactoring their docs. Its hard to maintain a good link.

KhushbuB:
I found [Creating a GitHub Pages site](https://docs.github.com/en/enterprise-server@2.22/github/working-with-github-pages/creating-a-github-pages-site) is the correct link.

PR #2210, opened 2020-10-19
Adding correct Github project links

KhushbuB:
Resolves issue #2104 
Correct Link : https://docs.github.com/en/enterprise-server@2.22/github/working-with-github-pages/creating-a-github-pages-site

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -357,7 +357,7 @@ the MkDocs IRC channel `#mkdocs` on freenode.
 [nav]: user-guide/configuration/#nav
 [discussion group]: https://groups.google.com/forum/#!forum/mkdocs
 [GitHub issues]: https://github.com/mkdocs/mkdocs/issues
-[GitHub project pages]: https://help.github.com/articles/creating-project-pages-manually/
+[GitHub project pages]: https://docs.github.com/en/enterprise-server@2.22/github/working-with-github-pages/creating-a-github-pages-site
 [pip]: https://pip.readthedocs.io/en/stable/installing/
 [Python]: https://www.python.org/
 [site_name]: user-guide/configuration/#site_name


waylan:
That link is specifically for GitHub's enterprise server. Shouldn't we be linking to the docs for [free pro team](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/creating-a-github-pages-site)?

FISHMANPET:
This is probably the correct link https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/creating-a-github-pages-site

ktomk:
Reported this upstream and it is already fixed with a redirect in place:

~~~
$ curl -Ls -o /dev/null -w '%{url_effective}\n' https://help.github.com/articles/creating-project-pages-manually/ 
https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/creating-a-github-pages-site
~~~

Ref: https://github.com/github/docs/discussions/1071

docs/index.md
```diff
@@ -357,7 +357,7 @@ the MkDocs IRC channel `#mkdocs` on freenode.
 [nav]: user-guide/configuration/#nav
 [discussion group]: https://groups.google.com/forum/#!forum/mkdocs
 [GitHub issues]: https://github.com/mkdocs/mkdocs/issues
-[GitHub project pages]: https://help.github.com/articles/creating-project-pages-manually/
+[GitHub project pages]: https://docs.github.com/en/enterprise-server@2.22/github/working-with-github-pages/creating-a-github-pages-site
```
waylan, line 360: Shouldn't this be changed to `https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/creating-a-github-pages-site`?

ktomk: From the conversation and the change-set upstream, my understanding is that the URI should be changed to:

* `https://docs.github.com/github/working-with-github-pages/creating-a-github-pages-site`

waylan:
I took a look at the content of the new page and it is not at all the same content as we linked to before. According to the wayback machine, most recently the link forwarded to [Creating Project Pages Using the Command Line](https://web.archive.org/web/20181112210755/https://help.github.com/articles/creating-project-pages-using-the-command-line/), which is nearly identical to the old page from when the link was first created:  [Creating Project Pages manually](https://web.archive.org/web/20160421004024/https://help.github.com/articles/creating-project-pages-manually/).

The content of the old page is not too long. I suspect it would make more sense to simply replicate the instructions from the old page in our own documentation. Of course, we should adapt it to be MkDocs specific and be careful not to just copy the old text.

ktomk:
Yes, good thinking to use wayback machine for this, didn't check that.

And I'm seeing that on the index it is giving pointers for hosting products (_GitHub Pages_ and _Amazon S3_), I'd suggest to to turn "GitHub project pages" into "GitHub Pages" (wording) as it looks like the correct name to me. It is also in use in Mkdocs docs already: https://www.mkdocs.org/user-guide/deploying-your-docs/#github-pages .

Instead of giving command line instructions about git (e.g. how to create a branch) an alternative that looks good to me is to link 
to the product page for GitHub Pages, of which I could gather two:

* "true" product page: https://pages.github.com/
* docs canonical: https://docs.github.com/github/working-with-github-pages/getting-started-with-github-pages

Can't say which of those two would be the better choice for Mkdocs visitors, maybe the one from the docs?

And when doing so, promoting Mkdocs built-in gh-pages command with a link to the "Deploying your Docs" page. So it's not all only to the outside when there is Mkdocs content about GitHub Pages support.

waylan:
I just realized this link is at the end of the "Getting Started" guide. I thought this was on the [deployment docs][1] page. That being the case, yes, a more generic link pointing to the product page (`https://pages.github.com/`) makes more sense. For details users can follow the link from there to [Deploying your Docs][1].

[1]: https://www.mkdocs.org/user-guide/deploying-your-docs/

PR #2219, opened 2020-10-30
show page nav if theres only one item

psy:
the mkdocs basic theme wont show nav items if there is only one defined.

### DIFF
--- a/mkdocs/themes/mkdocs/base.html
+++ b/mkdocs/themes/mkdocs/base.html
@@ -66,7 +66,7 @@
                 <a class="navbar-brand" href="{{ nav.homepage.url|url }}">{{ config.site_name }}</a>
                 {%- endblock %}
 
-                {%- if nav|length>1 or (page and (page.next_page or page.previous_page)) or config.repo_url %}
+                {%- if nav|length>0 or (page and (page.next_page or page.previous_page)) or config.repo_url %}
                 <!-- Expander button -->
                 <button type="button" class="navbar-toggler" data-toggle="collapse" data-target="#navbar-collapse">
                     <span class="navbar-toggler-icon"></span>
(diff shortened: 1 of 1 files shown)


waylan:
A single page document doesn't need a navigation menu as there are no other pages to link to. This was an intentional design decision. Why should we change it?

psy:
Ic, reasonable decision if there is only a single page. I'm trying to add a single external link to the navigation via mkdocs.yml, is there any other solution?

waylan:
If there are more than one entry in the navigation, then the length would be greater than 1 and your change would be unnecessary.

psy:
Yes, thats right, but i have only a single page and this one external link. My PR would fix my issue but raise another one so I'm going to close it. Any ideas how to fix my problem welcome ;)

waylan:
I don't see how you have a problem to fix. As the second example [here][1], you can include your external link right in the nav. That provides a nav of at least length two.

[1]: https://www.mkdocs.org/user-guide/configuration/#nav

psy:
Maybe it gets clearer if i provide an example: I have only a single page (index.md) and want to add a link. My mydocs.yml:

```
nav:
    - 'Foobar': 'https://example.com'
```

It wont be shown as long as I dont add a second link there.

waylan:
Yes, but you are expected to include your page in the nav as well:

```yaml
nav:
    - Home: index.md
    - 'Foobar': 'https://example.com'
```

If you don't want that, then you have an edge case that we can't support within reason. As a reminder, the primary purpose of the nav is to link to internal pages. The ability to include external links is an extra, not the man purpose. Therefore, having a nav of only external links is outside the scope of what we support.

However, you may be able to use [theme customization][1] to [override that block][2] of the template. No need to patch MkDocs.

Also, see how much more clear the information is I provided when I actually understood your use case. Next time, please provide that context up front. Don't expect us to guess at what you are actually trying to do. Personally, it never occurred to me that you would have a nav of only a single external link. I assumed you would always include the local page. And as you never expressly stated that the internal page was omitted, I never made the connection.

[1]: https://www.mkdocs.org/user-guide/styling-your-docs/#using-the-theme-custom_dir
[2]: https://www.mkdocs.org/user-guide/styling-your-docs/#overriding-template-blocks

PR #2222, opened 2020-11-04
Add page keywords, helpful for seo

Writeup007:
Add page keywords, I think it's helpful for [SEO](https://en.wikipedia.org/wiki/Search_engine_optimization) (Search Engine Optimization)

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -117,6 +117,12 @@ edit_uri: root/path/docs/
 `src/default/docs/` for a Bitbucket repo, if `repo_url` matches those domains,
 otherwise `null`
 
+### site_keywords
+
+Set the site keywords. This will add a meta tag to the generated HTML header.
+
+**default**: `null`
+
 ### site_description
 
 Set the site description. This will add a meta tag to the generated HTML header.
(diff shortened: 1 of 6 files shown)


ktomk:
I like the idea to have meta keywords support, that's helpful for HTML in general, not only SEO.

And I also like to have some central control, like in the config file.

However each site should have it's own say to add those (and then for example move the site-keywords behind the page keywords).

Isn't there an extension reading page keywords from the _frontmatter_ ?

coliff:
Meta keywords are no longer needed. No search engines have used them for many years. Search for 'meta keywords' and you'll see plenty of articles advising against using them. e.g.

Meta Keywords: Why You Shouldn't Use Them in 2020
https://www.clickminded.com/meta-keywords/

Meta keywords Why we don't use them - and neither should you
https://yoast.com/meta-keywords/

waylan:
@coliff thank you for doing the research on this. I was expecting to find the same thing, but hadn't found the time to look into it to confirm.  Given the results of that research, I'm closing this wontfix. If someone really wants this, they can create their own theme or customize/fork an existing one. A custom theme can always support custom config options under the [theme config][1], so MkDocs doesn't need to add support for a new config option.

In fact, if this PR was going to be accepted, I would have insisted that the new config option be a theme option, rather than a project option.

[1]: https://www.mkdocs.org/user-guide/custom-themes/#theme-configuration

ktomk:
It's not even 2020 but many years (!) earlier, however you may want to have it for other indexers that aren't blurred from googles and copycats (yahoo, bing, ....) thought. Still I see this in a frontmatter context and SEO most likely always is a close-direct keyword @waylan

PR #2230, opened 2020-11-17
Make clear the distinction between and use of `docs_dir` and `docs`

qmacro:
Wonderful starter guide, absolutely spot on, thank you! Please permit me to offer this minor typo fix. Thanks.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -168,7 +168,8 @@ Take a moment to review the initial project that has been created for you.
 ![The initial MkDocs layout](img/initial-layout.png)
 
 There's a single configuration file named `mkdocs.yml`, and a folder named
-`docs` that will contain your documentation source files. Right now the `docs`
+`docs` that will contain your documentation source files (`docs` is
+the default value for the `docs_dir` configuration setting). Right now the `docs`
 folder just contains a single documentation page, named `index.md`.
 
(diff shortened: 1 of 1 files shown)


waylan:
Thank you for your contribution. However, that is not a typo. While the default value of the [docs_dir][1] config setting is `docs`, we can't assume that users will not override that default. Therefore, we need to make reference to the directory specified in the `docs_dir` config setting, not the default value of that setting. If you want to suggest a wording that makes that more clear, then we are open to suggestions, but this suggested change is not that.

[1]: https://www.mkdocs.org/user-guide/configuration/#docs_dir

qmacro:
Understood. I'd been confused as this was the very first mention of `docs_dir`, whereas I'd seen `docs` used a couple of times already in the previous sections leading up to this point. Plus the language around the use of `docs_dir` ("_create an `img` subdirectory in your `docs_dir`_") suggested to me that it was a location (a directory) not a config setting. Hopefully this second commit will help. Cheers.

waylan:
Thanks @qmacro. I had missed that this is in the getting started guide. Of course, as we havn't instructed the user to change the setting, we can assume it will actually be `docs` here. Your edit certainly helps clear things up for the new user.

PR #2239, opened 2020-11-25
removed modernizr from readthedocs theme since it is not used

edoapra:
Not sure why Modernizr is shipped with the readocs theme.


oprypin:
The answer to "why" is that the upstream readthedocs theme includes it. *Maybe* if you can clearly show that the reason for the upstream theme to include this  doesn't apply to MkDocs, this could be considered.
Historically the answer to such inquiries is that MkDocs doesn't deviate from the upstream readthedocs theme (don't know if I agree so much with it though).

Issue #2252, opened 2020-12-13
Google Analytics v4 (gtag)


nemchik:
Google recently introduces Analytics v4.

ga4 no longer provides IDs compatible with `analytics.js` which are in the format `UA-#####-#`

ga4 now uses a new `gtag.js` and IDs in the format `G-ABC123`

Google no longer allows creation of properties with the older analytics IDs. Existing Google Analytics properties can be upgraded and converted to ga4 which will replace their `UA-#####-#` ID with a `G-ABC123` ID. There is no way to downgrade.

Google is emailing existing users encouraging them to upgrade to ga4. To my knowledge at this time they have not announced any kind of cut off date for when migrating to ga4 must be completed.

Currently it is possible to use template overrides to put the new ga4 into an mkdocs site, however the
```yaml
google_analytics:
  - UA-#####
  - auto
```
format is extremely convenient and easy to use (by comparison to the slightly more advanced setup of overrides).

For the purpose of making the transition possible without breaking existing setups without warning, I think it would be great to add
```yaml
gtag:
  - G-ABC123
```
options to `mkdocs.yml` that would use the new `gtag.js` which is documented by google to be implemented as
```html
<script async src="https://www.googletagmanager.com/gtag/js?id=G-ABC123"></script>
<script>
    window.dataLayer = window.dataLayer || [];
    function gtag() {
        dataLayer.push(arguments);
    }
    gtag('js', new Date());
    gtag('config', 'G-ABC123');
</script>
```

This would allow existing mkdocs users to continue using `google_analytics:` for as long as google continues to support it, or upgrade to `gtag:` very easily at their own discretion.

Alternatively it would also probably be file to implement something in `google_analytics:` handling that uses gtag depending on the ID specified.

waylan:
We have wanted to move analytics to a plugin for some time (which would make it easier for users to switch it out for another service of their choice). This would be a great opportunity to do that. As users would need to update their config to include the new `gtag` anyway, they can do so by enabling the plugin and providing the `gtag` as a config option on the plugin.

Personally, the time I have available to work on MkDocs is very limited these days, so any contributions would help this alone.

waylan:
I just noticed that there are some references to this issue from the material theme. Rather than moving analytics to a plugin, another possibility would be to make the new `gtag` config option a theme option (nested under the theme setting). This would give theme developers complete control over it and possibly allow users to support a different service through theme customization. 

[...]

PR #2253, opened 2020-12-14
Deprecate google_analytics config option.

waylan:
This implements the plan discussed in #2252.

On the built-in themes I added a new option `theme.analytics.gtag`. See the documentation for details. The idea is that multiple different services could be offered under `theme.analytics`. At this time only support for `gtag` exists, but more could be added in the future.

I made some changes and additions to the Deprecated option validator which still needs tests.

We also need to upgrade the Google Analytics account for https://mkdocs.org to use the new GA4 ID. Until we do, builds will issue a warning:

```text
$ mkdocs build
WARNING -  Config value: 'google_analytics'. Warning: The configuration option 
google_analytics has been deprecated and will be removed in a future release o
f MkDocs. See the options available on your theme for an alternative.
INFO    -  Building documentation to directory: /Users/waylan/Code/mkdocs/site
INFO    -  Documentation built in 0.99 seconds
```

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -36,6 +36,23 @@ mkdocs serve --wait 60
 
 ### Backward Incompatible Changes in 1.2
 
+The `google_analytics` configuration option is deprecated as Google appears to
+be fazing it out in favor of its new Google Analytics 4 property. See the
+documentation for your theme for alternatives which can be configured as part
+of your theme configuration. For exmaple, the [mkdocs][mkdocs-theme] and
+[readthedocs][rtd-theme] themes have each added a new `theme.analytics.gtag`
+configuration option which uses the new Google Analytics 4 property. See
+Google's documentation on how to [Upgrade to a Google Analytics 4
+property][ga4]. Then set  `theme.analytics.gtag` to the "G-" ID and delete the
+`google_analytics` configuration option which contains a "UA-" ID. So long
+as the old "UA-" ID and new "G-" ID are properly linked in your Google account,
+and you are using the "G-" ID, the data will be made available in both the old
+and new formats by Google Analytics. See #2252.
+
+[mkdocs-theme]: ../user-guide/styling-your-docs.md#mkdocs
+[rtd-theme]: ../user-guide/styling-your-docs.md#readthedocs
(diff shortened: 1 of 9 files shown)


squidfunk:
I haven't tested it but the proposal sound great. I also like the extensible approach got `theme.analytics.*` and will implement it in Material for MkDocs.

waylan:
The integration tests will continue to fail until we update our own config to use the new gtag.

waylan:
@tomchristie can we get the  Google Analytics account for https://mkdocs.org upgraded to V4 (as documented [here][1]). We then need the `G-ABC123` format ID to put in the config file.

```diff
diff --git a/mkdocs.yml b/mkdocs.yml
index f120128f..1aed065e 100644
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -8,6 +8,7 @@ edit_uri: ""

 theme:
     name: mkdocs
+    analytics: {gtag: 'G-ABC123'}
     highlightjs: true
     hljs_languages:
         - yaml
@@ -40,7 +41,6 @@ markdown_extensions:
         repo: mkdocs

 copyright: Copyright &copy; 2014 <a href="https://twitter.com/_tomchristie">Tom Christie</a>, Maintained by the <a href="/about/release-notes/#maintenance-team">MkDocs Team</a>.
-google_analytics: ['UA-27795084-5', 'mkdocs.org']

 plugins:
     - search
```

[1]: https://support.google.com/analytics/answer/9744165?hl=en

alex-klimov:
Hi guys, what is the status? Do you think this would be completed any time soon?

waylan:
@tomchristie can we upgrade the Google Analytics account as [discussed][1] above?

If not, then we will need to work out some other solution to avoid the tests failing for the build of our own documentation.

[1]: https://github.com/mkdocs/mkdocs/pull/2253#issuecomment-746891636

waylan:
@tomchristie sorry to pester you but this is the only issue blocking the next release. If we are not going to get this update, we need to know so that we can plan around it.

tomchristie:
The UI is a bit inscrutable, but.... is this what you need?

Screenshot 2021-06-02 at 10 33 20

waylan:
Yes, that resolves it. Thank you.

PR #2264, opened 2020-12-22
Don't re-query the list of doc pages for every page (for performance)

oprypin:
I.e. this list now gets formed a constant numer of times within a mkdocs build, instead of also doing it once per page

For compatibility (if someone happened to use this function externally), one can still pass a `Files` to get_context, but its main mode of operation is now to accept a direct list of `File`s

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -7,7 +7,7 @@ from jinja2.exceptions import TemplateNotFound
 import jinja2
 
 from mkdocs import utils
-from mkdocs.structure.files import get_files
+from mkdocs.structure.files import Files, get_files
 from mkdocs.structure.nav import get_navigation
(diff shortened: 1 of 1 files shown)


oprypin:
Profiling on a site with ~170 pages.

Before:

ncalls | tottime | percall | cumtime | percall | filename:lineno(function)
-- | -- | -- | -- | -- | --
**180** | 0.0003194 | 1.775e-06 | **0.1035** | 0.0005749 | files.py:44(documentation_pages)


After:

ncalls | tottime | percall | cumtime | percall | filename:lineno(function)
-- | -- | -- | -- | -- | --
**8** | 1.563e-05 | 1.954e-06 | **0.004359** | 0.0005449 | files.py:44(documentation_pages)

PR #2265, opened 2020-12-22
search: Merge the postprocessing done by two HTML parsers into one

oprypin:
This improves performance, as HTML of each page doesn't need to be parsed twice.

~Also replace list->string joining with StringIO, as that's very slightly faster and slightly more idiomatic.~

### DIFF
--- a/mkdocs/contrib/search/search_index.py
+++ b/mkdocs/contrib/search/search_index.py
@@ -1,4 +1,3 @@
-import io
 import os
 import re
 import json
@@ -69,7 +68,7 @@ class SearchIndex:
         # Create an entry for the full page.
         self._add_entry(
             title=page.title,
-            text=parser.stripped_html.getvalue().rstrip('\n'),
+            text=parser.stripped_html.rstrip('\n'),
             loc=url
         )
 
@@ -162,7 +161,7 @@ class ContentParser(HTMLParser):
         self.data = []
         self.section = None
         self.is_header_tag = False
-        self.stripped_html = io.StringIO()
+        self._stripped_html = []
 
     def handle_starttag(self, tag, attrs):
         """Called at the start of every HTML tag."""
@@ -195,7 +194,7 @@ class ContentParser(HTMLParser):
         Called for the text contents of each tag.
         """
 
-        self.stripped_html.write(data)
+        self._stripped_html.append(data)
 
         if self.section is None:
             # This means we have some content at the start of the
@@ -211,3 +210,7 @@ class ContentParser(HTMLParser):
             self.section.title = data
         else:
             self.section.text.append(data.rstrip('\n'))
+
+    @property
+    def stripped_html(self):
+        return '\n'.join(self._stripped_html)
--- a/mkdocs/contrib/search/search_index.py
+++ b/mkdocs/contrib/search/search_index.py
@@ -1,3 +1,4 @@
+import io
 import os
 import re
 import json
@@ -68,7 +69,7 @@ class SearchIndex:
         # Create an entry for the full page.
         self._add_entry(
             title=page.title,
-            text=self.strip_tags(page.content).rstrip('\n'),
+            text=parser.stripped_html.getvalue().rstrip('\n'),
             loc=url
         )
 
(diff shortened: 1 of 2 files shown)


oprypin:
Profile before/after on a site with ~320 pages

prof

waylan:
This changes the generated file `search_index.json`. According to jsondiff.com and jsondiff.org (I checked both), after the change, building mkdocs.org (the documentation for this project) results in 2 missing properties and 775 unequal values. I would expect the output to be the same before and after this change.

oprypin:
Oh, sorry about that. Thanks for checking. Let me take a look.

waylan:
Sorry, I just realized I was comparing builds from different points in history. Once I merged master into your branch, the differences were reduced to 10 unequal values which all appear to be related to insignificant whitespace.

waylan:
It appears the previously we were leaving an extra space between the last word on a line and the `\n`, while your implementation excludes that extra space. That seems like an improvement to me. Although, in the end, it shouldn't matter to the search engine.

oprypin:
I'm seeing more whitespace diffs on Pygments-highlighted areas.

```shell
colordiff -U0 <(gron site/search/search_index.json) <(gron site2/search/search_index.json)
```
```diff
-json.docs[267].text = "struct Athena::Routing::Arguments::Resolvers::Request inherits Struct Handles resolving a value for action arguments typed as HTTP::Request . @[ ART :: Get ( \"/\" ) ] def get_request_path ( request : HTTP :: Request ) : String request . path end Included modules Athena::Routing::Arguments::Resolvers::ArgumentValueResolverInterface Class methods .new Methods #resolve ( request : HTTP :: Request , argument : ART :: Arguments :: ArgumentMetadata ) Returns a value resolved from the provided request and argument . #supports? ( request : HTTP :: Request , argument : ART :: Arguments :: ArgumentMetadata ) : Bool Returns true if self is able to resolve a value from the provided request and argument .";                                                                                                                                                                    
+json.docs[267].text = "struct Athena::Routing::Arguments::Resolvers::Request inherits Struct Handles resolving a value for action arguments typed as HTTP::Request. @[ART::Get(\"/\")] def get_request_path(request : HTTP::Request) : String request.path end Included modules Athena::Routing::Arguments::Resolvers::ArgumentValueResolverInterface Class methods .new Methods #resolve(request : HTTP::Request, argument : ART::Arguments::ArgumentMetadata) Returns a value resolved from the provided request and argument. #supports?(request : HTTP::Request, argument : ART::Arguments::ArgumentMetadata) : Bool Returns true if self is able to resolve a value from the provided request and argument.";                                                                                                                                                                                                         
```
example from https://oprypin.github.io/athena-website/Routing/Arguments/Resolvers/Request/


This seems like even more of an improvement but still should check this in a bit more detail.

mkdocs/contrib/search/search_index.py
```diff
@@ -127,34 +128,6 @@ def generate_search_index(self):
...
-        self.data = []
-
-    def handle_data(self, d):
-        """
-        Called for the text contents of each tag.
-        """
-        self.data.append(d)
-
-    def get_data(self):
-        return '\n'.join(self.data)
```
oprypin, line 156: Right, I missed the fact that it used to be joined with `'\n'`, just assumed it's `''` somehow.

waylan:
So it appears that the previous implementation added a space between every HTML element. I suppose this ensured that two words always remained two words. After all, in certain edge cases a browser would render HTML with clear separation between elements even though there is no separation in the source. Extracting the text form those elements would loose that separation if no whitespace is added.

Given the above, I think we need to restore the previous behavior. We probably don't need it everywhere, but it is better to have it everywhere than not at all.

oprypin:
The main example I'd give here is something like

> *foo***bar**

Old index would see this as "foo bar" while the new one would see it as "foobar".

And the thing is, those spaces do actually make the user-visible output uglier.
If you have the text

> [Foo](#), [bar](#) and [baz](#)

when searching for it, the popup will show

> Foo , bar and baz



I'm pretty sure that the way that HTML works, it's always consistent: if there is no whitespace between two elements in the HTML, there will be no whitespace in the output, otherwise there will be one space (by default, but that can be configured to present the original).

So I think inserting the extra spaces is incorrect, and we would, in fact, be better off without them.
**But probably best to split that change out.**

waylan:
> I'm pretty sure that the way that HTML works, it's always consistent: if there is no whitespace between two elements in the HTML, there will be no whitespace in the output, otherwise there will be one space (by default, but that can be configured to present the original).

Consider this case:

```html
<p>foo</p><p>bar</p>
```

That should be either `foo bar` or `foo\nbar`, but not `foobar` in the index.

That said, I agree that `<strong>foo</strong>bar` should probably result in `foobar`.

oprypin:
Oh that's true. First I thought that in reality this can never happen (python-markdown can't output that), but one could put exactly that just into Markdown as HTML. And then what... to my surprise, it still somehow put `foo bar` into the result. Then I confirmed that indeed, something re-processes the HTML to be on separate lines somewhere along the way.

Anyway, I added a commit to drop this change now.

PR #2272, opened 2020-12-27
Optimize getting relative page URLs

oprypin:
This is a custom implementation that's ~5 times faster. That's important because calls to `normalize_url` on a site with ~300 pages currently take up ~20% of the total run time due to the sheer number of them. The number of calls is at least the number of pages squared.

The old implementation relied on `posixpath.relpath`, which (among other unnecessary parts) first expands every path with `abspath` for whatever reason, so actually it becomes even slower if you happen to be building under a deeply nested path.

### Profile before/after

profile

#### In context

profile

### DIFF
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -60,6 +60,60 @@ class UtilsTests(unittest.TestCase):
             is_html = utils.is_html_file(path)
             self.assertEqual(is_html, expected_result)
 
+    def test_get_relative_url(self):
+        expected_results = {
+            ('foo/bar', 'foo'): 'bar',
+            ('foo/bar.txt', 'foo'): 'bar.txt',
+            ('foo', 'foo/bar'): '..',
+            ('foo', 'foo/bar.txt'): '.',
+            ('foo/../../bar', '.'): 'bar',
+            ('foo/../../bar', 'foo'): '../bar',
+            ('foo//./bar/baz', 'foo/bar/baz'): '.',
+            ('a/b/.././../c', '.'): 'c',
+            ('a/b/c/d/ee', 'a/b/c/d/e'): '../ee',
+            ('a/b/c/d/ee', 'a/b/z/d/e'): '../../../c/d/ee',
+            ('foo', 'bar.'): 'foo',
+            ('foo', 'bar./'): '../foo',
+            ('foo', 'foo/bar./'): '..',
+            ('foo', 'foo/bar./.'): '..',
+            ('foo', 'foo/bar././'): '..',
+            ('foo/', 'foo/bar././'): '../',
+            ('foo', 'foo'): '.',
+            ('.foo', '.foo'): '.foo',
+            ('.foo/', '.foo'): '.foo/',
+            ('.foo', '.foo/'): '.',
+            ('.foo/', '.foo/'): './',
+            ('///', ''): './',
+            ('a///', ''): 'a/',
+            ('a///', 'a'): './',
+            ('.', 'here'): '..',
+            ('..', 'here'): '..',
+            ('../..', 'here'): '..',
+            ('../../a', 'here'): '../a',
+            ('..', 'here.txt'): '.',
+            ('a', ''): 'a',
+            ('a', '..'): 'a',
+            ('a', 'b'): '../a',
+            ('a', 'b/..'): '../a',  # The dots are considered a file. Documenting a long-standing bug.
+            ('a', 'b/../..'): 'a',
+            ('', ''): '.',
+            ('.', ''): '.',
+            ('', '.'): '.',
+            ('.', '.'): '.',
+            ('a/..../b', 'a/../b'): '../a/..../b',
+            ('a/я/b', 'a/я/c'): '../b',
+            ('a/я/b', 'a/яя/c'): '../../я/b',
+        }
+        # Leading slash intentionally ignored
+        for url_slash in ('', '/'):
+            for other_slash in ('', '/'):
+                for (url, other), expected_result in expected_results.items():
+                    # Acknowledge the only difference that a leading slash can cause:
+                    if url_slash + url == '/':
+                        expected_result += '/'
+                    relurl = utils.get_relative_url(url_slash + url, other_slash + other)
+                    self.assertEqual(relurl, expected_result)
+
     def test_create_media_urls(self):
 
         expected_results = {
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -255,13 +255,13 @@ def get_relative_url(url, other):
     """
     Return given url relative to other.
     """
-    other_parts = [p for p in other.split('/') if p not in ('.', '')]
-    dest_parts = [p for p in url.split('/') if p not in ('.', '')]
-
+    other_parts = other.split('/')
     # Remove filename from other url if it has one.
     if other_parts and '.' in other_parts[-1]:
         other_parts.pop()
 
+    other_parts = [p for p in other_parts if p not in ('.', '')]
+    dest_parts = [p for p in url.split('/') if p not in ('.', '')]
     common = 0
     for a, b in zip(other_parts, dest_parts):
         if a != b:
@@ -270,7 +270,6 @@ def get_relative_url(url, other):
 
     rel_parts = ['..'] * (len(other_parts) - common) + dest_parts[common:]
     relurl = '/'.join(rel_parts) or '.'
-
     return relurl + '/' if url.endswith('/') else relurl
 
 
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -251,9 +251,10 @@ def is_error_template(path):
     return bool(_ERROR_TEMPLATE_RE.match(path))
 
 
+@functools.lru_cache(maxsize=None)
 def _norm_parts(parts):
     dest_parts = []
-    for p in parts:
+    for p in parts.split('/'):
         if p not in ('.', ''):
             if p == '..':
                 if dest_parts:
@@ -267,14 +268,13 @@ def get_relative_url(url, other):
     """
     Return given url relative to other.
     """
-    other_parts = other.split('/')
     # Remove filename from other url if it has one.
-    if other_parts and '.' in other_parts[-1]:
-        other_parts.pop()
+    dirname, _, basename = other.rpartition('/')
+    if '.' in basename:
+        other = dirname
 
-    other_parts = _norm_parts(other_parts)
-    dest_parts = _norm_parts(url.split('/'))
-
+    other_parts = _norm_parts(other)
+    dest_parts = _norm_parts(url)
     common = 0
     for a, b in zip(other_parts, dest_parts):
         if a != b:
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -251,6 +251,18 @@ def is_error_template(path):
     return bool(_ERROR_TEMPLATE_RE.match(path))
 
 
+def _norm_parts(parts):
+    dest_parts = []
+    for p in parts:
+        if p not in ('.', ''):
+            if p == '..':
+                if dest_parts:
+                    dest_parts.pop()
+            else:
+                dest_parts.append(p)
+    return dest_parts
+
+
 def get_relative_url(url, other):
     """
     Return given url relative to other.
@@ -260,8 +272,9 @@ def get_relative_url(url, other):
     if other_parts and '.' in other_parts[-1]:
         other_parts.pop()
 
-    other_parts = [p for p in other_parts if p not in ('.', '')]
-    dest_parts = [p for p in url.split('/') if p not in ('.', '')]
+    other_parts = _norm_parts(other_parts)
+    dest_parts = _norm_parts(url.split('/'))
+
     common = 0
     for a, b in zip(other_parts, dest_parts):
         if a != b:
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -255,11 +255,22 @@ def get_relative_url(url, other):
     """
     Return given url relative to other.
     """
-    if other != '.':
-        # Remove filename from other url if it has one.
-        parts = posixpath.split(other)
-        other = parts[0] if '.' in parts[1] else other
-    relurl = posixpath.relpath(url, other)
+    other_parts = [p for p in other.split('/') if p not in ('.', '')]
+    dest_parts = [p for p in url.split('/') if p not in ('.', '')]
+
+    # Remove filename from other url if it has one.
+    if other_parts and '.' in other_parts[-1]:
+        other_parts.pop()
+
+    common = 0
+    for a, b in zip(other_parts, dest_parts):
+        if a != b:
+            break
+        common += 1
+
+    rel_parts = ['..'] * (len(other_parts) - common) + dest_parts[common:]
+    relurl = '/'.join(rel_parts) or '.'
+
     return relurl + '/' if url.endswith('/') else relurl
 
 
@@ -269,8 +280,8 @@ def normalize_url(path, page=None, base=''):
     if is_abs:
         return path
     if page is not None:
-        base = page.url
-    return _get_rel_path(path, base, page is not None)
+        return get_relative_url(path, page.url)
+    return posixpath.join(base, path)
 
 
 @functools.lru_cache(maxsize=None)
@@ -281,14 +292,6 @@ def _get_norm_url(path):
     if parsed.scheme or parsed.netloc or path.startswith(('/', '#')):
         return path, True
     return path, False
-
-
-@functools.lru_cache()
-def _get_rel_path(path, base, base_is_url):
-    if base_is_url:
-        return get_relative_url(path, base)
-    else:
-        return posixpath.join(base, path)
 
 
 def create_media_urls(path_list, page=None, base=''):
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -252,21 +252,27 @@ def is_error_template(path):
 
 
 @functools.lru_cache(maxsize=None)
-def _norm_parts(parts):
-    dest_parts = []
-    for p in parts.split('/'):
+def _norm_parts(path):
+    parts = []
+    for p in path.split('/'):
         if p not in ('.', ''):
             if p == '..':
-                if dest_parts:
-                    dest_parts.pop()
+                if parts:
+                    parts.pop()
             else:
-                dest_parts.append(p)
-    return dest_parts
+                parts.append(p)
+    return parts
 
 
 def get_relative_url(url, other):
     """
     Return given url relative to other.
+
+    Both are operated as slash-separated paths, similarly to the 'path' part of a URL.
+    The last component of `other` is skipped if it contains a dot (considered a file).
+    Actual URLs (with schemas etc.) aren't supported. The leading slash is ignored.
+    Paths are normalized ('..' works as parent directory), but going higher than the
+    root has no effect ('foo/../../bar' ends up just as 'bar').
     """
     # Remove filename from other url if it has one.
     dirname, _, basename = other.rpartition('/')


waylan:
So this change removes calls to the standard lib and replaces them with your own custom implementation. I realize that there may be some standard lib code that we don't need, but the assumption is that the standard lib code is well tested and accurate. Your custom code is an unknown. I suspect the existing tests don't cover nearly enough to be sure this doesn't break various edge cases (admittedly I didn't check). We will always go with accuracy over performance. If you want this accepted, then you will need to demonstrate that this is very well tested.

By the way, you have been providing a lot of performance patches lately. I don't object to that. In fact, I appreciate that someone is looking at the issue, as I don't have any time to dedicate to the issue. In fact, reviewing all these patches which don't improve the project in any perceivable way (from my perspective) just adds more burden on me. Therefore, from my perspective you need to demonstrate that you aren't breaking anything. Your comments always mention how much of a performance improvement this makes. But I don't care much about that. The largest site I have is less than 20 pages. I have the perception that my sites build instantly on my very old machine. Personally I see no need for performance improvements. What I care about is *ease of maintenance* and *accuracy*. If you would like me to continue to spend time reviewing your performance improvements, please focus on those two factors first. Any performance improvements will always be second to those in my reviews.

As an example, if the proposed change in this PR were to result in a future bug report, my "fix" would simply be to revert to the previous implementation. I'm not interesting in spending the time to work through the custom code to fix various edge cases. That is more of a maintenance burden than I want to take on. Therefore, before I am willing to accept it, I need to be sure that those bug reports aren't likely to happen in the future.

oprypin:
> the assumption is that the standard lib code is well tested and accurate

Well, *generally*, it's tested for file system paths relative to the current directory, not something virtual...
I think this argument should not be used like that; stdlib or not makes no difference. But sure, you can say this particular usage of it is tested by time.
Anyway, that's just a pointless comment by me.

---

> you will need to demonstrate that this is very well tested

So this function certainly supports less functionality than the old one, for example collapsing `foo/..` into nothing (and, well, the quite funny functionality of resolving paths relative to the current directory...). But it's easy to see that that functionality is not needed.

It's slightly unclear how to declare exhaustiveness, but the main idea is proving that inputs to it can only be some particular subset of strings, and then perhaps fuzz it within those constraints.

---

> If you would like me to continue to spend time reviewing your performance improvements

Well, don't worry much about that. I intended to have stopped already but this one is just such a big realization (also realization that stdlib usage isn't exempt from scrutiny).

> I'm not interesting in spending the time to work through the custom code to fix various edge cases

I can promise a 24-hour response time with a fix :D

waylan:
If you would like this merged, then you will need to either point to an existing set of tests (I haven't taken the time to check), or provide a comprehensive set of tests.

oprypin:
I have done fuzzing as mentioned. It asserts that the output of the old function is the same as of the new function.

It found mostly differences that I already knew about (and I did ensure that it found them for real), so I explicitly ruled out those inputs (see `assume` below).

The fuzzer is told to generate a concatenation of any combination of any number of "any unicode string", "slash", "dot"
(see `one_of` -> `lists` -> `join` below).
I added slash and dot explicitly because without them the fuzzer is totally lost in the search space, although with infinite time it'd make no difference.

```python
# Requires: `pip install pytest hypothesis atheris`

import posixpath

def get_relative_url_1(url, other):
    if other != '.':
        # Remove filename from other url if it has one.
        parts = posixpath.split(other)
        other = parts[0] if '.' in parts[1] else other
    relurl = posixpath.relpath(url, other)
    return relurl + '/' if url.endswith('/') else relurl

def get_relative_url_2(url, other):
    other_parts = other.split('/')
    # Remove filename from other url if it has one.
    if other_parts and '.' in other_parts[-1]:
        other_parts.pop()

    other_parts = [p for p in other_parts if p not in ('.', '')]
    dest_parts = [p for p in url.split('/') if p not in ('.', '')]
    common = 0
    for a, b in zip(other_parts, dest_parts):
        if a != b:
            break
        common += 1

    rel_parts = ['..'] * (len(other_parts) - common) + dest_parts[common:]
    relurl = '/'.join(rel_parts) or '.'
    return relurl + '/' if url.endswith('/') else relurl


from hypothesis import strategies as st, given, assume, settings, example

@st.composite
def path(draw):
    elements = st.one_of(st.text(), st.just('/'), st.just('.'))
    path = ''.join(draw(st.lists(elements)))

    assume(path != '')  # The original function has `raise ValueError("no path specified")`
    assume('/../' not in f'/{path}/')  # The new function doesn't support normalizing paths

    # print(repr(path))  # Uncomment to see produced examples.
    return path

@settings(max_examples=100000)
@given(path(), path())
def test_equal(url, other):
    assume(url.startswith('/') == other.startswith('/'))  # The original function sees one path as relative to the current directory!

    assert get_relative_url_1(url, other) == get_relative_url_2(url, other)

# Run with 'hypotesis' fuzzer: `pytest fuzz_get_relative_url.py -s`

# Run with 'atheris' fuzzer: `python fuzz_get_relative_url.py`
if __name__ == "__main__":
    import sys, atheris
    atheris.Setup(sys.argv, test_equal.hypothesis.fuzz_one_input)
    atheris.Fuzz()
```


I found that the function produced differences only for these classes of inputs that I didn't already know about:

```python
@example('foo', 'bar./')  # The trailing slash should disqualify it from being a file.
@example('foo', 'foo/bar./.')  # The trailing dot should disqualify it from being a file.
```

I fixed this and will also add unittests.

After that I continued to run on each of those 2 fuzzers for well over an hour and for well over 100000 examples, and it didn't find anything more.

Now what remains is to prove that resolving `..` is indeed not relied on anywhere, and we might be good to go.

oprypin:
And my exhaustive analysis of usages. Sadly the results aren't so clear cut.

Summary - the only things that could've regressed:

* If someone passed a path like `aaa/../aaa/x.html`  to the `|url` filter, previously it'd become `aaa/x.html`, now it will stay like that. And in a subdirectory, of course, it will become `'../aaa/../aaa/x.html'`. Which would actually probably still work but yeah...

* If someone passed a path like `aaa/../aaa/x.html`  to `config['extra_templates]`, previously it'd become `aaa/x.html`, now it will stay like that and affect the `base_url` in the template.

I really think nobody would've done such things, and also this old behavior was probably not even the preferred behavior among these two. It's also quite easy to look through a bunch of themes with `git grep '| *url'` and see that indeed there isn't anything that looks even remotely like that (actually this could even be the main way forward).

Otherwise, if we want to preserve this, in each of those cases, first wrapping those into a `normpath` would fix it.

I'm willing to go through the motions, whatever the choice. Just hope you're not too tired of this.

And I'll mention again that the old implementation has latent bugs, e.g. if leading slashes are involved (thankfully this very audit proves that such inputs can't happen) and if `other` has too many `..` then real filesystem paths are revealed (also never happens currently). Maybe figuring out *those* edge cases is the higher mental overhead.

## Full analysis (read only the parts in bold :yum:)

Listing all origins of the inputs:

1. https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L38
    * *url*: `'.'`, ok
    * *other*: `Page.url` below
1. https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L82
    * *url*: `'.'`, ok
    * *other*: from <code><a href="https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L64">_build_template</a>(<b>name</b>, &hellip;)</code>
        **actually** these are first guarded by 
        1. https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L106-L112
            * first this is guarded by `env.get_template` so absolute or empty paths can't pass.
            * from <code><a href="https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L101">_build_theme_template</a>(<b>template_name</b>, &hellip;)</code>
                1. https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L290-L291
                    * `config['theme'].static_templates` below
        1. https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L134-L146
            * first this is guarded by `files.get_file_from_path` so absolute or empty paths can't pass.
            * from <code><a href="https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L129">_build_extra_template</a>(<b>template_name</b>, &hellip;)</code>
                1. https://github.com/mkdocs/mkdocs/blob/3c9358bb6afedb559850a640961a691d698866ba/mkdocs/commands/build.py#L293-L294
                    * `config['extra_templates']` below
1. https://github.com/mkdocs/mkdocs/blob/56f98cb72264136aa017020c56885b49ba64a9b7/mkdocs/structure/files.py#L170
    * *url*: `File.url` below
    * *other*: `File.url` below
    * *other*: from <code><a href="https://github.com/mkdocs/mkdocs/blob/56f98cb72264136aa017020c56885b49ba64a9b7/mkdocs/structure/files.py#L168">url_relative_to</a>(<b>other</b>)</code>
        1. https://github.com/mkdocs/mkdocs/blob/56f98cb72264136aa017020c56885b49ba64a9b7/mkdocs/structure/pages.py#L227
            * `self.file` &rarr; only used as `File.url`, already mentioned
1. https://github.com/mkdocs/mkdocs/blob/8fa2a1a624104dda3eb93ec8ab41e74154bdc1b1/mkdocs/utils/__init__.py#L276-L282
    * *other*: `Page.url`, already mentioned
    * *url*: from <code><a href="https://github.com/mkdocs/mkdocs/blob/8fa2a1a624104dda3eb93ec8ab41e74154bdc1b1/mkdocs/utils/__init__.py#L276">normalize_url</a>(<b>path</b>, &hellip;)</code>
        1. https://github.com/mkdocs/mkdocs/blob/ff0b7260564e65b6547fd41753ec971e4237823b/mkdocs/utils/__init__.py#L260-L262
            * Note: emptiness gets replaced with '.', no other change
        1. https://github.com/mkdocs/mkdocs/blob/ff0b7260564e65b6547fd41753ec971e4237823b/mkdocs/utils/filters.py#L12-L14
            * Would have been problematic to pass '' as an input to `|url` filter, if the top of `normalize_url` didn't eliminate it.
            * `normalize_url` also has a check against absolute paths, so leading slash isn't let through.
            * **if someone was explicitly using `'..'` to pass to `|url` filter, that can be a problem**.
              for example `'css/../css/base.css'|url` stays as is, without being normalized. And in a subdirectory it becomes `'../css/../css/base.css'`.
              **So this is the big one, and the only one, really. If there's any crazy template explicitly adding '..' into the value passed `|url` filter (because where else would it come from?) then there is a regression for it.**
    * *url*: `path = _get_norm_url(path)`
        1. https://github.com/mkdocs/mkdocs/blob/8fa2a1a624104dda3eb93ec8ab41e74154bdc1b1/mkdocs/utils/__init__.py#L287-L293
            * Can't introduce `'..'` or empty output or leading slash

---

Indirect origins mentioned above:

* **`Page.url`**
    1. https://github.com/mkdocs/mkdocs/blob/520314fed933aed8de62b08dd7fc6e25c0ff482b/mkdocs/structure/pages.py#L87-L89
        * can't introduce `'..'` or leading slash, **can** introduce empty output, as `File.url` can indeed be `'.'`. **This is where I realize that my fuzzer should also allow `''` as the 2nd (`other`) arg, although it's not allowed as 1st (`url`). So [I revise it](https://gist.github.com/oprypin/e968c613bb9de87b2d2fbd4409bed3ff/revisions#diff-c83daefeac180919dd896fc9ca8c2daf6f28e78dab4018d011cce3c0e4181156) and run it a bunch more**. `Page.url` is never used for `url` arg, only `other`, so emptiness is ok.
* **`File.url`**
    1. https://github.com/mkdocs/mkdocs/blob/520314fed933aed8de62b08dd7fc6e25c0ff482b/mkdocs/structure/files.py#L125
        * https://github.com/mkdocs/mkdocs/blob/520314fed933aed8de62b08dd7fc6e25c0ff482b/mkdocs/structure/files.py#L157-L166
            * `replace`, `split`, `+ '/'`, `= '.'`, `urlquote` can't introduce `'..'` or empty output or leading slash
            * `self.dest_path`: below
* **`File.dest_path`**:
    1. https://github.com/mkdocs/mkdocs/blob/520314fed933aed8de62b08dd7fc6e25c0ff482b/mkdocs/structure/files.py#L123
        * https://github.com/mkdocs/mkdocs/blob/520314fed933aed8de62b08dd7fc6e25c0ff482b/mkdocs/structure/files.py#L144-L155
            * `os.path.join` can't introduce `'..'` or empty output or leading slash
            * `parent` (`os.path.split`) can't introduce `'..'` or leading slash other than through `self.src_path`;
              emptiness irrelevant: `os.path.join('', 'a') == 'a'`
            * `self.src_path`: below
* **`File.src_path`**:
    1. https://github.com/mkdocs/mkdocs/blob/520314fed933aed8de62b08dd7fc6e25c0ff482b/mkdocs/structure/files.py#L120
        * explicitly normalized, so all anomalies excluded
* **`config['theme'].static_templates`**:
    1. https://github.com/mkdocs/mkdocs/blob/c5e4018f042fefc573efa9766fcac59f645515b3/mkdocs/theme.py#L36
        * just a directory name, no anomalies
* **`config['extra_templates']`**:
    1. https://github.com/mkdocs/mkdocs/blob/ff0b7260564e65b6547fd41753ec971e4237823b/mkdocs/config/defaults.py#L82
        * above we already excluded leading slash or empty. **And I sure *hope* nobody was passing paths with `'..'` in their config here. But yes, this would be a regression.** E.g. `extra_templates: ['sub/../sub/../sub/aaa.html']` works but [sets the wrong `base_path`](https://github.com/mkdocs/mkdocs/blob/520314fed933aed8de62b08dd7fc6e25c0ff482b/mkdocs/commands/build.py#L82). But we could just explicitly `normpath` here and that works fine to eliminate this issue.

oprypin:
And, of course, there's the option of just implementing normalization as well, with almost no performance penalty compared to the previous state. Pushed the commits here for posterity, though you may not like the code size.

waylan:
I still want to see a comprehensive set of unittests included with this patch. And I don't care if some potential inputs will not currently be generated by MkDocs. We can't guarantee that will be the case in the future. If you want to do a custom function, then you ned to provide a complete set of unittests which demonstrate that any potential (past, present or future) input will return the correct normalized result. Until I see that, I'm not spending any more time on this.

oprypin:
Please don't always respond with the immediate assumption that I will not do something. First was just making sure that this wouldn't be discarded.
I even said

> will also add unittests

So, is the code acceptable? If so, great, I will definitely add unittests.

oprypin:
Added unittest.

Btw, the existing tests are actually quite comprehensive. They just didn't try to find super abnormal cases, I suppose. But now that's covered.

https://github.com/mkdocs/mkdocs/blob/ff0b7260564e65b6547fd41753ec971e4237823b/mkdocs/tests/structure/file_tests.py#L407-L591

oprypin:
And latest profile

s

oprypin:
BTW, in my view I have done what was requested so far. (so this is just a ping, I guess)

oprypin:
Please let me know your overall opinion on this direction. In case this just needs to wait before review, that's OK.

waylan:
I have given it a cursory look and it seems okay in principle, but I want to be extra sure your custom replacement for the standard lib function is comely reliable and that is going to take time. Time I havn't been motivated to spend on it.

oprypin:
I have done soooo much fuzzing, and they just give 100% the same result, other than known differences that [are being excluded](https://gist.github.com/oprypin/e968c613bb9de87b2d2fbd4409bed3ff/733dca62d8fad67218ddff2cfa6b56c57fc5b01f#file-fuzz_get_relative_url-py-L53-L65). Just no way to be more sure.

You can run it yourself and evaluate its thoroughness. Even intentionally introduce a bug and see how quickly it finds it.
https://gist.github.com/oprypin/e968c613bb9de87b2d2fbd4409bed3ff

[The differences](https://gist.github.com/oprypin/e968c613bb9de87b2d2fbd4409bed3ff/733dca62d8fad67218ddff2cfa6b56c57fc5b01f#file-fuzz_get_relative_url-py-L71) are just the fact that the old function can spew paths that depend on the current working directory (revealing parts of it). I presume that's not a wanted behavior. In any case, I have also [proven manually](#issuecomment-751924830) that these cases are never hit, but we will be safer with the new func.

waylan:
I'm closing this in favor of #2296. This adds our own implementation which is more of a maintenance burden.

oprypin:
What do you mean "in favor of"...

#2296 was a stepping stone to get **all** the breaking changes out of the way. This pull request now 100% exactly matches the implementation at master.

This change, now compared to master, still cuts off 9% of build times for a site with 142 pages. And more if there are more.
https://github.com/crystal-lang/crystal-book/tree/1fc786788f25f95dde767aad1676eed7c322eda3/docs
```shell
$ for i in {1..3}; do mkdocs build 2>&1 | grep built; done
INFO    -  Documentation built in 3.13 seconds 
INFO    -  Documentation built in 3.09 seconds 
INFO    -  Documentation built in 3.05 seconds 
```
```shell
$ for i in {1..3}; do mkdocs build 2>&1 | grep built; done
INFO    -  Documentation built in 2.88 seconds 
INFO    -  Documentation built in 2.83 seconds 
INFO    -  Documentation built in 2.81 seconds 
```

---

You had mentioned before

> The largest site I have is less than 20 pages

Well, there you go, the link above. That's a very real site with every page written by a human.

---

At the very least I would've wanted to merge master to preserve this PR in a better shape. But certainly I didn't predict this to be closed instantly.

Now the best view of this diff we have is not this pull request but rather https://github.com/oprypin/mkdocs/compare/relpath#diff

---

> more of a maintenance burden

1. The implementation is well-specified; if the result is exactly the same as the current one and isn't subject to change, I don't see why you'd need to work on it more. That is, the amount of code doesn't matter that much. At the same time, it's not even a lot of added code.

2. I would be very happy to step in for maintenance if you can't.

waylan:
> This pull request now 100% exactly matches the implementation at master.

Which means it adds more maintenance burden for no gain.

> This change, now compared to master, still cuts off 9% of build times for a site with 142 pages.

Meh. Not a priority. The amount of gain is not worth the extra burden.

oprypin:
Can you hand off the burden to me?

How do you calculate whether it's worth saving thousands of people a second at a time?

oprypin:
This issue really bothers me.

I am developing a plugin that produces a lot of documentation pages based on source code. My users immediately see that the build is slow. I know exactly how to improve it and have done all the work (even *way* more work than should be necessary) but I am still powerless to fix it.

I can tell the users "well, you can use this fork of MkDocs to speed up the build by 10%, because the maintainer of MkDocs doesn't accept my improvement". Um, well it is actually what I say, but they don't take me up on that.

But hey, maybe you don't care about that usecase -- maybe I'm holding it wrong. But I have also shown you a very real site that just has 142 pages, each of them hand-written, regardless of plugins.
https://github.com/crystal-lang/crystal-book/tree/1fc786788f25f95dde767aad1676eed7c322eda3/docs

I don't know where your impression of what performance gains are "worth it" is coming from, but I think a guaranteed 9% improvement would generally be considered *really good*, especially from a code delta of merely [30 lines added, 15 deleted](https://github.com/oprypin/mkdocs/compare/relpath#diff).
Again, considering just how many people use this project, *and* that they wait for the reloads "in real time", I don't know how you so easily weigh a small convenience in your favor over all of their experience.

I have also provided all kinds of other arguments. E.g. that the code has no reason to change so it's not exactly a maintenance burden. Also that I would gladly respond to any issues with this code myself. *Also* you can easily revert it any time.

I don't know how I deserve this -- you really just shrug away everything I say. My words, even perfectly backed up by facts, are instantly overridden just by your opinion, *feeling*, even.

oprypin:
@waylan 
Could you stop ignoring me whenever that is convenient?

This is still a very important improvement. I'm not going to just forget this. It comes up in my mind quite often actually.

You still haven't provided any solid arguments against this, and I'm appalled that you think it's OK to let your mere opinion stand above arguments.

And to reiterate on the claim that you think you have: any "maintenance burden" from this is instantly fixable by reverting this.

PR #2274, opened 2020-12-30
Don't neglect to register our Markdown extension

oprypin:
(all other "normal" extensions do this)

---

My use case for this is that I am rendering a Markdown sub-document that is to be re-integrated into the root document, and I want to inherit all the Markdown extensions.

I create a markdown instance like `new_md = Markdown(extensions=old_md.registeredExtensions)` but this `_RelativePathExtension` extension is skipped from that because it never registered itself.

I see that this extension is not public on purpose, and I'm trying not to instantiate it directly, but I think copying it like this should not be prevented.

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -240,5 +240,6 @@ class _RelativePathExtension(Extension):
         self.files = files
 
     def extendMarkdown(self, md):
+        md.registerExtension(self)
         relpath = _RelativePathTreeprocessor(self.file, self.files)
         md.treeprocessors.register(relpath, "relpath", 0)


waylan:
> (all other "normal" extensions do this)

This is absolutely not true.  In fact, many of Markdown's own first-party extensions do not register themselves. This was the subject of Python-Markdown/markdown#559/ Note the final decision made there:

> An extension is "registered" if it needs to utilize the "reset" feature. and that would only ever be a decision made by the developer of the extension. If you are trying to determine if a certain extension is in use, then that should be done in another way as it is outside the scope of the "register" feature.

That being the case, we will not be making this change as it offers no value.

Issue #2062, opened 2020-04-14
Install of mkdocs fails with nltk 3.5


klaascuvelier:
Hi,

First of all, thank you for mkdocs. We're using this in our company and are very happy to have it.

I noticed that the installation of mkdocs started failing for us yesterday. After some investigation I found out that the [nltk](https://www.nltk.org/install.html) lib had a version bump (3.5) yesterday and that's were it goes wrong.

I was able to use the previous nltk version (3.4.5) by adding it directly to our ci file
```
pages:
  image: python:3.7-alpine
  stage: 🥾 Bootstrap
  script:
    - pip install nltk==3.4.5
    - pip install mkdocs
    - pip install mkdocs-material
    - mkdocs build
```
This solves the issue for us.

I am not really used to python and its ecosystem, so perhaps I am overseeing an actual "fix".
But if this is a really issue, I guess the `3.5` nltk version should be _blacklisted_ from the install?

<details>
  <summary>Output of the error</summary>
  
```
$ pip install mkdocs
18 Collecting mkdocs
19   Downloading mkdocs-1.1-py2.py3-none-any.whl (6.4 MB)
20 Collecting click>=3.3
21   Downloading click-7.1.1-py2.py3-none-any.whl (82 kB)
22 Collecting Jinja2>=2.10.1
23   Downloading Jinja2-2.11.1-py2.py3-none-any.whl (126 kB)
24 Collecting Markdown>=3.2.1
25   Downloading Markdown-3.2.1-py2.py3-none-any.whl (88 kB)
26 Collecting PyYAML>=3.10
27   Downloading PyYAML-5.3.1.tar.gz (269 kB)
28 Collecting tornado>=5.0
29   Downloading tornado-6.0.4.tar.gz (496 kB)
30 Collecting livereload>=2.5.1
31   Downloading livereload-2.6.1-py2.py3-none-any.whl (23 kB)
32 Collecting lunr[languages]==0.5.6
33   Downloading lunr-0.5.6-py2.py3-none-any.whl (36 kB)
34 Collecting MarkupSafe>=0.23
35   Downloading MarkupSafe-1.1.1.tar.gz (19 kB)
36 Requirement already satisfied: setuptools>=36 in /usr/local/lib/python3.8/site-packages (from Markdown>=3.2.1->mkdocs) (46.1.1)
37 Collecting six
38   Downloading six-1.14.0-py2.py3-none-any.whl (10 kB)
39 Collecting future>=0.16.0
[...]

PR #2281, opened 2021-01-09
Unpin lunr.py dependency

yeraydiazdiaz:
Hi all, I was about to release a new version of lunr.py and started to get errors while building the docs, this is because mkdocs is pinning it to 0.5.8 which raises a version conflict error.

I've read through https://github.com/mkdocs/mkdocs/issues/2062 and decided to [restrict lunr.py to nltk<3.5](https://github.com/yeraydiazdiaz/lunr.py/pull/89/commits/b648c9206b5f0d113ab8725fda3a95d1a944ad91) to avoid these installation issues. However mkdocs does need to loosen the restrictions on lunr for us to build the docs moving forward.

This does introduce a slight problem though, when we release a new version of lunr.py users of mkdocs will get a version that may be out of sync with the version of lunr.js mkdocs ships with. In practice this is not a big problem because there's usually very minor differences between lunr.js versions. The solution is to make sure things are in sync before releasing so I'm happy to make a PR to mkdocs after each release of lunr.py to make sure things are sync.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
         'click>=3.3',
         'Jinja2>=2.10.1',
         'livereload>=2.5.1',
-        'lunr[languages]==0.5.8',  # must support lunr.js version included in search
+        'lunr[languages]>=0.5.8',  # must support lunr.js version included in search
         'Markdown>=3.2.1',
         'PyYAML>=3.10',
         'tornado>=5.0',


waylan:
As there is the need for a strict connection between the version of lunr.js and lunr.py, I would prefer to only pin a specific version (using `==`), which avoids any surprises. Yes, if you update lunr.py in the future, then we would need to update mkdocs to get access to the update.

Perhaps we could do something like `==x.x` rather than `==x.x.x`. That way, anything from `x.x.0` and up would work. However, for that to work, you would need to guarantee that any point release of lunr.py would stick to the same lunr.js version.

Also, note the requirements files. We run the tests twice; once with each of the following:

1. [requirements/project.txt][1]: this should match `setup.py` exactly and should always use the latest version we officially support.
2. [requirements/project-min.txt][2]: this should pin the absolute minimum version of a dependency that is known to work.

Therefore, any change to the dependencies in `setup.py` would also need to be reflected in `requirements/project.txt`. However,  we would only update `requirements/project-min.txt` if we updated the minimum dependency. For example, if in the future to update to a newer version of lunr.js, then we would need to also update `requirements/project-min.txt` to reflect the new minimum version of `lunr.py` that works with that version of `lunr.js`. However, in this instance, where we are not updating `lunr.js`, we would only need to update `requirements/project.txt` and `setup.py`.

[1]: https://github.com/mkdocs/mkdocs/blob/master/requirements/project.txt
[2]: https://github.com/mkdocs/mkdocs/blob/master/requirements/project-min.txt

yeraydiazdiaz:
> As there is the need for a strict connection between the version of lunr.js and lunr.py, I would prefer to only pin a specific version (using ==), which avoids any surprises. Yes, if you update lunr.py in the future, then we would need to update mkdocs to get access to the update.

I understand and I agree, but pinning to a specific version breaks the building of the documentation when bumping versions of lunr.py. The only way I can see of being able to pin to a specific version and is for lunr.py to not use mkdocs for its documentation, which is quite ironic but I think it's the best way forward to avoid complications in the future.

Thanks for your time 🙂

waylan:
We have run into a similar issue with some of the themes which are distributed separately from MkDocs. The updated theme requires the new version of MkDocs, but the new version hasn't been released yet. So how do we work on the theme?

I will run everything from a venv with the correct dependencies installed. I might install MkDocs from master or maybe even a fork which only has the dependencies changed. But even then, by CI builds won't run correctly until the release is made. It can be a challenge.

PR #2283, opened 2021-01-12
Modify the syntax of version option and checking python version

sa1ar:
- Use message  and  version arguments of click.version_option for better syntax instead string formatting.
- Modify pgk_dir typo.
- FIx pylint: C0103.
- DRY: remove repeated call of python version and add PYTHON_VERSION.

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -8,7 +8,7 @@ import click
 PYTHON_VERSION = sys.version[:3]
 # TODO: Remove this check at some point in the future.
 # (also remove flake8's 'ignore E402' comments below)
-if PYTHON_VERSION[0] != '3':  # pragma: no cover
+if PYTHON_VERSION < '3':  # pragma: no cover
     raise ImportError('A recent version of Python 3 is required.')
 
 from mkdocs import __version__                            # noqa: E402
@@ -115,8 +115,7 @@ PKG_DIR = os.path.dirname(os.path.abspath(__file__))
 @click.version_option(
     __version__,
     '-V', '--version',
-    message='%(prog)s, version %(version)s from {} (Python {})'.format(
-        PKG_DIR, PYTHON_VERSION)
+    message=f'(prog)s, version %(version)s from { PKG_DIR } (Python { PYTHON_VERSION })'
 )
 @common_options
 def cli():
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -5,11 +5,11 @@ import sys
 import logging
 import click
 
-from mkdocs import __version__                            # noqa: E402
-from mkdocs import utils                                  # noqa: E402
-from mkdocs import exceptions                             # noqa: E402
-from mkdocs import config                                 # noqa: E402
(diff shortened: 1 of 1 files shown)


mkdocs/__main__.py
```diff
@@ -107,13 +108,16 @@ def callback(ctx, param, value):
...
 
 
 @click.group(context_settings={'help_option_names': ['-h', '--help']})
 @click.version_option(
-    '{} from {} (Python {})'.format(__version__, pgk_dir, sys.version[:3]),
-    '-V', '--version')
+    __version__,
+    '-V', '--version',
+    message='%(prog)s, version %(version)s from {} (Python {})'.format(
+        PKG_DIR, PYTHON_VERSION)
```
waylan, line 119: I'm assuming the message argument is a more recent addition to click. In any event, I guess there is no reason to not make the change.

However, as we are making a change here, we might as well go all the way. As all supported Python versions now include support for f-strings, lets use them. They're not only easier to read, but they are more performant.

```python
message=f'%(prog)s, version %(version)s from { PKG_DIR } (Python { PYTHON_VERSION })'
```

waylan (changes requested):
This entire change seems completely unnecessary to me. But you have done the work and it is does update the code to Clicks current API. However, I have a couple concerned as described below. If you fix those, I'll merge this.

mkdocs/__main__.py
```diff
@@ -5,9 +5,10 @@
 import logging
 import click
 
+PYTHON_VERSION = sys.version[:3]
 # TODO: Remove this check at some point in the future.
 # (also remove flake8's 'ignore E402' comments below)
-if sys.version_info[0] < 3:  # pragma: no cover
+if PYTHON_VERSION[0] != '3':  # pragma: no cover
```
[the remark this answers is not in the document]
oprypin: Right. As is, this is a downgrade.
The check was separating on the boundary between Python 2 and 3, which is correct
but in this state it separates Python 3 and Python 30 etc. from Python 2 and Python 4 etc.

sa1ar:
> This entire change seems completely unnecessary to me. But you have done the work and it is does update the code to Clicks current API. However, I have a couple concerned as described below. If you fix those, I'll merge this.

Ok, I will fix issues

sa1ar:
Thanks for reviews, I've tried to fix them, I wish it's will be OK

mkdocs/__main__.py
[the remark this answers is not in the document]
sa1ar: I've removed this check completely.

mkdocs/__main__.py
```diff
@@ -5,11 +5,6 @@
 import logging
 import click
 
-# TODO: Remove this check at some point in the future.
-# (also remove flake8's 'ignore E402' comments below)
```
waylan, line 9: As we are removing this block, please also follow the advice on this line and remove the `# nova: E402` comments from the import lines below this block. They are no longer needed as this block is no longer before them. Otherwise, looks good.

sa1ar: Done.

waylan (approved):

waylan (approved):

PR #2286, opened 2021-01-13
Fix search function for large documents

Jmarshall95:
The search function does not function for documents of a certain size and above, this will fix that.
Hopefully it is not a controversial change.


waylan:
This is almost certainly not what we want. The `doSearch()` function is called when the user submits her search query. It should not be called while search is initialized. In the event that search results are displayed on their own separate page (rather than a modal on any page), then yes in that case `doSearch` would be called after search is initialized. However, that is not part of initialization. And because it is not always wanted after  initialization, then it does not belong anywhere in the code we provide. Instead it should be called from the template which is provided by your theme.

Aside from the above, I have no idea what problem you are trying to solve here. Perhaps an explanation of the problem would help us to find a workable solution.

oprypin:
If you are ready to follow up with an explanation why this pull request is good, we can reopen this. As of now, we can't do anything with this contribution.

Issue #1496, opened 2018-05-24
Document CNAME files for gh-pages custom domains.


waylan:
Previously GitHub's documentation throughly documented how to use CNAME files to configure custom domains on GitHub pages. However, recently they seem to have removed that documentation and instead instruct users to use the web interface to set up a custom domain. The one exception being [here][1].

The problem is that under the hood, the web interface simply creates a CNAME file. However, as documented, the next time the user uses `gh-deploy`, it completely overwrites the entire branch and the CNAME file is removed if the user has not manually added the file to their `docs_dir`. Of course, the user won't bother to do that if we don't tell them to in our docs.

There are a few ways we could document this:

1. Document how to manually create the file ourselves, including what content to add to the file.
2. Instruct the user to use GitHub's web interface and then to copy the CNAME file from the repo to their `docs_dir`.
3. Use a combination of the two by suggesting that advanced users can create the file themselves (the specifics of how to create the file and what it contains would be unnecessary for such users) and suggest that anyone who feels uncomfortable manually creating the file use the web interface and then copy the file.

I think option 3 makes the most sense.

Note that this issue came to my attention while attempting to answer [this question][2] on the mailing list.

[1]: https://help.github.com/articles/troubleshooting-custom-domains/#github-repository-setup-errors
[2]: https://groups.google.com/forum/#!topic/mkdocs/BnmTrL2VrUM

PR #2292, opened 2021-01-22
Add error when the GitHub Pages custom domain does not match local CNAME file

theacodes:
Although documentation was added in #1496 around needing to place the CNAME file for custom domains in your docs_dir, GitHub's default behavior of adding the CNAME directly to the gh-pages branch means that lots of folks will get bitten. This has the unfortunate and severe side effect of taking their site down and it's not immediately obvious what went wrong. This PR adds a new check: if there's a CNAME file in the gh-pages branch and it doesn't match the CNAME file in docs_dir (or the CNAME file in docs_dir doesn't exist) it will terminate the deployment. It adds a flag called "--ignore-cname" to bypass this check.

This is my first time contributing to mkdocs, so I hope I did everything as expected. Please let me know if there's any extra steps or stylistic changes needed.

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -51,6 +51,7 @@ remote_name_help = ("The remote name to commit to for Github Pages. This "
                     "overrides the value specified in config")
 force_help = "Force the push to the repository."
 ignore_version_help = "Ignore check that build is not being deployed with an older version of MkDocs."
+ignore_cname_help = "Ignore check that deploying will remove an existing CNAME file."
 watch_theme_help = ("Include the theme in list of files to watch for live reloading. "
                     "Ignored when live reload is not used.")
 wait_help = "Wait the specified number of seconds before reloading (default 0)."
@@ -168,10 +169,12 @@ def build_command(clean, **kwargs):
 @click.option('--force', is_flag=True, help=force_help)
 @click.option('--ignore-version', is_flag=True, help=ignore_version_help)
 @click.option('--shell', is_flag=True, help=shell_help)
+@click.option('--ignore-cname', is_flag=True, help=ignore_cname_help)
 @common_config_options
 @click.option('-d', '--site-dir', type=click.Path(), help=site_dir_help)
 @common_options
-def gh_deploy_command(clean, message, remote_branch, remote_name, force, ignore_version, shell, **kwargs):
+def gh_deploy_command(clean, message, remote_branch, remote_name, force, ignore_version, shell,
+                      ignore_cname, **kwargs):
     """Deploy your documentation to GitHub Pages"""
     try:
         cfg = config.load_config(
@@ -180,7 +183,8 @@ def gh_deploy_command(clean, message, remote_branch, remote_name, force, ignore_
             **kwargs
         )
         build.build(cfg, dirty=not clean)
-        gh_deploy.gh_deploy(cfg, message=message, force=force, ignore_version=ignore_version, shell=shell)
+        gh_deploy.gh_deploy(cfg, message=message, force=force, ignore_version=ignore_version,
+                            shell=shell, ignore_cname=ignore_cname)
     except exceptions.ConfigurationError as e:  # pragma: no cover
         # Avoid ugly, unhelpful traceback
         raise SystemExit('\n' + str(e))
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -51,7 +51,6 @@ remote_name_help = ("The remote name to commit to for Github Pages. This "
                     "overrides the value specified in config")
 force_help = "Force the push to the repository."
 ignore_version_help = "Ignore check that build is not being deployed with an older version of MkDocs."
-ignore_cname_help = "Ignore check that deploying will remove an existing CNAME file."
 watch_theme_help = ("Include the theme in list of files to watch for live reloading. "
                     "Ignored when live reload is not used.")
 wait_help = "Wait the specified number of seconds before reloading (default 0)."
@@ -169,12 +168,10 @@ def build_command(clean, **kwargs):
 @click.option('--force', is_flag=True, help=force_help)
 @click.option('--ignore-version', is_flag=True, help=ignore_version_help)
 @click.option('--shell', is_flag=True, help=shell_help)
-@click.option('--ignore-cname', is_flag=True, help=ignore_cname_help)
 @common_config_options
 @click.option('-d', '--site-dir', type=click.Path(), help=site_dir_help)
 @common_options
-def gh_deploy_command(clean, message, remote_branch, remote_name, force, ignore_version, shell,
-                      ignore_cname, **kwargs):
+def gh_deploy_command(clean, message, remote_branch, remote_name, force, ignore_version, shell, **kwargs):
     """Deploy your documentation to GitHub Pages"""
     try:
         cfg = config.load_config(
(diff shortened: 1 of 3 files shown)


waylan:
> GitHub's default behavior of adding the CNAME directly to the gh-pages branch

Is this a rather new behavior? I don't recall that happening the last time I looked into it, which was, admittedly, some years ago.

Regardless, I'm curious how this could be an issue as the deployment will overwrite everything in the gh-pages branch. In fact, every deployment overwrites the file all over gain. We certainly aren't having any issues deploying mkdocs.org, which is a GitHub pages site generated by MkDocs using a [CNAME][1] file.

Or is the problem occurring when the CNAME file pushed by MkDocs contains something different than the automatically generated one provided by GitHub? Is there an error returned, or does it just fail silently?

Also, as I understand it, any files already in the branch which are not included in a new deployment are deleted from the branch. So wouldn't adding an `--ignore-cname` option cause the existing file to be lost, which would also break things for the user?

I can see raising a warning if we detect that a deployment may fail (avoid a silent failure), but I don't see the need to add the `--ignore-cname` option. If the user does not need/want to deploy a CNAME file, then they can remove it from their source files.

In short, I'm trying to understand what problem this is supposed to resolve.

Assuming we have a problem that needs to be resolved, how does [c-w/ghp-import][2] deal with this? As we only provide a light wrapper around that lib, I would rather have the upstream lib deal with this issue.

Actually, since #2257 we support the latest release of ghp-import, which includes its own option `CNAME` option which accepts a string and generates a CNAME file using that string as the content. In #2257 our goal was to update the supported version of ghp-import, not add new features. However, I would be open to adding new features in the future, including support for generating a CNAME file rather than copying one from the source files. In that case, the only thing MkDocs would provide would be the option for the user to define the string. That string would be passed to ghp-import and everything else would be handled upstream. The point is that this "fix" would be completely ignored if we were to add such a feature. Therefore, any `fix` should probably be made with ghp-import directly.

[1]: https://github.com/mkdocs/mkdocs/blob/master/docs/CNAME
[2]: https://github.com/c-w/ghp-import

theacodes:
I think I maybe didn't communicate the issue well enough.

> Is this a rather new behavior? I don't recall that happening the last time I looked into it, which was, admittedly, some years ago.

I'm not sure, but as of right now if you add a custom domain through the UI github will [create a commit on your behalf with the CNAME file](https://github.com/theacodes/Winterbloom_Castor_and_Pollux/commit/677c6ddc19a813624ad7318975c9d7437dc4d582).


> Regardless, I'm curious how this could be an issue as the deployment will overwrite everything in the gh-pages branch. In fact, every deployment overwrites the file all over gain. We certainly aren't having any issues deploying mkdocs.org, which is a GitHub pages site generated by MkDocs using a CNAME file.

The issue arises when you *don't* have a CNAME file in your MkDocs directory. So, imagine this scenario:

1. User creates a new MkDocs project, publishes it to github pages under `https://{user}.github.io/{project}`
2. Later they decide to add a custom domain. They do it through the UI and GitHub adds the `CNAME` file to the `gh-pages` branch.
3. The user updates the docs and republishes. Since `ghp-import` completely replaces the `gh-pages` branch the `CNAME` file is lost and the custom domain no longer works.
4. The user's site is down and it's not obvious why it isn't working.

This change will issue a warning when the user tries to publish in step `3` telling them to add the `CNAME` file to the MkDocs project.

> Assuming we have a problem that needs to be resolved, how does c-w/ghp-import deal with this? As we only provide a light wrapper around that lib, I would rather have the upstream lib deal with this issue.

As you said, it explicitly just blows away all of the content. It added a `--cname` flag for exactly this reason - not breaking custom domains when publishing. Unfortunately, since MkDocs uses `ghp-import` under the covers their *explicit* behavior is now our *implicit* behavior.

>  including support for generating a CNAME file rather than copying one from the source files.

I'm all for adding the cname option into the `mkdocs.yml` and passing that to `ghp-import`, but this warning would still be needed if the user forgets to add it after configuring through the UI or if they add one that that isn't the same as the one configured in the UI.

waylan:
> if you add a custom domain through the UI github will create a commit on your behalf with the CNAME file.

Okay, this is a change. In the past that was an option which was disabled by default. I understand why they changed it, but I was unaware of the change as I haven't set up any new sites for some time. In any event, thank you for more fully spelling out the issue.

> This change will issue a warning when the user tries to publish in step 3 telling them to add the CNAME file to the MkDocs project.

That seems like a reasonable solution. However, I still think this makes more sense as a feature of ghp-import directly.

Regardless, if we were to add this directly to MkDocs, there is the issue that you are raising an error instead of a warning as you propose. Warnings don't interrupt MkDocs unless `--strict` is used. Therefore, the user can simply ignore the warning if they want to (no need to use the `--ignore-cname` option). In contrast, an error does stop the process with no way to ignore it outside of fixing the issue. I think it makes more sense to have it default to a warning and rely on the existing convention of having `--strict` escalate the warning to an error (and get rid of the `--ignore-cname option). The text of warning/error should instruct the user to either add a CNAME file or to disable the custom domain setting on GitHub.

waylan:
So I actually looked at your code and I now see that `--ignore_cname` causes the check to be skipped, rather than causing the error to not be raised. Maybe a name like `ignore-cname-check` or `skip-cname-check` would be a better name. But why would a user want to the check to be skipped except for when they don't use custom domains and want to avoid the delay caused by running the check? I'm still not convinced that we need the option.

theacodes:
> Maybe a name like ignore-cname-check or skip-cname-check

Sounds good. I'll make it `ignore-cname-check` to match the existing `ignore-version` option.

> But why would a user want to the check to be skipped except for when they don't use custom domains and want to avoid the delay caused by running the check? I'm still not convinced that we need the option.

Hmm. You might be right. I can think of a case where someone has forked a project that's using a custom domain and attempt to publish the docs for their fork, but the CNAME is mismatched. They could skip the check to make the docs publish without a CNAME or with a different CNAME specified in `{docs_dir}/CNAME`. But I think more broadly if there's a mismatch between the CNAME that GitHub sees and the CNAME that the user is trying to publish with it should fail until that's resolved.

And sorry for my ambiguous use of "warning". As you surmised, it actually is a fatal error.

waylan:
> publish the docs for their fork, but the CNAME is mismatched

This is the only use case for having an `ingore-cname-check` option. And its not a very strong one. If the fork is using a custom domain, then the user needs to update the CNAME file to match the different domain. We actually want the the check in that scenario. However, if the forked site does not use a custom domain, then it doesn't matter if there is a mismatched CNAME file. GitHub will just ignore it. Well, actually, when you push, GitHub will return a warning that you need to enable custom domains. I see this every time I push a build of MKDocs' documentation to my private fork for testing purposes. I just ignore the warning because I know that it doesn't effect the functionality and I can still visit the site at the GitHub based domain. So, not there is not good reason for this option to exist.

theacodes:
Yeah I agree. I'll take out the option to ignore mismatching/missing CNAME.
Let me know if you have any other thoughts.

theacodes:
Alright, code updated and I also updated the title of this PR to better reflect the change. :)

waylan:
Thanks, but that was the least of my concerns. You still have't addressed my other two concerns, listed in order of importance below.

1. Shouldn't this be addressed upstream? Either there needs to be a good reason why it can't or the devs of that project need to indicate that they are not willing to add such a feature. Then we will reevaluate if it should be added here.
2. Assuming you overcome that hurdle, I still think this should be a warning, not an error. If you believe an error is more appropriate, you will need to convince me.

PR #2299, opened 2021-02-10
Support theme localization (rework of #1778).

ultrabug:
This branch is an attempt to renew the efforts put by @anlau
in #1778 in providing builtin theme localization support.

mkdocs and readthedocs themes will support the 'locale' parameter
which will use localized messages.po files to translate the theme
templates.

Contributors will thus be free to propose new messages.po for
their own language.

This PR proposes the 'en' (default) and 'fr' locales which should
be compiled using : `pybabel compile -d locales` from the theme
directory before they can be used.

This code has also been tested and is compliant with the
`mkdocs-static-i18n` plugin. The two of them combined provide a
fully featured way to internationalize/localize your documentation.

PR #2301, opened 2021-02-11
Use f-strings instead of formatting

oprypin:
I saw in https://github.com/mkdocs/mkdocs/pull/2283 that you *support* f-strings.

Well, rather than changing them little by little (whenever a PR happens to touch something) why not change all at once.

Most of these were generated by automation (`pyupgrade`+`flynt`) but I excluded a few and added a few.

This is split into commits, ranging basically from ones requiring least caution (especially as the first commit is fully generated) to requiring most caution. So it's recommended to view this as commits, not as a whole.

And I can sympathize with merge conflicts with #2299, so you can leave my PR until that one is merged.

### DIFF
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -197,13 +197,13 @@ def load_config(config_file=None, **kwargs):
     errors, warnings = cfg.validate()
 
     for config_name, warning in warnings:
-        log.warning("Config value: '%s'. Warning: %s", config_name, warning)
+        log.warning(f"Config value: '{config_name}'. Warning: {warning}")
 
     for config_name, error in errors:
-        log.error("Config value: '%s'. Error: %s", config_name, error)
+        log.error(f"Config value: '{config_name}'. Error: {error}")
 
     for key, value in cfg.items():
-        log.debug("Config value: '%s' = %r", key, value)
+        log.debug(f"Config value: '{key}' = {value!r}")
 
     if len(errors) > 0:
         raise exceptions.ConfigurationError(
--- a/mkdocs/config/base.py
+++ b/mkdocs/config/base.py
@@ -67,7 +67,7 @@ class Config(UserDict):
 
         for key in (set(self.keys()) - self._schema_keys):
             warnings.append((
-                key, "Unrecognised configuration name: {}".format(key)
+                key, f"Unrecognised configuration name: {key}"
             ))
 
         return failed, warnings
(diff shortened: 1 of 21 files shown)


ultrabug:
Thanks for your sympathy @oprypin :sweat_smile:

waylan:
This looks good.

As an aside, normally I am opposed to doing these kinds of changes. I prefer waiting until PR happens to touch something to make the change. However, the work is already done here and it makes the code so much more readable.

Apparently, we are waiting until after #2299 is merged before merging this.

waylan:
#2299 has been merged, I'll happily merge this after the conflicts are resolved.

mkdocs/config/base.py
```diff
@@ -206,12 +206,12 @@ def load_config(config_file=None, **kwargs):
...
     if len(errors) > 0:
-        raise exceptions.ConfigurationError(
-            f"Aborted with {len(errors)} Configuration Errors!"
+        raise exceptions.Abort(
+            "Aborted with {} Configuration Errors!".format(len(errors))
         )
     elif cfg['strict'] and len(warnings) > 0:
-        raise exceptions.ConfigurationError(
-            f"Aborted with {len(warnings)} Configuration Warnings in 'strict' mode!"
+        raise exceptions.Abort(
```
oprypin: Note: these look like I'm changing from an f-string to `.format`, but actually I just decided to drop this change compared to master.

ultrabug:
@oprypin @waylan since we're in code linting here how would you feel in adding a github workflow to enforce this?

Also @oprypin , could I suggest we include black, isort and flake8 as well? I can help if you wish to.

oprypin:
Who said about code linting? This is not it.
Discussing that in a new thread would be much better.

ultrabug:
> Who said about code linting? This is not it.

Indeed, poor choice of words.

That being said, my question still stands here, what if `.format()` keep on being proposed/committed later on?

waylan:
@ultrabug we are already using flake8. Here's the [report][1] from the last run on this issue. And it is one of the tox envs in our [tox config][2]. I suspect that is good enough.  Currently we only use flake8's defaults (except for line length). While I don't see the need to make any changes, if you see something that could help, then that should be a separate PR from this.

Now, to focus on this issue specifically. I'm not sure requiring fstrings is desirable. Yes, in most cases, that is preferred. However, there may be a few occasions where it is not practical. For example, we could have a situation where a template string is generated and later the variables are generated and passed to the template. In that case, a call to format would be required. I'm sure other exceptions could arise. Therefore, I don't see the point in enforcing such a rule with a linter. Good review if PRs should catch these sorts of things.

[1]: https://github.com/mkdocs/mkdocs/pull/2301/checks?check_run_id=2521312479
[2]: https://github.com/mkdocs/mkdocs/blob/master/tox.ini

PR #2305, opened 2021-02-15
files structure: move src_paths as a dynamic property, add remove

ultrabug:
Plugins wishing to manipulate the Files structure have an append
method but can't dynamically work on src_paths since it's computed
once on Files instanciation

This commits allows plugins to manipulate Files trough append and
remove of file objects while not having to care about the src_paths
since it's now a dynamically computed property

### DIFF
--- a/mkdocs/structure/files.py
+++ b/mkdocs/structure/files.py
@@ -15,7 +15,6 @@ class Files:
     """ A collection of File objects. """
     def __init__(self, files):
         self._files = files
-        self.src_paths = {file.src_path: file for file in files}
 
     def __iter__(self):
         return iter(self._files)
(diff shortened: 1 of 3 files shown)


mkdocs/structure/files.py
```diff
@@ -35,6 +38,10 @@ def append(self, file):
         self._files.append(file)
         self.src_paths[file.src_path] = file
```
waylan, line 39: This line will need to be removed as it would not generate an error.

waylan (changes requested):
This looks good. However, we should have some tests added. That fact that we are not getting any failing tests with the below noted issue suggest that we need to add some.

ultrabug:
Thanks @waylan ; tests added

ultrabug:
Added mention to release notes; hopefully to the right place

Cheers

waylan (approved):

PR #2308, opened 2021-02-19
fix doc: template arg and return types of on_pre_template() have wrong type

roipoussiere:
In the documentation, updates the `__template__` argument and return type from the on_pre_template() method:

"the template contents as **string**" -> "the template contents as **a [Jinja2 Template object](http://code.nabla.net/doc/jinja2/api/jinja2/environment/jinja2.environment.Template.html)**"

test code:

```py
from mkdocs.plugins import BasePlugin

class TestPlugin(BasePlugin):
    def on_pre_template(self, template, template_name, config):
        print('on_pre_template:', type(template))
        return template
```

returns:

```
on_pre_template: <class 'jinja2.environment.Template'>
```

### DIFF
--- a/docs/user-guide/plugins.md
+++ b/docs/user-guide/plugins.md
@@ -253,12 +253,12 @@ called after the [env] event and before any [page events].
     loaded and can be used to alter the content of the template.
 
     Parameters:
-    : __template__: the template contents as string
+    : __template__: the template contents as a [Jinja2 Template object](http://code.nabla.net/doc/jinja2/api/jinja2/environment/jinja2.environment.Template.html).
     : __template_name__: string filename of template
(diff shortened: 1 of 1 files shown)


waylan:
Well that's a problem. The purpose of the `on_pre_template` event is to allow a plugin "to alter the content of the template." However, according to Jinja's documentation "a template object should be considered immutable. Modifications on the object are not supported."

However, looking through the commit history, it appears that that has always been the case. Therefore, we didn't change the behavior since first introducing it. In order words, the updated documentation is correct and always has been.

Unfortunately, there isn't much point in having the event if the `Template` object is immutable. That is unless a plugin replaced the template with a different one. In that case, we need to update the documentation to reflect that.

PR #2314, opened 2021-02-28
Highlight.js version update to 10.5.0

laoshaw:


waylan:
This needs to be added to #2314. Just push another commit to the same branch and it will be added automatically.

PR #2319, opened 2021-03-04
add a FAQ page and a newbie enty

mattip:
xref joerick/cibuildwheel#616

I spent an hour or so tracking this down, so maybe others are also surprised when they enable a plugin and search disappears.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -343,7 +343,8 @@ you're done. For specific instructions on a number of common hosts, see the
 
 ## Getting help
 
-To get help with MkDocs, please use the [discussion group], [GitHub issues] or
+Take a look at the [FAQ][FAQ], maybe the answer is already there? If not,
+please use the [discussion group], [GitHub issues] or
 the MkDocs IRC channel `#mkdocs` on freenode.
 
 [deploy]: user-guide/deploying-your-docs/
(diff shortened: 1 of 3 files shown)


waylan:
Not sure how this would make it any easier to find. As you note, it is already documented in what I would consider the most logical location. Is that really the first place you look when you have a problem, the FAQ?

I should also note that many plugins include a warning in their own documentation which informs their users that they may need to explicitly include `search` in the plugins config when their plugins enabled. I would suggest you propose the same to your plugin.

mattip:
> Not sure how this would make it any easier to find. 

The plugins page is a bit of a "wall of text".  I also expected to find a FAQ or "common problems" page for newbie errors, so I thought it would be nice to add one.

> I would suggest you propose the same to your plugin.

They suggested I document it here. See the issue I linked to mondeja/mkdocs-include-markdown-plugin#24.  Note that the project linked to at the top didn't even realize the search was gone.

pawamoy:
Doing some triaging.

In the docs, [Using plugins](https://www.mkdocs.org/dev-guide/plugins/#using-plugins) has a link to [configuration - plugins](https://www.mkdocs.org/user-guide/configuration/#plugins) which says:

> If the plugins config setting is defined in the mkdocs.yml config file, then any defaults (such as search) are ignored and you need to explicitly re-enable the defaults if you would like to continue using them:

I think that is enough. If not, maybe we can that warning in [Using plugins](https://www.mkdocs.org/dev-guide/plugins/#using-plugins) as well.

In any case, I don't think the FAQ section is useful right now. Such pages usually go unmaintained pretty quickly (from my experience at least).

pawamoy:
#3204 was deemed too redundant. The docs are pretty clear. Closing for now.

PR #2334, opened 2021-03-24
Link to GitHub discussions

### LINKED ISSUE #2256: Enable GitHub Discussions for this Repository
GitHub made their Discussions open Beta and I strongly believe that this repository could benefit from having this.

Why?
People like me feel kinda uncomfortable to use issues for asking basic questions regarding the software/tool in general and often don't like to use sites/services like IRC or similar.
Having a discussion here would help us to more easily ask questions since it now would have a dedicated place for it and as an added bonus could be used as a nice collection of answers to questions someone might have.

Andre601:
Updates links from the groups to the new discussions section.

Closes #2256

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -343,8 +343,7 @@ you're done. For specific instructions on a number of common hosts, see the
 
 ## Getting help
 
-To get help with MkDocs, please use the [GitHub discussions], [GitHub issues] or
-the MkDocs IRC channel `#mkdocs` on freenode.
+To get help with MkDocs, please use the [GitHub discussions] or [GitHub issues].
 
 [deploy]: user-guide/deploying-your-docs/
 [mkdocs]: user-guide/styling-your-docs/#mkdocs
--- a/docs/index.md
+++ b/docs/index.md
@@ -343,7 +343,7 @@ you're done. For specific instructions on a number of common hosts, see the
 
 ## Getting help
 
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -343,7 +343,7 @@ you're done. For specific instructions on a number of common hosts, see the
 
 ## Getting help
 
-To get help with MkDocs, please use the [discussion group], [GitHub issues] or
+To get help with MkDocs, please use the [GitHub discussions], [GitHub issues] or
 the MkDocs IRC channel `#mkdocs` on freenode.
```
d0ugal, line 347: I suspect we can drop the IRC channel too. I used to idle there and it was never that active. I doubt it is now.

Andre601: That's something I let @waylan decide, but yeah. The two or three times I was there was it pretty dead.

waylan: I never look at the IRC channel. I would suggest removing any mention of it.

Andre601: Removed them.

Andre601:
I took a bit of time and made a quick redesign of the readme, that I believe should be used here.
At the very least should the landscape links get removed as that service seems to be kind-of dead right now ([Here](https://landscape.io/the-current-state-of-landscape.html) is an article from the author announcing the death of the service).

Here's an image of the redesigned Readme.
image

waylan:
@Andre601 I think your changes to the readme are a great start. Feel free to include them in a separate PR.

PR #2344, opened 2021-03-29
Refactor docs

waylan:
This is the beginnings of the work summarized in #1629. A work in progress.


mkdocs.yml
```diff
@@ -15,13 +15,19 @@ theme:
...
+        - Overview: user-guide/index.md
+        - Installation: user-guide/installation.md
         - Writing Your Docs: user-guide/writing-your-docs.md
-        - Styling Your Docs: user-guide/styling-your-docs.md
+        - Choosing Your Theme: user-guide/choosing-your-theme.md
+        - Customizing Your Theme: user-guide/customizing-your-theme.md
         - Configuration: user-guide/configuration.md
         - Deploying Your Docs: user-guide/deploying-your-docs.md
-        - Custom Themes: user-guide/custom-themes.md
-        - Plugins: user-guide/plugins.md
```
oprypin, line 24: Please don't break URLs, there are many of them in the wild.
Some options would be:
* Don't change filenames despite changing sections.
* Use "redirects" plugin.

waylan: These are not `user-guide` material. They never should have been put in that directory. Therefore, they are being moved. Redirects could be an option.

waylan:
I redesigned the homepage. First I moved the "Installation" and "Getting Started" to their own respective pages. Then, as the homepage had no TOC to speak of, I removed it and styled the list of features as a grid of cards. See the screenshots below for the large, medium, and small sizes. Any and all feedback is welcome. 

# Large

homepage_large

# Medium

homepage_medium

# Small

homepage_small

waylan:
I should mention a few things:

1. These changes are only for the `mkdocs.org` website (our documentation). There have been no changes made to the theme. I used Bootstrap markup via raw HTML in Markdown and added a few custom CSS rules.
2. The changes in appearance are only to the homepage. The design of all other pages is unchanged (they still have their TOC, etc).

mkdocs.yml
waylan: Redirects have been defined using the plugin. Thanks for the suggestion.

waylan:
This doesn't address everything in #1629 (most notably the API docs), but I think this is as far as I'm taking this now. The refactor is done. Everything else are additions, which can be added at any time.

PR #2347, opened 2021-03-31
Require setuptools, mkdocs/commands/gh_deploy.py imports pkg_resources

hroncok:

PR #2354, opened 2021-04-06
Replace pkg_resources with importlib_metadata

waylan:
Related to #2347.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,8 @@ setup(
         'PyYAML>=3.10',
         'tornado>=5.0',
         'ghp-import>=1.0',
-        'pyyaml_env_tag>=0.1'
+        'pyyaml_env_tag>=0.1',
+        'importlib_metadata>=3.10'
     ],
     python_requires='>=3.6',
     entry_points={
--- a/mkdocs/plugins.py
+++ b/mkdocs/plugins.py
@@ -4,8 +4,8 @@ Implements the plugin API for MkDocs.
 """
 
 
-import pkg_resources
 import logging
+import importlib_metadata
 from collections import OrderedDict
 
 from mkdocs.config.base import Config
@@ -22,9 +22,9 @@ EVENTS = (
 
 
 def get_plugins():
-    """ Return a dict of all installed Plugins by name. """
+    """ Return a dict of all installed Plugins as {name: EntryPoint}. """
 
-    plugins = pkg_resources.iter_entry_points(group='mkdocs.plugins')
+    plugins = importlib_metadata.entry_points(group='mkdocs.plugins')
 
     return {plugin.name: plugin for plugin in plugins}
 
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -59,7 +59,8 @@ def get_packages(package):
         'PyYAML>=3.10',
         'tornado>=5.0',
         'ghp-import>=1.0',
-        'pyyaml_env_tag>=0.1'
+        'pyyaml_env_tag>=0.1',
+        'importlib_metadata>=3.10'
```
oprypin, line 63: Could this be specified as required only for Python<3.8  (which it is)?

waylan: No, the API changes in Python 3.10 and we are using the new API. Therefore, we need the backport for all supported version of Python.

waylan: While the change was introduced in 3.6 of the backport (as noted above), there have been various improvements to the new API up through version 3.10. Therefore we are requiring `importlib_metadata>=3.10`.

waylan: By the way, I had intended to include an explanation in the commit message. So thanks for the reminder.

oprypin:
This commit greatly slows down startup time.

For this very repo,
`for i in {1..5}; do time mkdocs build 2>/dev/null; done`

Before this change:
```
mkdocs build 2> /dev/null  1.10s user 0.06s system 99% cpu 1.165 total
mkdocs build 2> /dev/null  1.08s user 0.07s system 99% cpu 1.159 total
mkdocs build 2> /dev/null  1.07s user 0.08s system 99% cpu 1.159 total
mkdocs build 2> /dev/null  1.08s user 0.07s system 99% cpu 1.152 total
mkdocs build 2> /dev/null  1.07s user 0.08s system 99% cpu 1.160 total
```
After this change:
```
mkdocs build 2> /dev/null  1.54s user 0.12s system 99% cpu 1.662 total
mkdocs build 2> /dev/null  1.56s user 0.09s system 99% cpu 1.657 total
mkdocs build 2> /dev/null  1.57s user 0.08s system 99% cpu 1.657 total
mkdocs build 2> /dev/null  1.56s user 0.13s system 99% cpu 1.702 total
mkdocs build 2> /dev/null  1.57s user 0.09s system 99% cpu 1.663 total
```

oprypin:
Looks like that effect is so noticeable only when there are a lot of Python packages installed.

waylan:
The docs state:

> This functionality intends to replace most uses of `pkg_resources` entry point API and metadata API. Along with `importlib.resources` in Python 3.7 and newer (backported as importlib_resources for older versions of Python), this can eliminate the need to use the older and less efficient `pkg_resources` package.

So I find it interesting that the opposite effect is observed.

oprypin:
Um sorry, I just don't see the motivation behind this change. It replaces a common, stable package with an uncommon and unstable one.

PR #2093, opened 2020-05-05
Add --wait flag to serve command.

waylan:
This should address #2061. However, I don't perceive any actual delay. Perhaps those who have a need for could report back.

```
$ mkdocs serve --wait 60
INFO    -  Building documentation...
INFO    -  Cleaning site directory
INFO    -  Documentation built in 0.96 seconds
[I 200505 14:43:49 server:296] Serving on http://127.0.0.1:8000
INFO    -  Serving on http://127.0.0.1:8000
[I 200505 14:43:49 handlers:62] Start watching changes
INFO    -  Start watching changes
[I 200505 14:43:49 handlers:64] Start detecting changes
INFO    -  Start detecting changes
[I 200505 14:43:53 handlers:135] Browser Connected: http://127.0.0.1:8000/
INFO    -  Browser Connected: http://127.0.0.1:8000/
[I 200505 14:44:07 watcher:104] Running task: builder (delay: 60)
INFO    -  Running task: builder (delay: 60)
INFO    -  Building documentation...
[I 200505 14:44:08 handlers:95] Reload 1 waiters: /Users/waylan/Code/mkdocs/docs/index.md
INFO    -  Reload 1 waiters: /Users/waylan/Code/mkdocs/docs/index.md
[I 200505 14:44:08 handlers:135] Browser Connected: http://127.0.0.1:8000/
INFO    -  Browser Connected: http://127.0.0.1:8000/
```

Note that it took me a few seconds to make a change to the file. However, from the timestamps in the log it took less that 60 seconds from the time it started detecting changes until the entire site was rebuilt. However the log clearly shows that the builder has a delay: `Running task: builder (delay: 60)`.

Python-Livereload's documentation clearly stated that the value is seconds, but I'm wondering if perhaps its actually milliseconds. When I tried a really high number I did see a delay.

Issue #2285, opened 2021-01-12
`wait` option does not change the reload cycle


AnjoMan:
I wrote a plugin that generates files in `docs_dir`, the general structure of which is:

```python
class AutoDocumentExtension(mkdocs.plugins.BasePlugin):
    def on_pre_build(self, config):
        # generate some markdown and insert it into the docs tree
        with open(Path(config["docs_dir"]) / "special_folder" / "autogenerated_docs.md", "w") as f:
            f.write("# Automatic High-Quality Content!")
```

I have been trying to use the `wait` keyword argument (e.g. `mkdocs serve --wait 60) to keep the live-reload feature from watching my auto-generated files before they are updated after a reload, since this would cause a build loop. However, it seems that the wait number has no effect -- as noted in https://github.com/mkdocs/mkdocs/issues/2284, even if i make the delay extremely long (50,000 seconds) there is no effect on the cycle time. 

i think this may come down to an issue with the `livereload` package, although it provides the delay option there appears to be no unit tests and its not clear that the feature works. Perhaps it should be rolled back until we can find a better approach for achieving  this wait?

pawamoy:
Following up here as asked in #2093.

From what I understand, the situation I describe in #2093 (and that is originally described in #2061) is not the same as what is described here.

OP here seems to be auto-generating a file early in the build process. Apparently it triggers the livereload watcher, so the docs are immediately rebuilt after the first build. And then again since the second build also triggered the watcher. This is why OP mentions a *build loop*. Am I correct @AnjoMan? If I am, then yes, a delay here is the solution: discard watcher events during the first N seconds. This way the file auto-generation does not trigger a rebuild, and we do not enter a build loop.

But even if we use a delay, multiple events occurring after that delay will trigger multiple rebuilds, instead of just one (i.e. a *batching update*, mentioned by @squidfunk in #2061). What if a watched directory generates 1000 events in the span of 10 seconds? We have to set a 10 seconds delay? Then for every single change we do (for example saving a Markdown page), we'll have to wait 10 seconds before the reload? I'm just asking to make sure I correctly understand the situation.

So if I understand correctly, we need another option to wait until a specified period of inactivity is reached, i.e. only trigger the build once no changes were detected since the last N seconds.

I do agree that we must check if the delay feature works in upstream anyway :slightly_smiling_face:

waylan:
[...]

PR #2357, opened 2021-04-08
Ensure only one build runs at a time.

waylan:
This reverts #2093 (by removing the `--wait` option) and adds some logic to the `builder` so that only one build will run at a time.

It I could get some reports back from users who are experiencing the issues involved that would be helpful. Does this resolve the problem for you?

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -39,15 +39,6 @@ otherkey: !ENV [VAR_NAME, FALLBACK_VAR, 'default value']
 See [Environment Variables](../user-guide/configuration.md#environment-variables)
 in the Configuration documentation for details.
 
-#### A `--wait` flag has been added to the `serve` command (#2061)
-
-To delay a rebuild of the site when using the livereload server, use the
-`--wait` flag to specify the number of seconds to wait.
-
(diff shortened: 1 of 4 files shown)


oprypin:
Could you just revert #2093 separately first? To make it clearer both in the current review, and for future (commit history).
The `wait` option indeed didn't seem to solve anything.

oprypin (changes requested):
I don't think this PR does anything. One MkDocs process already has no circumstances under which it would run multiple builds at the same time.

waylan:
I just tested this using the sample plugin provided by @AnjoMan in [this comment][1] and confirmed that this does not resolve the issue. It appears that livereload does not call `builder` until after the previous call to `builder` completes. The only way this can be addressed is from livereload directly.

> Could you just revert #2093 separately first? 

Perhaps. Although, there is the `--wait forever` option which is documented as a way to address this issue. However, that seems to suggest that the `builder` would never be called, so that doesn't help either. I can't understand what use case livereload's `wait` feature is addressing.

[1]: https://github.com/mkdocs/mkdocs/issues/2285#issue-784614484

Issue #1888, opened 2019-10-19
Disable navigation warning


adibrastegarnia:
Hi,

Is it possible to disable navigation warning? I mean this message:
The following pages exist in the docs directory, but are not included in the "nav" configuration:

waylan:
That is an "info" message, not a warning. The only benefit to disabling it from being displayed it that it is not displayed. In other words, it does not cause "strict" mode to fail, for example.

Interestingly, some have requested the opposite (see #1755).

I guess I'll consider this a request for a "quite" option.

isedwards:
It would be useful to be able to mark a page as intentionally not included in the nav.  When new people contribute to the docs they don't know which pages are intentionally excluded and which may have been missed by accident.

NicolasMassart:
I'd like that as for me, the root `index.md` at least should not raise a warning as I don't need it in the nav...

majkinetor:
I have similar warning I want to hide.

I have bunch of files repoted as warning since one plugin keeps internal stuff in the docs structure. Its houndreeds of warnings like this:

![image](https://user-images.githubusercontent.com/85767/192093274-ba398345-06b4-45ae-b01d-c41656df9958.png)

Ideally I would be able to say that there should be no warning for path `/db/*`.

oprypin:
> I have bunch of files repoted as warning since one plugin keeps internal stuff in the docs structure

That's just the plugins fault. If you want help fixing it, you can point me to the issue.

oprypin:
Also that's not even what the current issue is about.

majkinetor:
@oprypin , its effectivelly the same problem - warning spam. Its IMO irrevelant what is the precise reason.

> That's just the plugins fault. If you want help fixing it, you can point me to the issue.

Its not plugins fault, its the way I use it, and it works great. In short, some pages describing the database automatically look for sql definition and txt examples in a same folder where md page is located and adds links to those files to the md page (also adding content to the md page). Except for huge amount of warnings there is absolutelly no reason to move those sql and txt files to other place. 

The warning is also incorrect and possibly even a bug - files are found among documentation.

But I digress, the issue is warning spam, when you know the warnings are irrelevant. Majority of systems allow for disabling specific or all of the warnings.

oprypin:
I am not dismissing that these warnings should be able to be silenced. I'll handle that in the upcoming months.

But please don't also dismiss what I am saying.

> The warning is also incorrect and possibly even a bug - files are found among documentation.

[...]

Issue #1577, opened 2018-08-04
Possible addition to deploying documentation


c0gent:
I put together a short guide for a work mate:

https://c0gent.github.io/mkdocs-info/

Would any of the information listed there be useful within the user guide? Perhaps a 'GitHub Pages & Travis - Quick Start' section? If so I can submit a PR.

waylan:
Possibly, but I couldn't say without seeing what you had in mind.

c0gent:
Whoops, pasted the wrong link :)

https://c0gent.github.io/mkdocs-info/

waylan:
Wow that very handy to have.  However, while interesting, the first section doesn't really fit with deployment. And the Travis section is just normal Travis stuff. Only the last line of `.travis.yml` file is MkDocs specific. 

It occurs to me that the existing docs mostly assume the user is already familiar with git/GitHub and MkDocs is the only new piece of the user's workflow. However your guide is written for a complete newbie, which has a certain value. Just not sure where or how it fits into the docs.

I wonder if it might make more sense on the [wiki](https://github.com/mkdocs/mkdocs/wiki). Not really sure though.

c0gent:
Sure, I was thinking more just the Travis section might be of use just copy pasted somewhere. I don't need any attribution or anything.

c0gent:
Or just a link to the existing gh-pages site from somewhere. It will exist there until GitHub shuts down :P

PR #2364, opened 2021-04-14
Refactor Logging and Error handling

waylan:
This provides for better reporting to users, makes for easier maintenance going forward, and will allow future expandability. The following major changes were made:

1. The logging config is now defined in a way which will allow future expandability. For example, we could add additional levels of `strict` or `verbosity`. All log messages are fed to a single log, which has two handlers (one for strict mode and one for verbosity). All level restrictions are defined per handler so that one restriction does not interfere with another. As filters do not have a level assigned to them, the previous count filter is now a nullhandler which counts all messages is receives (restricted only by level). The previous filter still exists (as a do-nothing instance) for any third-party plugins which called it. Those plugins should remove the call going forward.
2. A new exception `mkdocs.exceptions.Abort` was added. Rather than calling `SystemExit`, we now call `Abort`, which is a custom `click` exception that aborts with a readable error message and no traceback. Where appropriate, we catch errors, log them, and then call `Abort`. This allows us to remove all error handling from the CLI code and include it where errors originate from. Of course, as previously, unanticipated errors will still generate a traceback to allow debugging.
3. Log formatting has been improved. All log messages are now hard wrapped and indented for easier reading and color is used to identify each by type. Note that colors are only used to highlight errors, warnings and debug messages. Under normal operation with no issues, the user will not see any color.

### DIFF
--- a/docs/about/release-notes.md
+++ b/docs/about/release-notes.md
@@ -61,15 +61,17 @@ to execute code when an exception is raised while building the site.
 See [`on_build_error`](../dev-guide/plugins.md#on_build_error)
 in the Plugins documentation for details.
 
-#### Two new exceptions: BuildError and PluginError (#2103)
+#### Three new exceptions: BuildError PluginError and Abort (#2103)
 
-MkDocs now has two new exceptions defined in `mkdocs.exceptions`,
-`BuildError` and `PluginError`:
+MkDocs now has tree new exceptions defined in `mkdocs.exceptions`:
+`BuildError`, `PluginError`, and `Abort`:
 
 * `PluginError` can be raised from a plugin
   to stop the build and log an error message *without traceback*.
 * `BuildError` should not be used by third-party plugins developers
   and is reserved for internal use only.
+* `Abort` is used intenrally to abort the build and display an error
+  without a traceback.
 
 See [`Handling errors`](../dev-guide/plugins.md#handling-errors)
 in the Plugins documentation for details.
@@ -103,6 +105,11 @@ The `mkdocs.config.DEFAULT_SCHEMA` global variable has been replaced with the
 function `mkdocs.config.defaults.get_schema()`, which ensures that each
 instance of the configuration is unique (#2289).
 
+The `mkdocs.utils.warning_filter` is deprecated and now does nothing. Plugins
+ should remove any reference to is as it may be deleted in a future release.
+ To ensure any warnings get counted, simply log them to the `mkdocs` log (i.e:
+`mkdocs.plugins.pluginname`).
+
 ### Other Changes and Additions to Version 1.2
 
 * Bugfix: Properly process navigation child items in `_get_by_type` when
@@ -122,6 +129,8 @@ instance of the configuration is unique (#2289).
 * Correct documentation of `pages` template context variable (#1736).
 * The `lunr` dependency has been updated to 0.5.9, and `lunr.js` to
   the corresponding 2.3.9 version (#2306).
+* Color is now used in log messages to identify errors, warnings and debug
+  messages.
 
 ## Version 1.1.2 (2020-05-14)
 
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -4,6 +4,7 @@ import os
 import sys
 import logging
 import click
+import textwrap
 
 from mkdocs import __version__
 from mkdocs import utils
@@ -12,6 +13,33 @@ from mkdocs import config
 from mkdocs.commands import build, gh_deploy, new, serve
 
 log = logging.getLogger(__name__)
+
+
+class ColorFormatter(logging.Formatter):
+    colors = {
+        'CRITICAL': 'red',
+        'ERROR': 'red',
+        'WARNING': 'yellow',
+        'DEBUG': 'blue'
+    }
+
+    def format(self, record):
+        prefix = f'{record.levelname:<8} -  '
+        if record.levelname in self.colors:
+            prefix = click.style(prefix, fg=self.colors[record.levelname])
+        lines = textwrap.wrap(record.getMessage(), width=68)
+        lines[0] = prefix + lines[0]
+        msg = '\n            '.join(lines)
+        return msg
+
+
+class ClickHandler(logging.Handler):
+    def emit(self, record):
+        try:
+            msg = self.format(record)
+            click.echo(msg)
+        except Exception:
+            self.handleError(record)
 
 
 class State:
@@ -23,9 +51,8 @@ class State:
         self.logger.setLevel(1)
         self.logger.propagate = False
 
-        self.stream = logging.StreamHandler()
-        formatter = logging.Formatter("{levelname:<7} -  {message} ", style='{')
-        self.stream.setFormatter(formatter)
+        self.stream = ClickHandler()
+        self.stream.setFormatter(ColorFormatter())
         self.stream.setLevel(level)
         self.stream.name = 'MkDocsStreamHandler'
         self.logger.addHandler(self.stream)
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -24,7 +24,7 @@ class State:
         self.logger.propagate = False
 
         self.stream = logging.StreamHandler()
-        formatter = logging.Formatter("%(levelname)-7s -  %(message)s ")
+        formatter = logging.Formatter("{levelname:<7} -  {message} ", style='{')
         self.stream.setFormatter(formatter)
         self.stream.setLevel(level)
         self.stream.name = 'MkDocsStreamHandler'
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -8,7 +8,6 @@ import textwrap
 
 from mkdocs import __version__
 from mkdocs import utils
-from mkdocs import exceptions
 from mkdocs import config
 from mkdocs.commands import build, gh_deploy, new, serve
 
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -27,6 +27,7 @@ class State:
         formatter = logging.Formatter("%(levelname)-7s -  %(message)s ")
         self.stream.setFormatter(formatter)
         self.stream.setLevel(level)
+        self.stream.name = 'MkDocsStreamHandler'
         self.logger.addHandler(self.stream)
 
         # Add CountHandler for strict mode
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -19,13 +19,20 @@ class State:
 
     def __init__(self, log_name='mkdocs', level=logging.INFO):
         self.logger = logging.getLogger(log_name)
+        # Don't restrict level on logger; use handler
+        self.logger.setLevel(1)
         self.logger.propagate = False
-        stream = logging.StreamHandler()
-        formatter = logging.Formatter("%(levelname)-7s -  %(message)s ")
-        stream.setFormatter(formatter)
-        self.logger.addHandler(stream)
 
-        self.logger.setLevel(level)
+        self.stream = logging.StreamHandler()
+        formatter = logging.Formatter("%(levelname)-7s -  %(message)s ")
+        self.stream.setFormatter(formatter)
+        self.stream.setLevel(level)
+        self.logger.addHandler(self.stream)
+
+        # Add CountHandler for strict mode
+        self.counter = utils.log_counter
+        self.counter.setLevel(logging.WARNING)
+        self.logger.addHandler(self.counter)
 
 
 pass_state = click.make_pass_decorator(State, ensure=True)
@@ -69,7 +76,7 @@ def verbose_option(f):
     def callback(ctx, param, value):
         state = ctx.ensure_object(State)
         if value:
-            state.logger.setLevel(logging.DEBUG)
+            state.stream.setLevel(logging.DEBUG)
     return click.option('-v', '--verbose',
                         is_flag=True,
                         expose_value=False,
@@ -81,7 +88,7 @@ def quiet_option(f):
     def callback(ctx, param, value):
         state = ctx.ensure_object(State)
         if value:
-            state.logger.setLevel(logging.ERROR)
+            state.stream.setLevel(logging.ERROR)
     return click.option('-q', '--quiet',
                         is_flag=True,
                         expose_value=False,
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -167,15 +167,7 @@ def serve_command(dev_addr, livereload, **kwargs):
 
     logging.getLogger('tornado').setLevel(logging.WARNING)
 
-    try:
-        serve.serve(
-            dev_addr=dev_addr,
-            livereload=livereload,
-            **kwargs
-        )
-    except (exceptions.ConfigurationError, OSError) as e:  # pragma: no cover
-        # Avoid ugly, unhelpful traceback
-        raise SystemExit('\n' + str(e))
+    serve.serve(dev_addr=dev_addr, livereload=livereload, **kwargs)
 
 
 @cli.command(name="build")
@@ -185,12 +177,7 @@ def serve_command(dev_addr, livereload, **kwargs):
 @common_options
 def build_command(clean, **kwargs):
     """Build the MkDocs documentation"""
-
-    try:
-        build.build(config.load_config(**kwargs), dirty=not clean)
-    except exceptions.ConfigurationError as e:  # pragma: no cover
-        # Avoid ugly, unhelpful traceback
-        raise SystemExit('\n' + str(e))
+    build.build(config.load_config(**kwargs), dirty=not clean)
 
 
 @cli.command(name="gh-deploy")
@@ -206,17 +193,13 @@ def build_command(clean, **kwargs):
 @common_options
 def gh_deploy_command(clean, message, remote_branch, remote_name, force, ignore_version, shell, **kwargs):
     """Deploy your documentation to GitHub Pages"""
-    try:
-        cfg = config.load_config(
-            remote_branch=remote_branch,
-            remote_name=remote_name,
-            **kwargs
-        )
-        build.build(cfg, dirty=not clean)
-        gh_deploy.gh_deploy(cfg, message=message, force=force, ignore_version=ignore_version, shell=shell)
-    except exceptions.ConfigurationError as e:  # pragma: no cover
-        # Avoid ugly, unhelpful traceback
-        raise SystemExit('\n' + str(e))
+    cfg = config.load_config(
+        remote_branch=remote_branch,
+        remote_name=remote_name,
+        **kwargs
+    )
+    build.build(cfg, dirty=not clean)
+    gh_deploy.gh_deploy(cfg, message=message, force=force, ignore_version=ignore_version, shell=shell)
 
 
 @cli.command(name="new")
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -443,3 +443,8 @@ class CountHandler(logging.NullHandler):
 
 # A global instance to use throughout package
 log_counter = CountHandler()
+
+# For backward compatability as some plugins import it.
+# It is no longer nessecary as all messages on the
+# `mkdocs` logger get counted automaticaly.
+warning_filter = logging.Filter()
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -15,6 +15,7 @@ import fnmatch
 import posixpath
 import functools
 import importlib_metadata
+from collections import defaultdict
 from datetime import datetime, timezone
 from urllib.parse import urlparse
 from yaml_env_tag import construct_env_tag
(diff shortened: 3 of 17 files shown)


mkdocs/__main__.py
```diff
@@ -4,28 +4,62 @@
...
+        'ERROR': 'red',
+        'WARNING': 'yellow',
+        'DEBUG': 'blue'
+    }
+
+    def format(self, record):
+        prefix = f'{record.levelname:<8} -  '
+        if record.levelname in self.colors:
+            prefix = click.style(prefix, fg=self.colors[record.levelname])
+        lines = textwrap.wrap(record.getMessage(), width=68)
```
oprypin, line 29: Why wrap, *and* with this super specific choice of width? This looks horrible.

image

mkdocs/utils/__init__.py
```diff
@@ -422,15 +423,28 @@ def nest_paths(paths):
     return nested
 
 
-class WarningFilter(logging.Filter):
-    """ Counts all WARNING level log messages. """
-    count = 0
+class CountHandler(logging.NullHandler):
+    """ Counts all logged messages >= level. """
```
oprypin, line 427: Cheers for fixing it to be `>=` rather than `==` WARNING

docs/about/release-notes.md
```diff
@@ -61,15 +61,17 @@ to execute code when an exception is raised while building the site.
...
-MkDocs now has two new exceptions defined in `mkdocs.exceptions`,
-`BuildError` and `PluginError`:
+MkDocs now has tree new exceptions defined in `mkdocs.exceptions`:
+`BuildError`, `PluginError`, and `Abort`:
 
 * `PluginError` can be raised from a plugin
   to stop the build and log an error message *without traceback*.
 * `BuildError` should not be used by third-party plugins developers
   and is reserved for internal use only.
+* `Abort` is used intenrally to abort the build and display an error
```
oprypin, line 73: Typo "intenrally"

mkdocs/utils/__init__.py
```diff
@@ -422,15 +423,28 @@ def nest_paths(paths):
...
+    def get_counts(self):
+        return [(logging.getLevelName(k), v) for k, v in sorted(self.counts.items(), reverse=True)]
 
 
 # A global instance to use throughout package
-warning_filter = WarningFilter()
+log_counter = CountHandler()
+
+# For backward compatability as some plugins import it.
+# It is no longer nessecary as all messages on the
```
oprypin, line 448: Typo "nessecary"

mkdocs/__main__.py
waylan: > This looks horrible.

That is a matter of opinion. Regardless, I chose that width because I still run my terminal with a width of 80. With the indentation of 12, that leaves a width of 68. It is much easier to read with the indentation than without.

oprypin: In this example not a matter of opinion because the wrapping is literally incorrect and it's very hard to deduce what this was supposed to be.

But if generally it is a matter of opinion, then indeed, why should you enforce your opinion?

Can you name me another tool that wraps its output at 68 characters? Especially in logs???

For reference, my terminal is usually around 230 characters wide.

waylan: You seem to be missing the purpose of this. So let me illustrate with the screenshot below.

Screen Shot 2021-04-18 at 5 26 22 PM

The first `build` was run with the text wrap enabled. The second was with it disabled. Notice how much easier it is to read the first. The difference is that the left side is indented on secondary lines. I suppose if you have a wider terminal window that may not be so evident. But it was a huge improvement for me. 

If you have a better way to get the same effect, I'm open to suggestions. But barring anything better, the current behavior is an improvement over what we had.

I don't know, but is there an easy way to obtain the terminal width from Python? If so we could wrap based on that and no-one would get what they perceive to be weird results.

oprypin: > You seem to be missing the purpose of this.

No of course I'm not missing it. I just don't want it, it's a downgrade in almost every case.

It changes from an unambiguous representation to an ambiguous one. (see my example)

It breaks line-based searches (yes, I have done `grep` on the errors) and line counts and just everything.

Other tools just don't do this!

And why does the 80 char terminal suddenly become so important *now* after so many years. Did anyone else at all mention this?

waylan: > And why does the 80 char terminal suddenly become so important now after so many years. Did anyone else at all mention this?

It's not 80 chars specifically, but the general motivation for all of the changes here are related to #1577 and #1888. As you can see, those two requests are in opposition to each other. In the case of #1888, the primary issue is that the user is misunderstanding the logs. So my goal was to try to make the logs more clear and easier to read. The color certainly helps by better differentiating between warnings and informational messages. But, at the same time, new users often simply see a wall of text and don't try to understand it. By providing some formatting (indenting secondary lines), it makes it easier to read However, the only way to indent the lines is to also hard-wrap them.

Truth be told,  when I set out I didn't have text wrapping in mind at all. However, as I was working on it, it occurred to me that it might help and I tried it. I was surprised how much easier it made the logs to read, so I moved on and never gave it another thought.

Things like greping logs never occurred to me. After all, the formatter is specifically on the StreamHandler which writes to the terminal. If someone is writing to a log file, there technically should be a different handler which does that. True, a user could be redirecting std.out/std.error to a file, but that means the user is most likely using the server in something other than a local dev environment, which we have never officially supported. In that case, you should be using a different server anyway.

My point is that the only use case that we officially support is to have the logs written directly to the terminal. In that case there is little harm in splitting lines (I suppose you could grep the output of `mkdocs build`, but who has that much output from a single build?). And if I had thought to call `shutil.get_terminal_size` to calculate line length, you may have never even noticed so long as you are using the server as it is officially supported.

To be clear, I am not opposed to removing the textwrapping for a good reason. I just haven't seen one that fits into the officially supported use case. In my view, the change is such a huge improvement that I'm okay with leaving it despite it being a little unconventional. 

As a compromise, perhaps we could use `shutil.get_terminal_size` with a `fallback` of `None` (or `(0, 0)` or some other obviously invalid value). That way, if `shutil.get_terminal_size` returns the fallback, then we can assume there is no "terminal" and skip the wrapping. However, if a valid value is returned, we can assume the user is on the terminal and then we can wrap the text for readability using the column width. Of course, this assumes `shutil.get_terminal_size` isn't able to obtain a terminal size when MkDocs is run as a system service or from a docker instance or whatever other unsupported environments users may be using. I may be wrong about that assumption, in which case, relying on the fallback won't help. That may require some investigation.

oprypin: > the general motivation for all of the changes here are related to

OK but those are only tangential points for your motivation, not someone actually asking for (or even mentioning) that.

> In the case of #1888, the primary issue is that the user is misunderstanding the logs

Hm? I don't see any message there that can be interpreted that way.

> the user is most likely using the server in something other than a local dev environment, which we have never officially supported

> I suppose you could grep the output of mkdocs build, but who has that much output from a single build?

Please let's not turn this into "you're holding it wrong".

> In that case, you should be using a different server anyway.

Took me a while to realize that you mean `mkdocs serve`. But that's not a strong point, as all of this applies to `mkdocs build` too.

oprypin: > However, if a valid value is returned, we can assume the user is on the terminal and then we can wrap the text for readability using the column width.

Yes, this approach would probably be (barely) acceptable to me.

```shell
$ python3 -c 'import shutil; print(shutil.get_terminal_size(fallback=(0, 0)))'      
os.terminal_size(columns=195, lines=42)

$ python3 -c 'import shutil; print(shutil.get_terminal_size(fallback=(0, 0)))' | cat
os.terminal_size(columns=0, lines=0)
```

Even so, I still think the `textwrap` implementation is not safe enough for technical text. As you can see in my screenshot, it shows `--project-` `version=` and there's no way to tell whether that was `--project- version=` or `--project-version=`.

Additionally, I can no longer select the whole log line by triple-clicking it, and even after I select things manually, it still includes extraneous spaces that were not there.

So again, it's just a downgrade for me in any case, I really just prefer the terminal to do its own wrapping.

I literally don't know any tool that does this automated word wrapping. Can you point out one?

<table><tr><td></td><td></td></tr></table>

There are some tools that wrap only on known already-separated tokens. Perfect example:

<table><tr><td></td></tr></table>

Note how the package names are carefully wrapped without splitting them across several lines on the dashes.
And then note how textual messages are *not* wrapped, *although* (or, as I'm convinced, *because*) clearly a lot of thought was put into this UI.

waylan: > I literally don't know any tool that does this automated word wrapping.

MkDocs' own `--help` does. But then that is just the default behavior of [click][1], the lib we use for option parsing.

```
$ mkdocs build --help
Usage: mkdocs build [OPTIONS]

  Build the MkDocs documentation

Options:
  -c, --clean / --dirty           Remove old files from the site_dir before
                                  building (the default).

  -f, --config-file FILENAME      Provide a specific MkDocs config
  -s, --strict                    Enable strict mode. This will cause MkDocs
                                  to abort the build on any warnings.

  -t, --theme [mkdocs|readthedocs]
                                  The theme to use when building your
                                  documentation.

  --use-directory-urls / --no-directory-urls
                                  Use directory URLs when building pages (the
                                  default).

  -d, --site-dir PATH             The directory to output the result of the
                                  documentation build.

  -q, --quiet                     Silence warnings
  -v, --verbose                   Enable verbose output
  -h, --help                      Show this message and exit.
```

Interestingly, that is hardwrapped at 80. And we didn't choose that. That is click's default behavior. Now, I know `--help` messages are not logs, but this is pretty standard for `--help`. It is also where I got the idea from to begin with.

[1]: https://click.palletsprojects.com/

mkdocs/__main__.py
```diff
@@ -4,28 +4,62 @@
...
+        lines[0] = prefix + lines[0]
+        msg = '\n            '.join(lines)
```
oprypin, line 31: Seems like this custom behavior can be achieved by just using [`subsequent_indent`](https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.subsequent_indent)

mkdocs/__main__.py
oprypin: I think all of these flags would be very important improvements if this implementation is here to stay.

* [replace_whitespace]=False
* **[drop_whitespace]=False**
* [break_long_words]=False
* **[break_on_hyphens]=False** (addresses my biggest complaint)

[replace_whitespace]: https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.replace_whitespace
[drop_whitespace]: https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.drop_whitespace
[break_long_words]: https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.break_long_words
[break_on_hyphens]: https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.break_on_hyphens

I checked that GitHub Actions also doesn't report a terminal width, so that's good from the perspective of this autodetection.
[Did it in a random repo, disregard the rest](https://github.com/oprypin/mkdocs-literate-nav/runs/2384012316?check_suite_focus=true#step:4:1)

oprypin:
Tried out some error output and it's not looking good at all.
I have no idea what the wrapper is doing, but it's completely unfit for the job of wrapping text that already has multiple lines (maybe it just counts them as a normal character?)

Before:

image


After (including #2376)

image

waylan:
So there are two different paths for errors to take:

1. Known (expected) errors (usually based on bad user input) are caught. logged, and the build is aborted without any traceback. This is the desired behavior as non-developer users find tracebacks intimidating and often completely miss the obvious (to me or you) error message telling them to fix a config setting (or whatever problem they are having).
2. Unexpected errors are not caught or logged and the traceback is output directly to std.error and not passed through the logger. Therefore, the logger formatting should have no effect on those errors. Of course, these errors should be eliminated by good testing during development and are therefore not likely to be encountered by users. However, they are necessary for debugging during development and that is why we don't catch all errors, or when we do, we reraise the unexpected errors. In the case where the error is reraised, we first log the message (without the traceback). So again, the logging formatter should have no effect on it.

In my experience, any unexpected errors are behaving as described above. I'm not sure what you did to generate the above error, so it makes it hard to replicate. Additionally, it appears that you have 3 errors. However, MkDocs should always abort on one error. So maybe we have a very specific error handling bug here.

oprypin:
Those are just two `logger.error` followed by `raise PluginError`.
This is indeed an "unexpected error" and the traceback is intentionally perserved in the code.

waylan:
Please file a bug report.

pawamoy:
The warning in the screenshot above comes from pip, it doesn't pass through mkdocs loggers, this is why it's not wrapped. The traceback however is part of the log message, so this is why it is wrapped. So I guess there's no formatting issue here? But correct me if I'm wrong!

waylan:
> The traceback however is part of the log message, so this is why it is wrapped.

But is shouldn't be.  Nowhere should we be passing a traceback to the logger. If we are, then that is the bug which should be fixed. Not removing the changes we made which @oprypin keeps objecting to. I'm not backing those changes out. But if he has identified a bug (where a traceback is being passed to the logger), we will fix that bug.

pawamoy:
The traceback comes from a subprocess. In this case it's a Python subprocess but it could be anything else. How should we log a formatted multi-line message with mkdocs without breaking its format?

oprypin:
Another before / after comparison. Usage of `mkdocs serve` where `PluginError` is not raised initially but is raised on a rebuild.

Before:
```
INFO    -  Building documentation... 
INFO    -  Documentation built in 48.29 seconds 
[I 210425 17:50:08 server:335] Serving on http://127.0.0.1:8000
INFO    -  Serving on http://127.0.0.1:8000
[I 210425 17:50:08 handlers:62] Start watching changes
INFO    -  Start watching changes
[I 210425 17:50:08 handlers:64] Start detecting changes
INFO    -  Start detecting changes
[I 210425 17:50:13 watcher:110] Running task: builder (delay: None)
INFO    -  Running task: builder (delay: None)
INFO    -  Building documentation... 

'shard.yml' not found anywhere above ''
```
**[exits with error 1]**

After:

```
INFO     -  Documentation built in 41.10 seconds
[I 210425 17:51:33 server:335] Serving on http://127.0.0.1:8000
INFO    -  Serving on http://127.0.0.1:8000
[I 210425 17:51:33 handlers:62] Start watching changes
INFO    -  Start watching changes
[I 210425 17:51:33 handlers:64] Start detecting changes
INFO    -  Start detecting changes
[I 210425 17:51:39 watcher:110] Running task: builder (delay: None)
INFO    -  Running task: builder (delay: None)
ERROR    -  'shard.yml' not found anywhere above ''
[E 210425 17:51:41 ioloop:907] Exception in callback <bound method LiveReloadHandler.poll_tasks of <class 'livereload.handlers.LiveReloadHandler'>>
      File "/home/blaxpirit/.local/lib/py/mkdocs/commands/build.py", line 274, in build
        files = config['plugins'].run_event('files', files, config=config)
      File "/home/blaxpirit/.local/lib/py/mkdocs/plugins.py", line 94, in run_event
        result = method(item, **kwargs)
      File "/home/blaxpirit/.local/lib/py/mkdocs_gen_files/plugin.py", line 32, in on_files
        runpy.run_path(file_name)
      File "/usr/lib/python3.9/runpy.py", line 268, in run_path
        return _run_module_code(code, init_globals, run_name,
      File "/usr/lib/python3.9/runpy.py", line 97, in _run_module_code
        _run_code(code, mod_globals, init_globals,
      File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "docs/gen_doc_stubs.py", line 54, in <module>
        for loc in typ.locations
      File "/home/blaxpirit/.local/lib/python3.9/site-packages/cached_property.py", line 36, in __get__
        value = obj.__dict__[self.func.__name__] = self.func(obj)
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/items.py", line 232, in locations
        return [
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/items.py", line 233, in <listcomp>
        self.root.update_url(DocLocation(loc["filename"], loc["line_number"], loc["url"]))
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 168, in update_url
        location.url = dest.substitute(location)
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 131, in substitute
        return self.dest_url.format_map(collections.ChainMap(data, self))  # type: ignore
      File "/usr/lib/python3.9/collections/__init__.py", line 938, in __getitem__
        return mapping[key]             # can't use 'key in mapping' with defaultdict
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 139, in __getitem__
        return getattr(self, key)
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 145, in shard_version
        return self._shard_version(os.path.dirname(self.src_path))
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 159, in _shard_version
        return cls._shard_version(os.path.dirname(path))
      File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 158, in _shard_version
        raise PluginError(f"'shard.yml' not found anywhere above {path!r}")
    mkdocs.exceptions.PluginError: 'shard.yml' not found anywhere above ''
    
    During handling of the above exception, another exception occurred:
    
      File "/home/blaxpirit/.local/lib/python3.9/site-packages/tornado/ioloop.py", line 905, in _run
        return self.callback()
      File "/home/blaxpirit/.local/lib/python3.9/site-packages/livereload/handlers.py", line 69, in poll_tasks
        filepath, delay = cls.watcher.examine()
      File "/home/blaxpirit/.local/lib/python3.9/site-packages/livereload/watcher.py", line 119, in examine
        func()
      File "/home/blaxpirit/.local/lib/py/mkdocs/commands/serve.py", line 138, in builder
        build(config, live_server=live_server, dirty=dirty)
      File "/home/blaxpirit/.local/lib/py/mkdocs/commands/build.py", line 323, in build
        raise Abort('\nAborted with a BuildError!')
    mkdocs.exceptions.Abort: 
    Aborted with a BuildError!
ERROR   -  Exception in callback <bound method LiveReloadHandler.poll_tasks of <class 'livereload.handlers.LiveReloadHandler'>>
  File "/home/blaxpirit/.local/lib/py/mkdocs/commands/build.py", line 274, in build
    files = config['plugins'].run_event('files', files, config=config)
  File "/home/blaxpirit/.local/lib/py/mkdocs/plugins.py", line 94, in run_event
    result = method(item, **kwargs)
  File "/home/blaxpirit/.local/lib/py/mkdocs_gen_files/plugin.py", line 32, in on_files
    runpy.run_path(file_name)
  File "/usr/lib/python3.9/runpy.py", line 268, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/usr/lib/python3.9/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "docs/gen_doc_stubs.py", line 54, in <module>
    for loc in typ.locations
  File "/home/blaxpirit/.local/lib/python3.9/site-packages/cached_property.py", line 36, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/items.py", line 232, in locations
    return [
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/items.py", line 233, in <listcomp>
    self.root.update_url(DocLocation(loc["filename"], loc["line_number"], loc["url"]))
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 168, in update_url
    location.url = dest.substitute(location)
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 131, in substitute
    return self.dest_url.format_map(collections.ChainMap(data, self))  # type: ignore
  File "/usr/lib/python3.9/collections/__init__.py", line 938, in __getitem__
    return mapping[key]             # can't use 'key in mapping' with defaultdict
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 139, in __getitem__
    return getattr(self, key)
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 145, in shard_version
    return self._shard_version(os.path.dirname(self.src_path))
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 159, in _shard_version
    return cls._shard_version(os.path.dirname(path))
  File "/home/blaxpirit/.local/lib/py/mkdocstrings/handlers/crystal/collector.py", line 158, in _shard_version
    raise PluginError(f"'shard.yml' not found anywhere above {path!r}")
mkdocs.exceptions.PluginError: 'shard.yml' not found anywhere above ''

During handling of the above exception, another exception occurred:

  File "/home/blaxpirit/.local/lib/python3.9/site-packages/tornado/ioloop.py", line 905, in _run
    return self.callback()
  File "/home/blaxpirit/.local/lib/python3.9/site-packages/livereload/handlers.py", line 69, in poll_tasks
    filepath, delay = cls.watcher.examine()
  File "/home/blaxpirit/.local/lib/python3.9/site-packages/livereload/watcher.py", line 119, in examine
    func()
  File "/home/blaxpirit/.local/lib/py/mkdocs/commands/serve.py", line 138, in builder
    build(config, live_server=live_server, dirty=dirty)
  File "/home/blaxpirit/.local/lib/py/mkdocs/commands/build.py", line 323, in build
    raise Abort('\nAborted with a BuildError!')
mkdocs.exceptions.Abort: 
Aborted with a BuildError!
[I 210425 17:52:03 handlers:135] Browser Connected: http://127.0.0.1:8000/api/Crystal/Program/CompiledMacroRun.html
```
**[keeps running]**



Other than the obvious huge traceback, the other problem I'm showing here is that log lines from livereload don't line up with log lines from mkdocs. And that in one case `mkdocs serve` aborts and in the other one it doesn't.

waylan:
Sorry but I am not able to to divine from your traceback how you encountered it? If you have found a bug, please open a new bug report with sufficient information for me (or anyone else) to be able to reproduce the bug.

PR #2381, opened 2021-04-23
Revert the addition of Abort exception and clean up abort messages

oprypin:
I see no upside to having that exception type, but the downside is that it always requires some message, even when we have nothing additional to say.

I also revert some recently added exclamation marks and fix messages where each word is capitalized.


waylan:
I know you don't like those changes, but please stop trying to find reasons to back them out.  If there is a genuine bug, then let's fix that bug. Otherwise, please stop wasting my time.

oprypin:
What are "those changes"? I haven't even said anything so far on the particular topic that this PR addresses. But I guess you haven't even looked at the content.

PR #2382, opened 2021-04-23
Fix wrapping of multiline strings for the newly added ColorFormatter

oprypin:
1. Fix handling of messages that already have multiple lines.
  It looks like the wrapping code in stdlib just considers '\n' like any other character, so the result is really weird.
2. ~~Decrease indentation by one, to match the default logger.~~
3. ~~Semi-related: Drop one case of manual indentation now that it's centrally handled.~~

See examples of (1) and (2) at https://github.com/mkdocs/mkdocs/pull/2364#issuecomment-824381108

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -28,19 +28,22 @@ class ColorFormatter(logging.Formatter):
         replace_whitespace=False,
         break_long_words=False,
         break_on_hyphens=False,
-        initial_indent=' '*12,
-        subsequent_indent=' '*12
+        initial_indent=' '*11,
+        subsequent_indent=' '*11
     )
 
     def format(self, record):
-        prefix = f'{record.levelname:<8} -  '
+        prefix = f'{record.levelname:<7} -  '
         if record.levelname in self.colors:
             prefix = click.style(prefix, fg=self.colors[record.levelname])
         if self.text_wrapper.width:
             # Only wrap text if a terminal width was detected
-            msg = self.text_wrapper.fill(record.getMessage())
+            msg = '\n'.join(
+                self.text_wrapper.fill(line)
+                for line in record.getMessage().splitlines()
+            )
             # Prepend prefix after wrapping so that color codes don't affect length
-            return prefix + msg[12:]
+            return prefix + msg[11:]
         return prefix + record.getMessage()
 
 
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -28,12 +28,12 @@ class ColorFormatter(logging.Formatter):
         replace_whitespace=False,
         break_long_words=False,
         break_on_hyphens=False,
-        initial_indent=' '*11,
-        subsequent_indent=' '*11
+        initial_indent=' '*12,
+        subsequent_indent=' '*12
     )
 
     def format(self, record):
-        prefix = f'{record.levelname:<7} -  '
+        prefix = f'{record.levelname:<8} -  '
         if record.levelname in self.colors:
             prefix = click.style(prefix, fg=self.colors[record.levelname])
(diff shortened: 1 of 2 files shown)


oprypin:
Compare https://github.com/mkdocs/mkdocs/pull/2364#issuecomment-824381108 to how it looks with this change:

Screenshot_20210423_210102

waylan:
The indent was set based on the length of `CRITICAL`. Your change would cause `CRITICAL` to not be aligned with the other message types.

Also, under what circumstances are messages being logged which contain newlines? As I've stated elsewhere, we should never be logging tracebacks. If we are, then that is a bug which should be fixed at the source, not by modifying the logging formatter.

oprypin:
**Plugins** can access logging. This is from a plugin.

oprypin:
> Also, under what circumstances are messages being logged which contain newlines?

> that is a bug which should be fixed at the source

Is this one a bug?

https://github.com/mkdocs/mkdocs/blob/6a694fd87ba5576a51c4060ab111a5cec3a9fbaf/mkdocs/structure/nav.py#L114-L117

oprypin:
> Is this one a bug?

Well, to kinda answer my own question, there is indeed currently a bug with core mkdocs relating to these particular newlines.

image

oprypin:
> The indent was set based on the length of CRITICAL. Your change would cause CRITICAL to not be aligned with the other message types.

This is true. However, I think the smaller alignment is still better, considering that:

* well, it's just closer, so nicer to look at
* the previous code had the same alignment of 7 and it was fine
* CRITICAL will basically never appear anywhere
* at least one other tool (pip in my example) also happens to use alignment level of 7
    * (which makes me wonder, where did that number come from? maybe there are even more tools copying that example?)

waylan:
> 
> 
> > Also, under what circumstances are messages being logged which contain newlines?
> 
> > that is a bug which should be fixed at the source
> 
> Is this one a bug?
> 
> https://github.com/mkdocs/mkdocs/blob/6a694fd87ba5576a51c4060ab111a5cec3a9fbaf/mkdocs/structure/nav.py#L114-L117

Ah, yes, that is a valid case of a log message which includes newlines. Thank you. In the future, please provide this sort of information upfront, rather than expecting us to remember everything.

waylan:
>     * CRITICAL will basically never appear anywhere

It may not now, but that could change in the future.  In which case most of this change would need to be reverted. So lets just leave it as-is.

> Plugins can access logging. This is from a plugin.

There is specific direction about how plugins should deal with errors in the [docs][1]. If the plugin is following that direction and this is still a problem, then we have a bug. Please provide a minimal reproducible example of a plugin which exhibits the behavior and we'll work out a fix either by fixing the bug or by adjusting the documented policy. If, on the other hand, the plugin is passing a traceback into the logger, then that is something the plugin needs to fix/change. In fact, in that case, breaking the traceback is a good thing in that it is more likely to get the plugin developer of the plugin to stop including the traceback.

[1]: https://github.com/mkdocs/mkdocs/blob/master/docs/dev-guide/plugins.md#handling-errors

oprypin:
> There is specific direction about how plugins should deal with errors in the docs.

Following that doc would mean that a plugin is not allowed to emit warnings and if it emits an error, at most one line is allowed. That is too limiting.

> In fact, in that case, breaking the traceback is a good thing in that it is more likely to get the developer of the plugin to stop including the traceback.

How do we debug user-reported issues then? We can't just go and change every error message to "oops something went wrong"

oprypin:
To clarify on "oops something went wrong" -- that isn't fair to say -- the actual error message would be:
"Error parsing external JSON"  (as you see above)
"Error running a subprocess"  (which subprocess? what did it output? are we not allowed to know?)

PR #2388, opened 2021-05-02
For extra_css, use the paths already populated in create_media_urls

oprypin:


waylan:
For reference, this is related to this code, which is run for every page. Therefore, the URLs are already relative to the page and do not need to be passed through the `url` filter in the template.

https://github.com/mkdocs/mkdocs/blob/d0202a512e55a4e341b0195f1770f1ddaf41f4cf/mkdocs/commands/build.py#L39-L41

PR #2395, opened 2021-05-07
Replace pkg_resources with importlib_metadata wrt #2354

ultrabug:
As per discussion on #2347 our new babel code base should be
migrated to not require setuptools and use importlib_metadata
instead

### DIFF
--- a/mkdocs/commands/babel.py
+++ b/mkdocs/commands/babel.py
@@ -1,6 +1,6 @@
 from distutils.errors import DistutilsOptionError
 from os import path
-from pkg_resources import EntryPoint
+from importlib_metadata import entry_points
 from babel.messages import frontend as babel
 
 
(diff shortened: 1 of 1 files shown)


ultrabug:
Interestingly enough our babel tests are based on `setuptools.dist.Distribution` since babel itself relies on distutils...

Switching to importlib metadata renders some tests useless since mkdocs code base always have at least 2 `mkdocs.themes` entry points.

@waylan I'd like your opinion if you may.

waylan:
I debated about this internally when working on the code. In the end I determined that our commands classes will always require setuptools so it doesn't matter here. As a reminder, there are ways to install Python package without using the setup script, which means there are ways to install MkDocs without setuptools/pkg_resources installed. However, the commands classes only get imported into the `setup.py` script and the `theme` option will only work if an entry point is defined. And Distutils does not support entry points, only setuptools does. Therefore, it is impossible to use this code without setuptools/pkg_resources anyway.

Finally, note that the docs for importlib_metadata specifically states that its purpose is to replace calls to pkg_resources from within library code. This is not library code, but setup script code, where pkg_resources is meant to be called.  

Given the above, I never even explored trying to importlib_metadata here. If you can make it work, that's fine, but I don't think its necessary. However, if we were to make the command classes available as subcommands of the `mkdocs` command, then yes, we should be using importlib_metadata.

PR #2396, opened 2021-05-07
Add 'es' themes translations

mondeja:


mondeja:
> Could you add "Spanish" to the note in the release notes? Also, the language needs to be added to the list of supported languages for each theme in docs/user-guide/choosing-your-theme.md.

> We should probably update the translation docs to reflect the need to do that when new locales are added.

Done in https://github.com/mkdocs/mkdocs/pull/2396/commits/c869947fa8049dea42fe8753724a6c4bd79d45f0

PR #2397, opened 2021-05-07
Update IpAddress normalization test broken in Python3.9.5

mondeja:
[`ipaddress.ip_address`](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address) behaviour has been changed in Python3.9.5 due to a [security fix](https://www.ehackingnews.com/2021/05/python-affected-by-critical-ip-address.html). Now, leading zeros are not treated as part of valid ip addresses.

~~This pull proposes the change of the affected test, adding a leading zero at the beginning to make it break in earlier versions of Python, but the other option would be to delete it because this normalization would be invalid.~~

### DIFF
--- a/mkdocs/tests/config/config_options_tests.py
+++ b/mkdocs/tests/config/config_options_tests.py
@@ -132,13 +132,17 @@ class IpAddressTest(unittest.TestCase):
         self.assertEqual(value.host, '127.0.0.1')
         self.assertEqual(value.port, 8000)
 
-    def test_invalid_leading_zeros(self):
-        addr = '0127.000.000.001:8000'
+    @unittest.skipIf(
+        sys.version_info >= (3, 9, 5),
+        "Leading zeros not allowed in IP addresses since Python3.9.5",
+    )
+    def test_IP_normalization(self):
+        addr = '127.000.000.001:8000'
         option = config_options.IpAddress(default=addr)
-        self.assertRaises(
-            config_options.ValidationError,
-            option.validate, addr
-        )
+        value = option.validate(None)
+        self.assertEqual(str(value), '127.0.0.1:8000')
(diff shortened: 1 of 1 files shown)


waylan:
Let's skip the new test if < 9.5. Then we don't need the weird hack of adding an extra 0 to `127` to make the test fail. That completely defeats the purpose of the test as it could be failing for a different reason than what we are testing for.

Our tests are always run against the latest of version 9 in CI, so all PRs will still get checked on this behavior. In other words, any breakages will still be caught.

mondeja:
> Let's skip the new test if < 9.5. 

Okay, it's skipped now.

waylan:
> 
> 
> > Let's skip the new test if < 9.5.
> 
> Okay, it's skipped now.

No, you did the opposite. You skipped if >= 3.9.5, not skip if < 3.9.5. We should be testing the new behavior, not the old behavior.

mondeja:
I don't understand, the new behaviour is [this](https://github.com/mkdocs/mkdocs/runs/2527935269#step:5:30).

waylan:
Include the test you had in your first commit (`test_invalid_leading_zeros`), but only run the test for Python 3.9.5 and greater. The test should verify that the IP address `127.000.000.001:8000` causes an exception to be raised. As there will not be normalization going forward, we don't need to test that it continues to work. Therefore the preexisting test (`test_IP_normalization`) can be removed.

waylan:
@ultrabug it appears that the "translation" check is failing because this PR is not against the latest in master. There should be no failure here, except that one of the git commands in the action fails to run. If you have any ideas about how to address it, that would be great.

Regardless, I am going to merge this as the failure should have no bearing on this change.

PR #2403, opened 2021-05-12
Replace use of deprecated alias `jinja2.Markup`

oprypin:
-- use `markupsafe.Markup` directly instead.

Note: the choice of version is to match https://github.com/pallets/jinja/blob/2.10.1/setup.py

See the deprecation message already popping up in CI:
https://github.com/mkdocs/mkdocs/runs/2567454550?check_suite_focus=true#step:5:17
`DeprecationWarning: 'jinja2.Markup' is deprecated and will be removed in Jinja 3.1. Import 'markupsafe.Markup' instead.`

### DIFF
--- a/mkdocs/utils/filters.py
+++ b/mkdocs/utils/filters.py
@@ -1,11 +1,13 @@
 import json
+
 import jinja2
+import markupsafe
 
 from mkdocs.utils import normalize_url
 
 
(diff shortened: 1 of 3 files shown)


waylan:
Cool. I was just about to look into this. Thanks.

PR #2405, opened 2021-05-13
Ensure site_url and use_directory_urls do not conflict.

### LINKED ISSUE #2189: Use use_directory_urls: false breaks on empty site_url

waylan:
The site_url config option is now required. If it is set to an empty
string, then use_directory_urls will be forced to false. Each will
issue a warning if not properly set. In a future release we may
raise an error instead.

Fixes #2189.


waylan:
Through config validation, this makes it impossible to have `site_url` be empty and `use_directory_urls` be `True`. Therefore, the situation described in the #2189 will never occur.

PR #2296, opened 2021-02-06
Exclude edge cases where get_relative_url depends on CWD

oprypin:
When only one of the two passed paths starts with a slash, this function would produce results that depend on (and expose parts of) the actual current working directory.
In a similar manner, it was also possible to "break out" of the "current directory" by starting one of the paths with `../../..` etc.

Additionally, the fact that paths always end up being resolved relative to the current working directory made this function less efficient than it needs to be.
Fun fact: `get_relative_url('path_a', 'path_b')` actually ended up looking up how to get to `/current/working/directory/path_a` from `/current/working/directory/path_b`.

Luckily, none of these behaviors can ever actually come into effect (the function is always called without a leading slash and without paths deliberately trying to escape), but the fix is still good to reduce surprise.

*The actual fix* is that we make the leading slash ignored (or quite the opposite - we always add it).
Now when either of the two arguments try to go up higher than the top level, they just end up at the top level (e.g. `foo/../..` is effectively `.`), only then the "relative" calculation happens.

---

This pull request makes the behavior 100% the same as in https://github.com/mkdocs/mkdocs/pull/2272, but without the disadvantage of being a "custom replacement for the standard lib function". It also brings 35% of the performance advantage.

Build times of a site that has ~350 pages:

* When running from directory `/home/oprypin/repos/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/a/athena-website`:
  * Before: 11.1 sec
  * After: 8.22 sec
* When running from directory `/home/oprypin/repos/athena-website`:
  * Before: 9.02 sec
  * After: 8.15 sec

(meanwhile, https://github.com/mkdocs/mkdocs/pull/2272 achieves 6.65 sec)

PR #2407, opened 2021-05-16
Optimize getting relative page URLs, now with less custom code

oprypin:
This is a custom implementation that's significantly faster but always gives the exact same results as the current one. The use of `posixpath.relpath` pulled in several path-specific transformations that are never needed here.

Efficiency is important because calls to `normalize_url` (e.g.) on a site with ~300 pages currently take up ~10% of the total run time due to the sheer number of them. The number of calls is at least the number of pages squared.

Generally, approximating that the number of these calls is N&times;N&times;2, this is the graph that we end up with:
image
[Full source code how I got this result](https://gist.github.com/b7d9b1e6bbecd4415dabb70aaba4c581)

This shows the total time spent getting relative paths (Y axis) over the course of building a site with that many pages (X axis).
With red color being "before" and green color being "after".
This does not show the total site build times, rather you can only subtract "red" from "green" to approximate the absolute savings of time.

In all other regards the sites' build times grow linearly, but only this particular place it grows as N^2 (because the templates end up linking to every page on every built page). So, with more and more pages, more and more percentage of time is spent just generating these relative URLs.
So I'm making it very optimal, because it's the place that really matters.

Previously: https://github.com/mkdocs/mkdocs/pull/2272, https://github.com/mkdocs/mkdocs/pull/2296 (a big number of tests was already added there)

### DIFF
--- a/mkdocs/utils/__init__.py
+++ b/mkdocs/utils/__init__.py
@@ -12,7 +12,6 @@ import shutil
 import re
 import yaml
 import fnmatch
-import pathlib
 import posixpath
 import functools
 import importlib_metadata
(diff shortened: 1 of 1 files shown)


oprypin:
I intend to merge this soon

PR #2421, opened 2021-05-24
No exception if directory already removed, closes #2420

### LINKED ISSUE #2420: file-system race condition while testing with mkdocs serve
when mkdocs serve is shutting down, it cleans up after itself removing the temporary directory created earlier.

the condition whether the directory needs to be removed is unchecked and prone to a file system race condition (shared resource).

given the directory is already removed on the file-system while shutting down, mkdocs serve throws an exception and exits in failure.

this error can be safely defined out of existence by removing the temporary directory only if it exists, will suggest a PR for this report. /E: created #2421

ktomk:
When `mkdocs serve` is shutting down, it cleans up after itself removing the temporary directory created earlier.

Previously the condition whether the directory is to be removed was unchecked and prone to a file system race condition (shared resource).

Given the directory is absent on the file-system while shutting down, `mkdocs serve` threw an exception and exited in failure.

Change is to override the cause of the exception by preventing the attempt to remove a directory that is already gone.

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -2,7 +2,7 @@ import logging
 import shutil
 import tempfile
 
-from os.path import isfile, join
+from os.path import isdir, isfile, join
 from mkdocs.commands.build import build
 from mkdocs.config import load_config
 from mkdocs.exceptions import Abort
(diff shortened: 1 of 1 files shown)


waylan:
I've never encountered this error. Can you demonstrate how to reproduce the error?

ktomk:
one way to provoke the stacktrace:, `mkdocs serve` and then (caution! - you get the idea) `rm -rf -- /tmp/mkdocs_*`, then press ctrl+c to shutdown server.

I only ran over this in testing while had a couple of servings and then cleaning up after sigkilling the one or other mkdocs serve - one did still run and then on clean shutdown I got the backtrace - the usual development ^^

waylan:
Ah, I see. This is one of those "protect the users from themselves" sort of things. Generally, my response is, 'well then, don't do that.'

In any event, there is no harm in adding this check.

PR #2426, opened 2021-05-25
restore re-building with symbolic-links, closes #2425

### LINKED ISSUE #2425: Re-building w/ symbolic links stopped working, regression after #2385

Preparing a PR with a patch.

ktomk:
previously (1.1.2 + master at 23e2051) building was triggered by changes
of file-content that was symbolically linked within docs_dir while
`mkdocs serve` was running.

since migrating from livereload>=2.6.1 to watchdog>=2.0.0 to detect
file-system changes (triggering the re-build) it stopped working.

this is because watchdog does not support symbolic links out of the box,
e.g. see [1].

change is to provide additional observe instructions on the realpath [2]
for the following cases:

1. docs_dir & config_file_path path deviation:

   when the absolute path to either the `docs_dir` or the `config_file` is
   different from its realpath, the realpath is added for observing (as
   well).

2. symbolic links within docs_dir:

   if a file within docs_dir is a symbolic link, the files real path
   is added for observing. sub-directories (that are not symbolically
   linked) are traversed up to a depth of nine levels (only if the
   recursive flag is enabled, otherwise no traversal into sub-directories).

Ref: 23e205153f01d24d50fe9ba18e5186cdbc2c2dbe
[1]: https://github.com/gorakhargosh/watchdog/issues/365
[2]: <https://docs.python.org/3.8/library/os.path.html#os.path.realpath>

### DIFF
--- a/mkdocs/livereload/__init__.py
+++ b/mkdocs/livereload/__init__.py
@@ -95,6 +95,26 @@ class LiveReloadServer(socketserver.ThreadingMixIn, wsgiref.simple_server.WSGISe
         handler.on_any_event = callback
         self.observer.schedule(handler, path, recursive=recursive)
 
+        # 1. path deviation of docs_dir, config_file_path
+        if not path == os.path.realpath(path):
+            self.observer.schedule(handler, os.path.realpath(path), recursive=recursive)
+
+        # 2. symbolic links of config_file_path and (in) docs_dir
+        def schedule(handler, path, recursive=True):
+            if schedule.depth >= schedule.limit:
+                return
+            schedule.depth += 1
+            if os.path.islink(path):
+                self.observer.schedule(handler, os.path.realpath(path), recursive=recursive)
+            elif recursive and os.path.isdir(path):
+                for name in os.listdir(path):
+                    schedule(handler, os.path.join(path, name), recursive=recursive)
+            schedule.depth -= 1
+
+        schedule.limit = 9
+        schedule.depth = 1
+        schedule(handler, os.path.realpath(path), recursive=recursive)
+
     def serve(self):
         self.observer.start()
 


waylan:
@ktomk there is no need to open an issue if you are going to open a PR. You can report the issue as part of the PR. To be clear, there is no rule one way or the other. Just trying to save extra work if it is not needed.

@oprypin as this relates to your code, I will give you a chance to review it.

ktomk:
good to know for next time @waylan

yes, please review and bear with me, my python skills are still young @oprypin

oprypin:
Thanks for the find. I wasn't considering symlinks at all.

From a quick analysis of the code flow (and I could be mistaken), it seems that the recursion in this code applies to just walking subdirectories, but multiple levels of symlinks don't actually seem to be followed. Is that intentional?

And, this will certainly benefit from tests. There are existing tests that won't be difficult to adapt -- just need to add creation of symlinks at the beginning. And, as per the previous paragraph, a test with a symlink to a directory with a symlink will definitely be needed, regardless if following multiple levels of symlinks is actually done.

ktomk:
> From a quick analysis of the code flow (and I could be mistaken), it seems that the recursion in this code applies to just walking subdirectories, but multiple levels of symlinks don't actually seem to be followed. Is that intentional?

First of all yes, it keeps the logic simple and I don't know if this worked earlier as this is not part of the setup I noticed it in and then checked. Maybe defer unless needed? @oprypin

mkdocs/livereload/__init__.py
```diff
@@ -86,6 +86,21 @@ def callback(event):
         handler.on_any_event = callback
         self.observer.schedule(handler, path, recursive=recursive)
```
waylan, line 87: Shouldn't this line be removed as we are now doing this down below?

oprypin:
Symlinks to symlinks would've worked before. And allowing only 1 level of symlinks just seems arbitrary.
That's about all I have for arguments. I never use symlinks with MkDocs, so I don't know.

In any case, this needs unittests.

mkdocs/livereload/__init__.py
```diff
@@ -86,6 +86,21 @@ def callback(event):
...
+            schedule.depth += 1
+            if os.path.islink(path):
+                self.observer.schedule(handler, os.path.realpath(path), recursive=recursive)
+            elif recursive and os.path.isdir(path):
+                for name in os.listdir(path):
+                    schedule(handler, os.path.join(path, name), recursive=recursive)
+            schedule.depth -= 1
+
+        schedule.limit = 9
+        schedule.depth = 1
```
oprypin, line 115: **Untested code**, and just for your reference, because I don't think this should be the final version.

A simpler way to write the same:

```suggestion
        def schedule(handler, path, limit=9):
            if os.path.islink(path):
                self.observer.schedule(handler, os.path.realpath(path), recursive=recursive)
            elif recursive and os.path.isdir(path) and limit > 0:
                for name in os.listdir(path):
                    schedule(handler, os.path.join(path, name), limit - 1)

```

mkdocs/livereload/__init__.py
```diff
@@ -86,6 +86,21 @@ def callback(event):
...
+        schedule(handler, os.path.realpath(path), recursive=recursive)
```
oprypin, line 116: Again untested code. This is the real simplification applicable only if we *actually* want to ignore nested symlinks.

```suggestion
        if recursive:
            for root, dirs, _ in os.walk(path):
                for relpath in dirs:
                    subpath = os.path.join(root, relpath)
                    if os.path.islink(subpath):
                        self.observer.schedule(handler, subpath, recursive=recursive)
```

Two additional comments on this front:

* I think traversing anything at all shouldn't be done with `recursive=False`. So I make it as such here.
* What was the reason to limit the depth of directory trees to 9? There is no threat of infinite recursion.

oprypin: On my last point:
Ah no, I understand how traversal works in your code. All is correct there (or at least is more likely to make sense), as traverses 1 file symlink at the root. My code fails to do that.

mkdocs/livereload/__init__.py
ktomk: no (no to remove, no its not doing the same below). this is the initial schedule and the main operation.

good question thought, after adding schedule I removed the following lines directly following, so it got lost in transpositions and your question made me aware that change 1. got completely under the wheels.

~~~
if not path == os.path.realpath(path):
    self.observer.schedule(handler, os.path.realpath(path), recursive=recursive)
~~~

if my concentration is still well enough this should be edited in.

mkdocs/livereload/__init__.py
ktomk: Yes I tried that as well (in a similar variant passing the limit / depth counter as argument), however it changes the signature compared to the parenting schedule implementation.

functionality wise it works, can confirm that as I tried it out already earlier.

mkdocs/livereload/__init__.py
ktomk: > * I think traversing anything at all shouldn't be done with recursive=False. So I make it as such here.

This might have been flawed, too tired to check, just take care that "traversing" (starting at) the first level must not mean it is recursive.

What I specifically don't know is the effect of recursive in watchdogs observe for a path that is a directory.

The implementation was under my assumption that the non-recursive observing of a directory path results in the observation of all files in that directory plus the addition and removal of files there-in.

You already added an additional remark yourself, I'm still interested in case you want to share your understanding of how the effect of recursive in watchdogs observe comes into play.

> * What was the reason to limit the depth of directory trees to 9? There is no threat of infinite recursion.

About why there is a limit at all: My very first approach had no function at all, so there was no recursion. Then I added the function as I realized that w/ recursion it is very straight forward for directory tree iteration (mind that I'm a python newbie). Then I reviewed the implementation and I realized that there were no limits. So I decided to add one, as even the code could endlessly run, any system it runs on has limits.

About why 9: I just needed a number. Three was too low. Six seemed cheap, seven nice and then gave two more. One for the part that the with any such limit it might be one less on how to start the traversal and one for the goddesses or guests.

Yes, the limit is synthetic, more research and development is likely to its benefit.

Reason: Filesystem I/O. Mitigation of the risk of tree traversal circles = endless loop/recursion. You're right when you look at the code, it does not look prone to it. But the code only shows one half of the story. While it runs, the file-system beneath is not frozen. That is the other half and there you have these race-conditions and what not.

The limit of nine might be wrong by the number, point in case is to have a limit. If its too strict, the number can be raised. That in short perhaps summarizes how I thought about it when/after adding the limiting.

This is missing a bit in the commit message, I shortened it while preparing the patch.

ktomk: > Symlinks to symlinks would've worked before.

I wasn't aware of this before, so from how I understand what a regression is, this just extended the scope. I'm glad this could be figured out that quickly.

> And allowing only 1 level of symlinks just seems arbitrary.

Well, there was one point I didn't explicitly mention before, but when following symlinks more level deeps while traversing a directory structure there is the problem of endless recursion:

~~~
* docs_dir
    * blue
        * red
          * yellow (symbolic link to ../../../docs_dir)  
~~~

this would then require bookkeeping to detect such a circular reference. this is above my paygrade in python.


> In any case, this needs unittests.

Big fan of those, but maybe before adding them, what your insightful comments on the realpath traversal depth and also on the general docs_dir traversal depth have shown to me, I would prefer a regression test-case. Honestly, I don't know what the processes are in this project, so see it as a suggestion on how to proceed and if you like let me know what you think about it.

mkdocs/livereload/__init__.py
oprypin: > however it changes the signature

It does indeed change the signature of the function (the args it accepts, if that's what you're referring to), but that doesn't matter because the function is fully internal (impossible for a user to call).

---

And a comment that I add since you mention wanting to improve Python knowledge.

`some_function.some_var = 2` is usually seen as "obscure". The main argument against using this ability is that most readers would be surprised that it exists, or generally not expect it, meaning it might take away from their process of understanding the code.

The direct alternative here is plain variables but with the `nonlocal` keyword. But in this case I still think that's moot because I think my suggestion removes that need entirely.

mkdocs/livereload/__init__.py
oprypin: > The implementation was under my assumption that the non-recursive observing of a directory path results in the observation of all files in that directory plus the addition and removal of files there-in.

I think that's right.

---

Regarding loops in file systems.

If we support symbolic links, yes, they are definitely possible and need to be detected.
And then also a recursion-based solution is required; `os.walk` is not usable.

If we don't support, you are also correct that exceptional cases can still cause loops. But I think in that case it is very typical to ignore such a possibility (and certainly way more typical than limiting traversal at N nested directories). I mean, if this was a ubiquitous concern, `os.walk` would basically never be usable. Instead, I'd say it's considered fair game for non-specialized tooling to enter infinite recurstion when encountering a hard loop in the file system; and usually people who have such loops are also aware of the complications. But this is just coming from my observations, I don't have hard evidence.

And if we talk about `livereload` library, well, it just follows all symlinks and all dirs with absolutely no attempt to avoid loops or putting other limits on traversal.
https://github.com/lepture/python-livereload/blob/ecabebaeee2a40743f5a8216e1856c9f9150913f/livereload/watcher.py#L197.

---

Regarding how to actually detect loops in symlinks:
I think this is better done by remembering seen directories so far. Then depth limiting is not needed.
I see you are aware of this idea already, but sure, maybe it needs more specific Python knowledge.

> this is above my paygrade in python

Mhm well I would be glad to take over implementing this. Just didn't want to rush in and be like "no, let *me* do it".
Please let me know what you think.
And I would indeed do it with the approach that doesn't require limiting traversal depth, and I'd make it traverse nested symlinks.

---

> > In any case, this needs unittests.
>
> Big fan of those, but maybe before adding them

Ah I think there's no "before" that applies here. This case to me seems like a poster child for writing tests first (or at least right after you realize that the code might go through several iterations before reaching the final one). You do the setup with whatever directory structures are relevant, define what should be done in each scenario and test accordingly. Then you have a tool to always instantly tell you: "is the code I just wrote correct?". Moreover, it even gives you the ability to answer review comments with something like "no, I can't remove that line, see, then this test fails".

Of course, I realize that we haven't yet decided what *should* indeed be done in each scenario. But for our particular case that's not a significant obstacle, because it's super cheap to adapt the tests if needed (we expect a rebuild here <-> we expect no rebuild here).

To me, tests here are almost more valuable than the solution itself. Otherwise we can't even be sure that the issue is fixed.

ktomk:
> Thanks for the find. I wasn't considering symlinks at all.

If watchdog would just support them out of the box, this would be a no-brainer. You linked livereload impl. earlier, perhaps this should just be ported in-place (for resolution of symlinks only). Was not aware of it, likely the sweet spot.

> Symlinks to symlinks would've worked before.

This then taken care of, too.

> And if we talk about livereload library, well, it just follows all symlinks and all dirs with absolutely no attempt to avoid loops or putting other limits on traversal.
https://github.com/lepture/python-livereload/blob/ecabebaeee2a40743f5a8216e1856c9f9150913f/livereload/watcher.py#L197.

Exactly that. (I just didn't see recursion there but only looked two times, so could be me)

> Ah I think there's no "before" that applies here.

Maybe a miscommunication. Before is at 23e2051 and it exists. Regression test must pass there and fail with feature merge, then must pass for the fix. Give the test a date to be removed in fix-time + one year. Just my 2 cents.

> Regarding loops in file systems.

(also race conditions and that material) Whatever works for you, make the decisions you consider wise. There is something to learn every day in that area and I trust you to handle this well for mkdocs.

> Mhm well I would be glad to take over implementing this. Just didn't want to rush in and be like "no, let me do it".
Please let me know what you think.

Well, do what suites you. I marked the PR as draft and allowed edits, there is nothing that should have prevented you from forking earlier and thanks for being verbal.

So just take the report / PR as inspiration and improve the original feature. Totally fine with that.

I only updated to match with the commit message as change 1. went bonkers, this is edited in.

And you can always ping me for feedback, I might get my mkdocs build pushed which triggered this, so that there is also the use-case available and visible. If I do, I'll comment here with a link.

Today were too many things to do, so I could only summarize here quick and I hope I have addressed all the parts you wanted to have feedback for.

waylan:
I haven't read through everything here, so maybe this is all covered. Unfortunately, I have very limited time this week (more than usual) and will not be able to give any time to this until next week at the earliest.

So I will just briefly summarize what I expect the behavior should be. Essentially our behavior should exactly match livereload's in this case. Based on reports, some of our users rely heavily on symbolic links and this would be an unacceptable regression for them. Those users would likely argue that we either need to fix this or revert the changes we have now. 

I see three possible solutions:

1. Fix this by modifying Watchdog's behavior locally so that it matches  liverelaod's (no limits at all).
2. Replace watchdog with a competing library which supports this.
3. Revert back to the new server.

My hope is that we can find a reasonable solution for option 1. I would prefer to not need to do option 3. However, if I had recognized this was an issue before, I would have considered this a blocker to accepting that PR. Personally, I don't use symbolic links in my MkDocs projects, but we have always guaranteed they will work, so we can't be breaking them. And we shouldn't be creating any artificial limits either.

oprypin:
I'm on it

PR #2427, opened 2021-05-28
Restore the functionality of watching files through symlinks

### LINKED ISSUE #2425: Re-building w/ symbolic links stopped working, regression after #2385

oprypin:
### Fix the ability to watch single files

Watchdog doesn't support it directly but it happened to work everywhere except Windows

---

### Restore the functionality of watching files through symlinks

(lost in commit a444c43)

This is achieved by walking through the target directory, finding any symlinks in it, and watching their target too (+recursive walking of those too).

Important note: only symlinks that exist during startup will be taken into account, new ones aren't added.

Fixes #2425
Closes #2426
cc @ktomk

### DIFF
--- a/mkdocs/livereload/__init__.py
+++ b/mkdocs/livereload/__init__.py
@@ -4,6 +4,7 @@ import logging
 import mimetypes
 import os
 import os.path
+import pathlib
 import re
 import socketserver
 import threading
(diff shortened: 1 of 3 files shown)


waylan:
As the failing tests are all related to the fix in #2430, they can be ignored here.

ktomk:
Alright, preparing to run tests on my end again.

PR #2438, opened 2021-06-02
Restore the ability to watch files through relative symlinks

### LINKED ISSUE #2425: Re-building w/ symbolic links stopped working, regression after #2385

oprypin:
(i.e. symlinks that have a non-absolute path as the target)

This was not fully fixed in #2427.

Fixes #2425

The problem is with the workaround I added for Windows + Python < 3.8, where `realpath` is a no-op.
https://bugs.python.org/issue9949#msg350109
Now I remove the workaround and do a different approach that, surprisingly, works even on older Python.

Of course, I am adding a test here that failed before but will now pass.


ktomk:
Your efforts deserve a medal. Personally I would not give Windows Symlink support much of chance. They have neglected them for too long, not to say relative ones, and now make it over complicated just to continue that history. That ship has sailed. Rely on the Python interface, and if it does not support it, don't play whack-a-mole.

No need to solve problems that weren't any for mkdocs users in the past. Might be worth to take notes, if there are upcoming reports thought. My 2 cents.

PR #2439, opened 2021-06-02
Allow CI to run when pushing to a branch

oprypin:
The motivation is that contributors currently have no good way to test their changes before opening a pull request.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,14 +1,6 @@
 name: CI
 
-on:
-  push:
-    branches:
-      - master
-    tags:
-      - '**'
-  pull_request:
-    branches:
-      - '**'
+on: [push, pull_request]
 
 jobs:
   test:


oprypin:
When trying some change to MkDocs, I have no good way to run CI / tests with GitHub Actions without opening a pull request. Which maybe I don't want to open if it turns out that my change has big problems on other platforms.

So, what I usually do is push a branch to my repo that has both the relevant change and this edit to the workflow file just so it doesn't block my push from being tested. Which I later need to undo before opening a PR.
Other times I may not bother to do that and end up having to scramble for a fix in an already-open PR.

I've also been watching @waylan for a while - opening a pull request, not waiting for any comments, then merging it. The first time I saw that I perceived it as rude. But after thinking for some time, I realize, could it be that the biggest reason for doing this was just to be able to run GitHub Actions CI?

The thing is, it doesn't have to be this way. This configuration change can just be applied permanently with no real downsides.
Well, there is actually a slight downside - it's that a commit will be tested twice: first as a push in the contributor's repo, then as a pull request. Which is obviously wasteful and some people find it annoying, but it's not our fault that GitHub does not provide a good solution to this.
Anyway, all my repositories and also all projects where I'm a major contributor do it that way too. [One example.](https://github.com/crystal-lang/crystal/blob/fc8ae1e93bc88f63339cf2967f2d69dce4166548/.github/workflows/win.yml#L3)

oprypin:
Further notes why I remove other parts.

Note that [as per the docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags), the list `on.<push|pull_request>.<branches|tags>` is exhaustive.

```yaml
on:
  push:
    branches:  
      - master
```
As I want to change this to all branches, this becomes '**', making this whole specification mean "on any push", and not specifying anything is also "on any push"
```yaml
    tags:
      - '**'
  pull_request:
    branches:
      - '**'
```
As pull requests only apply to branches, this specification just means "on any pull request", and not specifying anything is also "on any pull request"

waylan:
The reason for the current setup is that as a policy (which I don't think is fully documented) we do not want any other branches created on this repo. Branches should only be created in the contributor's fork. Additionally, no commits should ever be pushed directly to the master branch. All changes must be run through CI first. Therefore, the current config forces all contributors (including those with full privileges) to always make changes on branches on their own forks and then use those to create PRs. So yeah, sometimes I may make a change which I do not see any need to get feedback on. But I still need to create a PR to ensure it runs through CI before committing. However, I won't wait for comments and merge it once CI is done. Other times I will wait for comments. If depends on what the changes are.

Personally, I have no interest in changing that policy. Of course, we could keep that policy and still make this change. However, with this change we are not "forced" to follow the policy, which I'm not crazy about.

Regardless, I'm not sure I understand the purpose of this change. Apparently, you want to be able to run CI on branches on your own fork? Why not just create a PR within your own fork against master of your fork? Personally, I don't want CI running on my own fork every time I push a change. CI should only run when I create a PR. And as a PR will always be against `master` (either of my fork or of this repo), then this change is unnecessary.

oprypin:
Regarding most of the first paragraph: I don't think any of this is relevant. Stopping people from being able to run CI doesn't stop them from pushing to master.

> But I still need to create a PR to ensure it runs through CI before committing.

Again not really related.
As I'm trying to explain, creating a PR is not a requirement for running through CI.
Moreover, creating a PR doesn't actually ensure anything, you could just merge it immediately.

> Apparently, you want to be able to run CI on branches on your own fork

Yeah for sure!
Would be really nice to not need to create these commits over and over.
https://github.com/oprypin/mkdocs/commits/90ea7f5afad5097e4fb34e18e6c01e95b232a9a4
https://github.com/oprypin/mkdocs/commits/ff1df7611f9e48cda6ad9f4d84bcba66bcdcd367
https://github.com/oprypin/mkdocs/commits/5987a59008d6081d75c74231114a087ca417f47c
https://github.com/oprypin/mkdocs/commits/777e7493dc3908c022acc7af6fa6708803dad97e

> Why not just create a PR within your own fork against master of your fork

~~No, excuse me. Why not just *not* completely fill my thread history with garbage?~~

Having garbage PRs in my history is a big reason why not.
I'd rather create a commit than create a PR, anyway.

waylan:
> > Why not just create a PR within your own fork against master of your fork
> 
> ~No, excuse me. Why not just _not_ completely fill my thread history with garbage?~
> 
> Having garbage PRs in my history is a big reason why not.
> I'd rather create a commit than create a PR, anyway.

Whats the difference? Personally, I never even look at the PR history in my own personal fork. If I needed to run CI privately, creating a PR on my personal fork is the best place to do it. It will never bother me as I never look there.

What I don't want is for CI to run *every single time I make a commit* which will fill my CI history with garbage. It should only run when I explicitly want it to run. And that is already possible by limiting runs to PRs.

PR #2440, opened 2021-06-04
Bump version to 1.2

waylan:
I believe we are ready to release version 1.2. Thanks to everyone who contributed. There is a lot of long-desired stuff in this release.


waylan:
@Ultrabug the `translation` test is failing here and I don't think it should be. Any insight you can provide?

oprypin:
It's almost as if... [running checks only on release is a bad idea](https://github.com/mkdocs/mkdocs/pull/2299#issuecomment-804362611)

ultrabug:
@waylan yep, never had the chance to look it up yet sorry. I think this is related to the actions "plugin" that I use to iterate over file changes.

PR #2477, opened 2021-06-27
Revert livereload file watching to use polling observer

### LINKED ISSUE #2457: mkdocs 1.2.x livereload no longer working
mkdocs==1.2 and latest no longer works
mkdocs==1.1.2 works

MacOS 11.4
Docker version 20.10.7, build f0df350
python:3.8-alpine (3.9.5 and 10-rc same behavior)

I checked the Docker container and the markdown is updating but the generated html is not. This seems to be associated with the 1.2 change with watcher.

LMK what logs/config/diag I can provide to help RCA the issue.

oprypin:
This goes back to the approach that was always used with 'livereload' library (but now just using the 'watchdog' implementation of the same), meaning the same downsides with latency and CPU usage.
But we have to do this to reasonably support usages that span virtual filesystems such as non-native Docker and network mounts.

This also simplifies the code, as the polling observer already follows symlinks and happens to support watching paths of files directly

Fixes #2457

### DIFF
--- a/mkdocs/livereload/__init__.py
+++ b/mkdocs/livereload/__init__.py
@@ -4,7 +4,6 @@ import logging
 import mimetypes
 import os
 import os.path
-import pathlib
 import re
 import socketserver
 import threading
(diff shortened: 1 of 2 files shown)


oprypin:
I intend to merge this in a day

PR #2478, opened 2021-06-27
Allow mkdocs.yaml when '--config' is not passed + refactor

### LINKED ISSUE #2213: Allow mkdocs.yaml config file (.yaml extension)
Even though `.yml` is commonly used, `.yaml` is actually preferred:

https://yaml.org/faq.html

> Is there an official extension for YAML files?
>
> Please use ".yaml" when possible.

It would therefore be great if you would allow `mkdocs.yaml`.

Somewhat related t #2164

oprypin:
Fixes #2213
Supercedes / closes #2206﻿ by @hwmrocker


hwmrocker:
thank you :)

PR #2490, opened 2021-07-08
Revert "Ensure site_url and use_directory_urls do not conflict."

### LINKED ISSUE #2189: Use use_directory_urls: false breaks on empty site_url

### LINKED ISSUE #2479: Create site for usage in a Docker image that can be deployed to any domain/URL
We use mkdocs to create a static web site that is deployed via Docker images.
We build the static HTML files with mkdocs, then put them in a Docker image that includes an Apache web server.
And then we (our customer actually) deploy the Docker image 8 times, each to a different domain, even with different sub URLs (like e.g. https://domain1/path/one and https://domain2/longer/different/path/two)

With mkdocs 1.2, mkdocs complains that site_url is now mandatory.

How can we solve our use case with mkdocs 1.2?

Is setting site_url to an empty string the recommended way? I think I've read in some tickets ("somewhere"), that setting site_url to an empty string has some negative side effects.

I personally don't think that making site_url mandatory is the right way. In today's containerized world, you often don't know the final domain/URL that your container will be deployed to. And rebuilding a static site for each container deployment is not something that can be done by the image creators.

oprypin:
This reverts commit b89ec5792ed35aacf2da00c34d4d62ba45fe08c9 (https://github.com/mkdocs/mkdocs/pull/2405)

That commit claims to fix some bugs in interactions with `use_directory_urls`, but in fact there are currently no known unfixed bugs.
So there is no need to have that breaking change.

See https://github.com/mkdocs/mkdocs/issues/2189#issuecomment-876733593

Closes #2189
Closes #2479


oprypin:
I intend to merge this in 1 day

ultrabug (approved):

PR #2501, opened 2021-07-13
CI: Run all lint checks in one job

oprypin:
This is more considerate of the computational resources. The lint checks are not the bottleneck, and now they won't starve other jobs from starting.
All checks will still run and a failure in any of them will fail the job.

Also upgrade Python and Node versions

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -55,36 +55,35 @@ jobs:
           fail_ci_if_error: false
 
   lint:
-    strategy:
-      fail-fast: false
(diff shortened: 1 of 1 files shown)


oprypin:
Timing example:

Before | After
-|-
image | image

As you can see, the total run time of the "lint" job is now smaller than the previous sum. The duration is still much smaller than the average of normal test jobs

oprypin:
I intend to merge this in a day

ultrabug (approved):

PR #2503, opened 2021-07-13
Allow using just an absolute path as the `site_url`

oprypin:
Then it will be used only for `mkdocs serve` integration.

Example:

```yaml
site_url: http://example.org/foo
site_url: /foo/
```

The two above configs would behave exactly the same for `mkdocs serve` root offset. (The 2nd one is the new ability)
But only the first one would also set a canonical URL on each page.

References https://github.com/mkdocs/mkdocs/issues/2189#issuecomment-847816816, https://github.com/mkdocs/mkdocs/pull/2424


squidfunk:
Does the 404 page still work with non-absolute URLs? Asset URLs are all relative.

oprypin:
@squidfunk There is no change in behavior for all situations that were already valid. And no change to how URLs work in general. Just that the `site_url` allows specifying a path rather than a full URL. This also doesn't touch anything regarding relative URLs: if you look at the example above, both URLs are "absolute" just that one of them is without a domain.

oprypin:
Oops, I was messing around with dependent pull requests but they're still very janky.

Anyway, I'm having second thoughts on this. It's a low benefit change that has the potential to be costly long term

ultrabug:
> It's a low benefit change that has the potential to be costly long term

Can you elaborate on this?

oprypin:
@ultrabug 

Low benefit: who cares about a config that just does an inconsequential change to `mkdocs serve`'s behavior.

Potentially costly: maybe this will turn out to be prone to bugs and difficult to maintain. Or maybe we will want to assign different functionality to just specifying a path but that wouldn't be backwards compatible anymore.

ultrabug:
> Or maybe we will want to assign different functionality to just specifying a path but that wouldn't be backwards compatible anymore.

Thanks for clarifying, indeed this can maybe wait for the day we'd want or be asked for that

Tejes:
I was surprised by the "Error: The URL isn't valid, it should include the http:// (scheme)" message. We are making the documentation of our system with mkdocs. We build the site in a docker container, then use the results to build an image with nginx serving the result. We push this image to our registry and then deploy them to various kubernetes environments. Knowing the site_url in advance is not feasible, as it could be `http://localhost:8084/docs` for a developer or `https://app-dev.example.com/docs` for a staging environment, etc. So we do not specify `site_url` (I know, madness), and everything works fine, except for the 404 page. I thought setting `/docs/` as `site_url` would solve problem with minimal restrictions (all our environments have the site available there), but it sadly isn't allowed.

PR #2506, opened 2021-07-16
Tighten the docs and tests around `site_url`

oprypin:
Some of the docs were accidentally reverted with other things, so I'm restoring them and adding a bit more.

Fixing an edge case: it is in fact possible to specify a URL with scheme and without netloc. Adding new tests to also catch that.

### DIFF
--- a/mkdocs/tests/config/config_options_tests.py
+++ b/mkdocs/tests/config/config_options_tests.py
@@ -280,38 +280,33 @@ class IpAddressTest(unittest.TestCase):
 class URLTest(unittest.TestCase):
 
     def test_valid_url(self):
-
-        url = "https://mkdocs.org"
-
         option = config_options.URL()
-        value = option.validate(url)
-        self.assertEqual(value, url)
+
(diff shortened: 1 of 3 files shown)


mkdocs/tests/config/config_options_tests.py
```diff
@@ -280,39 +280,34 @@ def test_invalid_IPv6_address(self):
...
                           option.validate, 1)
 
-    def test_url_is_dir(self):
-        url = "https://mkdocs.org/"
-
-        option = config_options.URL(is_dir=True)
-        value = option.validate(url)
-        self.assertEqual(value, url)
-
-    def test_url_transform_to_dir(self):
```
ultrabug, line 309: It looks like what this test was covering is not checked anymore if you remove that test or I might be missing something?

oprypin: The line `self.assertEqual(option.validate("https://mkdocs.org"), "https://mkdocs.org/")` covers exactly this

ultrabug (approved):

PR #2507, opened 2021-07-17
Don't require trailing slash when serving a directory index

### LINKED ISSUE #2456: V.1.2: Markdown links earlier processed as html files, are now processed as directories
The new version brings a curious issue. With versions < 1.2, I used internal links in this way, to make a call to page `bar.md`, from page `foo.md`:

```markdown
[linking to content in bar](../bar#content)
```

Which was transformed correctly into:

```html
<a href="../bar/#content">linking to content in bar</a>
```

Since version 1.2, the `bar` is intepreted differently by `mkdocs serve`, which breaks all such links.  I realize this is highly irregular and correct call should have been:

```markdown
[linking to content in bar](bar.md#content)
```

So please dont throw rotten tomatoes at me 🍅 .

But just to help me fix this: why this change of behavior? Has some part been reimplemented? Or has this been intentional, to prevent that kind of error in the future?

oprypin:
Don't require trailing slash when serving a directory index in the livereload/local server.

Instead of showing a 404 error, detect if it's a directory and redirect to a path with a trailing slash added.

Fixes #2456

### DIFF
--- a/mkdocs/livereload/__init__.py
+++ b/mkdocs/livereload/__init__.py
@@ -183,10 +183,10 @@ class LiveReloadServer(socketserver.ThreadingMixIn, wsgiref.simple_server.WSGISe
         if path == "/js/livereload.js":
             file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "livereload.js")
         elif path.startswith(self.mount_path):
+            rel_file_path = path[len(self.mount_path):].lstrip("/")
             if path.endswith("/"):
-                path += "index.html"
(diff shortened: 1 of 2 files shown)


oprypin:
In https://github.com/mkdocs/mkdocs/issues/2456#issuecomment-881876681 I see that people can perceive this as a blocker even regardless of the fact that it used to work before. So I decided to finally tackle it.

@ultrabug Any objections? Otherwise, merging this tomorrow

ultrabug:
Sorry too slow to respond but I agree it's a good move

PR #2515, opened 2021-07-21
Fix minor errors in documentation

mondeja:
Fix whitespace added to codespan and ISO link not consistent across entire docs. Also add links that I've missed reading these parts.


ultrabug:
Thanks mate

PR #2529, opened 2021-08-05
Ignore hidden files during live reload

### LINKED ISSUE #2519: Live Reload / DIrty Reload and hidden or temporary files
When using `mkdocs serve --livereload` or `--dirtyreload`, `mkdocs` detects changes to editor temp files like `.foo.md.swp` or hidden files like `.foo` and constantly regenerates the documentation.  I tried adding the `mkdocs-exclude` plugin to ignore the glob `.*.swp`, but this does not stop the live reload from executing.

It seems like `mkdocs` shouldn't be scanning dot files anyway since these files aren't included in the output.  This would produce less noise on the console and fewer browser page re-loads when editing the Markdown files with `vi` and using the live reload / dirty reload functions.

(although it would be nice to be able to include the `.well-known/` directory or have a way to specify which dot files to be included, which I guess is still waiting for a design decision in #2139 )

osresearch:
This patch resolves #2519 by changing the callback used by `--livereload` and `--dirtyreload` to ignore hidden files such as editor temporary swap files.  They are not included in the built site anyway, so they should not trigger a documentation rebuild.

### DIFF
--- a/mkdocs/livereload/__init__.py
+++ b/mkdocs/livereload/__init__.py
@@ -79,6 +79,9 @@ class LiveReloadServer(socketserver.ThreadingMixIn, wsgiref.simple_server.WSGISe
         def callback(event):
             if event.is_directory:
                 return
+            filename = os.path.basename(event.src_path)
+            if filename[0] == '.':
+                return
             log.debug(str(event))
             with self._rebuild_cond:
                 self._to_rebuild[func] = True


ultrabug:
Thank you for your first contribution to this project @osresearch :)

Running the CI, @oprypin I think it looks sane but you're the master of the liveserver

oprypin:
Hm actually this might be invalid, because, for example, *mkdocs-awesome-pages-plugin* does look at such files, and then users would be confused why it doesn't autoreload

pawamoy:
image

@oprypin, @lukasgeiter, care to elaborate on the impact it would have on [mkdocs-awesome-pages-plugin](https://github.com/lukasgeiter/mkdocs-awesome-pages-plugin) or other plugins, so that we decide what to do this PR?

lukasgeiter:
In case of awesome-pages, the default file for configuring the navigation is called `.pages`. If all dotfiles would be ignored, users wouldn't see the effect of changes in `.pages` until they restart the server or trigger a rebuild by editing a file that is not excluded.

PR #2545, opened 2021-08-19
Add missing es translations to readthedocs theme

mondeja:
I'm not sure why these weren't added in #2396, maybe I forget to run `update_catalog` after `init_catalog`. This is the cause of the [failure ocurred in latest release workflow](https://github.com/mkdocs/mkdocs/runs/3098566096).


ultrabug:
Thanks @mondeja

PR #2550, opened 2021-08-20
Add __repr__ method to File class

mondeja:
I've found myself when developing plugins printing `File.src_path`, a lot of times, so I've created the reproduction method based on the same defined in `Page` class.

### DIFF
--- a/mkdocs/structure/files.py
+++ b/mkdocs/structure/files.py
@@ -141,6 +141,9 @@ class File:
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    def __repr__(self):
+        return "File(src_path='{}', url='{}')".format(self.src_path, self.url)
+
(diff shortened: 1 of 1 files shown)


ultrabug:
Do we want to keep using `.format` for this kind of stuff instead of going straight to f-strings?

I doubt it's being used in logs so I'd rather go for a more modern standard maybe?

mondeja:
I agree but this should be done for all formatted strings, right?. Maybe using [pyupgrade f-strings](https://github.com/asottile/pyupgrade#f-strings) feature? We want to do other upgrades as well and consider adding pyupgrade as part of CI?

mondeja:
> Do we want to keep using `.format` for this kind of stuff instead of going straight to f-strings?

Opened #2612 to replace most of them.

PR #2535, opened 2021-08-06
Added translation for Brazilian Portuguese

piantino:
Hi,

This commit is enough?

Best regards,

PR #2497, opened 2021-07-13
zh_CN: Complete the translation of the mkdocs theme

xingkong0113:
Complete translation both
        mkdocs
        and
        readthedocs
        themes.
        
        Signed-off-by: xingkong0113 <26911141+xingkong0113@users.noreply.github.com>

PR #2564, opened 2021-09-07
tox translation build: add tests for lunr.js language support

ultrabug:
this should avoid forgetting that a locale added to mkdocs theme
translations should be supported by lunr.js to work

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -99,6 +99,12 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install -r requirements/project.txt
+      - name: Check if the locales are supported by lunr.js
+        run: |
+          for locale in $(ls mkdocs/themes/mkdocs/locales/); do
+            python -m pip show -f mkdocs | grep -q "lunr.${locale}.js" || exit 1;
+          done
+        shell: bash
       - name: Check if Portable Object Templates should have been updated
         run: |
           python setup.py extract_messages -t mkdocs


ultrabug:
@oprypin I intend to revert https://github.com/mkdocs/mkdocs/pull/2497 and merge this in 2 days

ultrabug:
Unless we want to disable the search plugin for missing locales with a WARNING log instead of crashing like mad?

oprypin:
@ultrabug Yes, I definitely don't think search plugin support should affect availability of languages. I was actually thinking perhaps a fallback language would be used in that case. Or whatever mkdocs-material does.

ultrabug:
mkdocs-material does crash if you select a language it does not support

```
  File "/home/alexys/venv/ultrabug.fr/lib/python3.9/site-packages/material/partials/language.html", line 4, in top-level template code
    {% import "partials/languages/" ~ config.theme.language ~ ".html" as lang %}
  File "/home/alexys/venv/ultrabug.fr/lib/python3.9/site-packages/jinja2/loaders.py", line 197, in get_source
    raise TemplateNotFound(template)
jinja2.exceptions.TemplateNotFound: partials/languages/pp.html
```

ultrabug:
With latest commit, the search plugin will only WARN on unsupported lang configuration:

```
WARNING  -  Ignoring search.lang "ezn" as it is not a supported language code.
WARNING  -  Ignoring search.lang "coco" as it is not a supported language code.
```

Removing the offending locales from the search.lang config is fine since this config option allows to be empty and falls back to english anyway.

I've made the tox CI to WARN as well but not fail the CI when a lunr locale is missing.

If you're okay with this, I'll update the tests accordingly

oprypin:
@ultrabug Hmm sorry I'm not quite following the whole scope of this. I have given an example of the ideal state I imagine this in: https://github.com/mkdocs/mkdocs/pull/2535#issuecomment-915385322

Secondary concerns: surely we don't want a warning if MkDocs has a translation but the search plugin doesn't know the locale for it? In that situation I'd prefer choosing (centrally, within MkDocs) a fallback language (normally English)  the search plugin.

The reason I don't want it as an actual warning is that then people can't use --strict mode. If MkDocs has the right translation, I think the message should be INFO level.

PR #2565, opened 2021-09-08
fix translation installation for locales with script name

ultrabug:
when a locale like zh_CN is parsed by babel.core.Locale, the
result Locale is zh_Hans_CN which is then used to find out if
translations exist for this locale name

the Hans part is the guessed script name part of the babel Locale
but is almost never contributed as translation locale name so
we have to make sure what we look for the language + territory
locale name instead of the fully parsed name from babel

### DIFF
--- a/mkdocs/localization.py
+++ b/mkdocs/localization.py
@@ -56,9 +56,13 @@ def _get_merged_translations(theme_dirs, locales_dir, locale):
     merged_translations = None
 
     log.debug(f"Looking for translations for locale '{locale}'")
+    if locale.territory:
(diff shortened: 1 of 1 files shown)


ultrabug:
With this + #2564 we will be able to support zh_CN language

prcr:
Good one, I was thinking of proposing this functional approach on https://github.com/mkdocs/mkdocs/pull/2564 to solve the issue of the `pt_BR` translation in https://github.com/mkdocs/mkdocs/pull/2535. :+1:

Issue #193, opened 2014-11-06
Support multiple versions of docs.


jimporter:
I'd like to be able to host different docs for different versions of my code. I know there are services like ReadTheDocs that handle this for you, but it would be nice to be able to deploy all the versions to GitHub instead. This would make the gh-deploy script a bit more complex (and there'd need to be some way of switching versions when browsing), but I think it would be generally useful.

Maybe there's an argument to be made that GitHub Pages are for hosting the in-development version of your docs, and places like ReadTheDocs are for released versions; I'm not sure. My instinct is that all the versions of the docs should be hosted in the same place, though.

Any thoughts about this?

d0ugal:
I've started making a tool that makes this possible. It is a very rough and early alpha, but it is a good start. I just need to add a version switcher ala ReadTheDocs or the django docs and it should work well.

It is somewhat opinionated, as it assumes you want to build all tags and that you are using git. This is why I am building it externally for now, this could be changed but I didn't want to get blocked from experimentation.


http://d0ugal.github.io/mkdocs-versioned/

tomchristie:
@d0ugal Sounds neat!

ePirat:
I agree with @d0ugal and think that the versioning of docs should be based on git tags as most docs are under git version control anyway. There are some important things to consider, though.
Given that older versions might have older configs, one possible approach would be that only the config of latest should be used to build all older docs (one corner case would be that older versions could require special markdown extensions that were disabled in newer configs, so this needs to be thought through a bit how it can be solved).
A version switcher like readthedocs has would be great, although this would make it mandatory to regenerate all old docs as well, when publishing new ones. So this would be another important thing to discuss. It would as well have the benefit that older docs can include a warning that these are for older versions and deprecated. (There are actually a lot of people reading old docs by accident)

To sum up the important points I would like to suggest for discussion:

- Use git tags for versioning?
- Only use current (latest) config?
- Have a version switcher in the docs/deprecation warnings?

iwootten:
Hey @dougal - that looks great, but I can't see any files listed in pypi? https://pypi.python.org/pypi/mkdocs-versioned/0.1.0.dev0 I get an error about 'No distributions at all found for mkdocs-versioned==0.1.0.dev0'. I attempted installing from your setup.py script in the repo (which seems to have worked), but can't see it on the commandline?

iwootten:
[...]

PR #2585, opened 2021-09-23
Upgrade readthedocs theme to v1.0.0

### LINKED ISSUE #1763: Update jquery version in readthedocs base.html
Would it be possible to update the JQuery version in the readthedocs base.html from the current 2.1.1 (observed in master) to 2.2.4 (the latest)?

Reason for request:
The jQuery.globalEval function in JQuery version 2.1.1 presents a cross-site scriping vulnerability that apparently has been fixed in the latest version of JQuery.

 mkdocs/mkdocs/themes/readthedocs/base.html

line 42:  <script src="{{ 'js/jquery-2.1.1.min.js'|url }}" defer></script>

### LINKED ISSUE #2467: update all jquery usage to remove quarantining
per #1763 the jquery versions used in *both* the readthedocs and the mkthdocs themes.  Locations:

  - https://github.com/mkdocs/mkdocs/blob/master/mkdocs/themes/mkdocs/js/jquery-1.10.2.min.js
  - https://github.com/mkdocs/mkdocs/blob/master/mkdocs/themes/readthedocs/js/jquery-2.1.1.min.js

This is triggering quarantines on mkdocs for any group using NVD (6.1 rating) or NexusIQ (9.8) rating.

I'll attempt to make a pull against this issue in the next days.

### LINKED ISSUE #2537: Missing logo in RTD theme
How would we go about adding the site/project logo top left in the RTD theme? I am a little confused where the logo actually comes from in projects on RTD that _do_ have one (such as [this](https://psyplot.readthedocs.io/en/latest/)).

https://www.mkdocs.org/user-guide/choosing-your-theme/#readthedocs -> 'include_homepage_in_sidebar' states

> Note that the site name/logo always links to the homepage

However, that appears to be the only place a logo is mentioned at all. At https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html#adding-a-logo (and some [SO questions](https://stackoverflow.com/q/59215996/131929)) it appears configuring logo is unrelated to RTD or MkDocs but is actually targeting Sphinx.

### LINKED ISSUE #2538: Update RTD theme from upstream
@waylan last updated the RTD theme from upstream 0.4.1 in 2018 with #1594. Upstream is now at [0.5.2](https://github.com/readthedocs/sphinx_rtd_theme/releases/tag/0.5.2) and I think it's time to bring  MkDocs' copy on-par.

While I _could_ just copy the version into a PR I was wondering if there will be dragons lurking with such an approach. So, maybe someone a bit more familiar with this project could provide insights as for how to properly test such a change.

Btw, what prompted me to look into this was the improved styling of the navigation captions.  
 
Screen Shot 2021-08-08 at 21 35 32

I could fix this in my site by adding a few lines to my `extra.css` but I thought updating the theme would be a bit more "sustainable".

```css
.wy-menu-vertical header,
.wy-menu-vertical p.caption {
  color: #55a5d9;
  font-size: 85%;
  margin: 12px 0 0;
}
```

Sparticuz:
I've gone through each release we've missed on the readthedocs theme, and updated the mkdocs fork. I've mainly kept to just implementing the changes as able and not rewrote functionality, (see Copyright in [mkdocs fork](https://github.com/Sparticuz/mkdocs/blob/master/mkdocs/themes/readthedocs/footer.html#L19) vs [upstream](https://github.com/readthedocs/sphinx_rtd_theme/blob/1.0.0/sphinx_rtd_theme/footer.html#L19)).

There are a few features that I've not implemented, due to them missing in our original fork, or just for ease of upgrading our fork:
1. ~The ability to use a logo in place of site_name in base.html [mkdocs fork](https://github.com/Sparticuz/mkdocs/blob/master/mkdocs/themes/readthedocs/base.html#L91), [upstream](https://github.com/readthedocs/sphinx_rtd_theme/blob/1.0.0/sphinx_rtd_theme/layout.html#L136) **Planned**~ Done in 3cf33d9
image
2. Adding the version under the site_name in base.html, [upstream](https://github.com/readthedocs/sphinx_rtd_theme/blob/1.0.0/sphinx_rtd_theme/layout.html#L154-L164), Not exactly sure what this version is pointing to, is it a git tag, or a config option, etc...? Might be related to #193 Added support in 4c66a1a, but it just displays a string
image
3. Article comments in base.html under page.content, [upstream](https://github.com/readthedocs/sphinx_rtd_theme/blob/1.0.0/sphinx_rtd_theme/layout.html#L215-L219).
4. Build info in footer.html, [upstream](https://github.com/readthedocs/sphinx_rtd_theme/blob/1.0.0/sphinx_rtd_theme/footer.html#L27-L43). I would like to implement this, but I believe I would need to add the [mkdocs-git-revision-date-localized-plugin](https://github.com/timvink/mkdocs-git-revision-date-localized-plugin) plugin
5. ~Upstream has deprecated and removed the 'isogram' version of google analytics, but I wanted comment here before I also removed it.~ **Keeping `isogram` for now**

---

Fixes #2538
Fixes #2537
Fixes #2467
Fixes #1763
Fixes #2239


Sparticuz:
Refs at least these #2538, #2537, #2467, #1763, maybe more?

Sparticuz:
Example of `logo` image, as well as `style_nav_header_background`
image

Sparticuz:
I've added support in 4c66a1a for
```
extra:
  version: "string"
```
It's just a string, doesn't pull anything from git.
image

mkdocs/themes/readthedocs/mkdocs_theme.yml
```diff
@@ -20,3 +20,5 @@ navigation_depth: 4
 titles_only: false
 sticky_navigation: true
 collapse_navigation: true
+# should be url('')
```
oprypin: Hmm I'm not sure about this; I think there's no precedent for asking people to write `url(anything)` in any theme's config

mkdocs/themes/readthedocs/base.html
```diff
@@ -84,36 +85,36 @@
     <nav data-toggle="wy-nav-shift" class="wy-nav-side stickynav">
     <div class="wy-side-scroll">
       <div class="wy-side-nav-search" {% if config.theme.style_nav_header_background %} style="background: {{config.theme.style_nav_header_background}}" {% endif %}>
-    {%- block site_name %}
-        <a href="{{ nav.homepage.url|url }}" class="icon icon-home"> {{ config.site_name }}</a>
-	  {%- endblock %}
-	  {%- block search_button %}
+      {%- block site_name %}
+        <a href="{{ nav.homepage.url|url }}" class="icon icon-home" alt="{% trans %}Docs{% endtrans %}"> {{ config.site_name }}</a>
```
oprypin: What's the meaning of this "Docs" if the original has "Documentation Home"?
[`0.4.3...0.5.0rc1`#diff-0f78e90ab4](https://github.com/readthedocs/sphinx_rtd_theme/compare/0.4.3...0.5.0rc1#diff-0f78e90ab4f9d57568ee00cdd81040c20bfbdaf7f19169e82e10a5e11e249052R121)

mkdocs/themes/readthedocs/base.html
```diff
@@ -61,7 +61,7 @@
     function gtag(){dataLayer.push(arguments);}
     gtag('js', new Date());
 
-    gtag('config', '{{ config.theme.analytics.gtag }}');
+    gtag('config', '{{ config.theme.analytics.gtag }}'{%- if config.theme.analytics.anonymize_ip %}, {'anonymize_ip': true}{%- endif %});
```
oprypin: Why is this analytics option being set in two different places?
Below there's this
`ga('set', 'anonymizeIp', true);`
And in the original that one is the only addition.
[`0.5.0...0.5.1`#diff-0f78e90ab4](https://github.com/readthedocs/sphinx_rtd_theme/compare/0.5.0...0.5.1#diff-0f78e90ab4f9d57568ee00cdd81040c20bfbdaf7f19169e82e10a5e11e249052R238)

mkdocs/themes/readthedocs/base.html
```diff
@@ -122,34 +123,36 @@
 
     <section data-toggle="wy-nav-shift" class="wy-nav-content-wrap">
 
-      {# MOBILE NAV, TRIGGLES SIDE NAV ON TOGGLE #}
-      <nav class="wy-nav-top" role="navigation" aria-label="top navigation">
-        <i data-toggle="wy-nav-top" class="fa fa-bars"></i>
-        <a href="{{ nav.homepage.url|url }}">{{ config.site_name }}</a>
+      {#- MOBILE NAV, TRIGGLES SIDE NAV ON TOGGLE #}
+      <nav class="wy-nav-top" role="navigation" aria-label="{% trans %}Top Navigation{% endtrans %}" {% if config.theme.style_nav_header_background %} style="background: {{config.theme.style_nav_header_background}}" {% endif %}>
```
oprypin: Where is this string "Top Navigation" coming from?
[`0.5.1..1.0.0rc1`#diff-0f78e90ab4](https://github.com/readthedocs/sphinx_rtd_theme/compare/0.5.1..1.0.0rc1#diff-0f78e90ab4f9d57568ee00cdd81040c20bfbdaf7f19169e82e10a5e11e249052R195)

mkdocs/themes/readthedocs/base.html
```diff
@@ -122,34 +123,36 @@
...
 
-      {# PAGE CONTENT #}
+      {#- PAGE CONTENT #}
       <div class="wy-nav-content">
         <div class="rst-content">
-          {% include "breadcrumbs.html" %}
-          <div role="main">
-            <div class="section">
+          {%- include "breadcrumbs.html" %}
+          <div role="main" class="document" itemscope="itemscope" itemtype="http://schema.org/Article">
```
oprypin, line 151: This is not actually added in 1.0.0 but in 0.1.9.
https://github.com/readthedocs/sphinx_rtd_theme/commit/af89f48341ab826cfa6e14425c12fca2078485c3
[`0.5.1..1.0.0rc1`#diff-0f78e90ab4](https://github.com/readthedocs/sphinx_rtd_theme/compare/0.5.1..1.0.0rc1#diff-0f78e90ab4f9d57568ee00cdd81040c20bfbdaf7f19169e82e10a5e11e249052R210)
Do you insist that this should be added anyway?

mkdocs/themes/readthedocs/mkdocs_theme.yml
```diff
@@ -13,10 +13,16 @@ hljs_languages: []
...
+  anonymize_ip: false
 
 include_homepage_in_sidebar: true
 prev_next_buttons_location: bottom
 navigation_depth: 4
 titles_only: false
 sticky_navigation: true
 collapse_navigation: true
+# Should be the css background shorthand, using single quotes.
+style_nav_header_background:
```
oprypin: I'd much prefer this `style_nav_header_background` to be dropped. MkDocs already makes it customary to add CSS to a separate file, and people can just write `.wy-side-nav-search { whatever }`

oprypin (commented):
This is all great work, thank you very much!

I have checked all the content of this pull request and it all makes sense and corresponds well to the actual changes in *sphinx_rtd_theme*.

I have only left small comments.

Now I'm just thinking how we can involve more people to try out this theme to make sure it actually works well on the frontend side.

mkdocs/themes/readthedocs/base.html
Sparticuz: I believe I left it as `Docs` because that's what MkDocs had and I didn't want to change the `msgid` in the `messages.pot` file causing all the previous translation work to be null for that key.

mkdocs/themes/readthedocs/base.html
Sparticuz: This is because MkDocs supports both the old 'isogram' analytics, as well as the newer gtag version. RTD had deprecated the 'isogram' version, but I didn't want to do that without comment.

mkdocs/themes/readthedocs/mkdocs_theme.yml
Sparticuz: I agree. I'll remove it.

mkdocs/themes/readthedocs/mkdocs_theme.yml
Sparticuz: That was an internal comment to myself, I adjusted the comment in a later commit, but also (as discussed below), just removed that functionality.

mkdocs/themes/readthedocs/base.html
Sparticuz: Yep, looks like a typo, RTD states `Mobile navigation menu`. I'll change it to that.

mkdocs/themes/readthedocs/base.html
Sparticuz: I added the extra metadata because it was missing from the MkDocs fork. I'm not sure the benefit, but it's probably something that is for accessibility? It wasn't any structural changes, just the following added.
`class="document" itemscope="itemscope" itemtype="http://schema.org/Article"`
` itemprop="articleBody"`
https://github.com/Sparticuz/mkdocs/blob/master/mkdocs/themes/readthedocs/base.html#L151-L152

mkdocs/themes/readthedocs/base.html
Sparticuz: I see now that MkDocs has [deprecated it as well](https://github.com/mkdocs/mkdocs/commit/80feaf03a42dacdd0740259b647014d5f0f9d839). Let me know if I should just delete the isogram code.

oprypin: No, let's keep it. Thanks for the explanation

mkdocs/themes/readthedocs/base.html
oprypin: Ah my main point was, what if it was omitted from MkDocs for some good reason. But yeah I'm not gonna say anything against.

oprypin:
I was strongly considering merging this PR with individual commits, but now I think it's slightly less organized for that. And, the main reason is that the commits contain large files which we may as well omit from storing in the repo. So I think I'll squash.

Also did you see that CI is failing?

Sparticuz:
I had kindof planned on it being squashed. Most of the commits are really unneeded, but it helped me visualize the changes by going release by release.

I did see the CI was failing, but I have no idea why. The error it's giving doesn't make any sense to me because the IP it's giving IS an IPv4 address.... It's also unrelated to any changes I made.

```
ERROR: test_IP_normalization (tests.config.config_options_tests.IpAddressTest)
  File "/home/runner/work/mkdocs/mkdocs/mkdocs/config/config_options.py", line 258, in run_validation
    host = str(ipaddress.ip_address(host))
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/ipaddress.py", line 53, in ip_address
    raise ValueError('%r does not appear to be an IPv4 or IPv6 address' %
ValueError: '127.000.000.001' does not appear to be an IPv4 or IPv6 address

During handling of the above exception, another exception occurred:

  File "/home/runner/work/mkdocs/mkdocs/mkdocs/tests/config/config_options_tests.py", line 207, in test_IP_normalization
    value = option.validate(None)
  File "/home/runner/work/mkdocs/mkdocs/mkdocs/config/config_options.py", line 130, in validate
  File "/home/runner/work/mkdocs/mkdocs/mkdocs/config/config_options.py", line 260, in run_validation
    raise ValidationError(e)
mkdocs.config.base.ValidationError: '127.000.000.001' does not appear to be an IPv4 or IPv6 address
```

mkdocs/themes/readthedocs/base.html
Sparticuz: A repo search for `itemprop` didn't reveal anything, so unless anyone objects, I always prefer complete syntax.

pawamoy:
It's possible that GitHub actions upgraded their 3.8 version to a later patch that changed how IP addresses are normalized.

oprypin:
OK! Yes, that fails on 'master', reproduced it here: https://github.com/mkdocs/mkdocs/runs/3723870272 and will fix it here: https://github.com/mkdocs/mkdocs/pull/2587

oprypin:
I have not decided when to merge it, and whether it's appropriate for a patch release, but this is definitely accepted.

mkdocs/themes/readthedocs/mkdocs_theme.yml
```diff
@@ -13,10 +13,13 @@ hljs_languages: []
...
+  # Set to true to anonymize the ip for Google Analytics
+  anonymize_ip: false
 
 include_homepage_in_sidebar: true
 prev_next_buttons_location: bottom
 navigation_depth: 4
 titles_only: false
 sticky_navigation: true
 collapse_navigation: true
+logo:
```
oprypin: Wouldn't it be less confusing to write `logo: null`?

Sparticuz:
I agree, I don't think this is a patch release. There were very few structural changes to the HTML, just a few things added. Mainly for the logo and version in the header, which are both optional additions. Actually the largest change was probably dropping Modernizr and replacing it with html5shiv. I guess, technically, that would have the possibility of breaking changes if someone were using the Modernizr CSS classes in their `css-extra.css` file.

For #2467, my PR only addresses the readthedocs theme. Would you like me to also add a fix for the mkdocs theme, or would you prefer a separate PR? (I could also take a look at updating that theme as well)

What are your thoughts on 3 and 4 above? 

For 3, It doesn't appear that MkDocs supports comments, but that doesn't necessarily mean that someone couldn't use something like disqus with a override on the comments block? Adding this should be trivial.

For 4, That would add this stuff
image
to this area of our fork
image
IMO, this might be better as a separate PR, as it might need to add in a few plugins

oprypin:
> I agree, I don't think this is a patch release. There were very few structural changes to the HTML, just a few things added. 

Uhh I think there's some misunderstanding here. I was thinking whether it's appropriate for 1.2.3 (patch release) or 1.3.0. So which one do you mean?

> For #2467, my PR only addresses the readthedocs theme. Would you like me to also add a fix for the mkdocs theme, or would you prefer a separate PR?

You're correct. Definitely a separate PR would be best. Thank you if you take up that work.

(3) and (4) I have not evaluated, but definitely let's not do those in this PR

Sparticuz:
> > I agree, I don't think this is a patch release. There were very few structural changes to the HTML, just a few things added.
> 
> Uhh I think there's some misunderstanding here. I was thinking whether it's appropriate for 1.2.3 (patch release) or 1.3.0. So which one do you mean?

I think it's a minor release, not a patch release.

oprypin:
@Sparticuz  I would agree that the naming is unfortunate, but my use of it was correct as per https://semver.org/

But regardless, could you clarify your opinion, do you think it's appropriate for 1.2.3?

Sparticuz:
> @Sparticuz I would agree that the naming is unfortunate, but my use of it was correct as per https://semver.org/
> 
> But regardless, could you clarify your opinion, do you think it's appropriate for 1.2.3?

Talking for the Theme itself, according to semver, it's a major release. Modernizr got removed, which removes a bunch of css classes that get added by the modernizr javascript. If a user was using those css classes, it would be a breaking change. (Also jquery went from v2 to v3, but I'm not sure jquery adheres to semver)

As for the version of MkDocs, it's a decision of whether or not a major update of a dependency justifies a major update of MkDocs itself. I'm not familiar with what Semver says for a breaking change in a dependency. (If we consider the readthedocs theme a dependency of MkDocs)

If you wanted, I could add in
```
theme:
  name: readthedocs
  modernizr: True
```
which would keep modernizr. I'm not familiar with the local dev server, but maybe a deprecation warning could be added if `config.theme.modernizr` was True. Then when MkDocs v2 is published, remove modernizr entirely.

oprypin:
Hmm well sorry for bringing up semver, because MkDocs doesn't actually follow it directly. It's more like "anything goes"."no major breaking changes"."no obvious breaking changes"

Anyway, from what you described (and I'm still not fully clear on it), seems like something *could* reasonably break, so then it's more like 1.3.0.
And to be fully clear again, it is fully out of the question to push it to 2.0.0, I was only considering 1.2.3 as the other option.

And thanks for suggesting a compatibility option, but uhh let's just rule that out entirely due to the big complexity.

oprypin:
Thank you very much!!

marcelstoer:
@oprypin 

> I'm just thinking how we can involve more people to try out this theme

You could have tagged the people who created the issues this MR fixes (I am one of them). That way I'd have found out and could have helped testing _before_ it was merged.

@Sparticuz thank you so much for all the hard work!

oprypin:
Well it's not a problem to try it out now, certainly please do :)

ajthealchemist:
> Example of `logo` image, as well as `style_nav_header_background` image

Hi. Why is this not working for me despite including the following lines in my `mkdocs.yml` file?
`theme:
   name: readthedocs
   logo: img/logo.png
`
This is my `mkdocs` version:
`mkdocs, version 1.2.3 from /home/username/.local/lib/python3.8/site-packages/mkdocs (Python 3.8)`

Sparticuz:
> This is my `mkdocs` version:
> 
> ```
> mkdocs, version 1.2.3 from /home/username/.local/lib/python3.8/site-packages/mkdocs (Python 3.8)
> ```

I don't believe this PR has been released yet. Should be in 1.3

ajthealchemist:
> I don't believe this PR has been released yet. Should be in 1.3

I see. Thanks for the clarification. I thought this feature was already released. That being said, is there a tentative release schedule? Can't wait for this feature! <3

oprypin:
Sorry, there is no tentative release schedule.

PR #2594, opened 2021-10-04
Add --no-history option to not push parent history for gh_deploy

pryme-svg:


mkdocs/__main__.py
```diff
@@ -91,6 +91,7 @@ def __init__(self, log_name='mkdocs', level=logging.INFO):
 remote_name_help = ("The remote name to commit to for Github Pages. This "
                     "overrides the value specified in config")
 force_help = "Force the push to the repository."
+no_history_help = "Force new commit without parent history."
```
oprypin: Uhh so I can easily guess what this description means, but still not fully sure. Also maybe the severity of this action is underrepresented.

So maybe something like this?

```suggestion
no_history_help = "Replace the whole Git history with one new commit."
```

oprypin: Oops, now this triggered a lint warning that the line is too long

oprypin:
Actually you know what? I have tried all kinds of combinations and I still can't guess what this flag does, or, to me it appears to not do anything. I used `mkdocs gh-deploy -b test --no-history` and it still creates a new commit on top of old ones.

oprypin:
Is it supposed to affect the time of creation of a new branch, or is it about an existing destination branch?

pawamoy (commented):
So it seems this PR simply makes use of [ghp-import's `no_history` option](https://github.com/c-w/ghp-import#python-usage), which in turn creates a commit without a parent one. I'll have to test it on one of my projects to really see how it behaves.

oprypin (approved):
Sorry for the delay.

Indeed it works as per the description.

> Replace the whole Git history with one new commit.

PR #2602, opened 2021-10-08
search plugin: do not crash on missing lang options, inform instead

ultrabug:
- fallback to english on missing search.lang option
- support search.lang territory pt_BR -> pt

### DIFF
--- a/mkdocs/tests/search_tests.py
+++ b/mkdocs/tests/search_tests.py
@@ -68,7 +68,7 @@ class SearchConfigTests(unittest.TestCase):
         value = option.validate(['en', 'foo'])
         self.assertEqual(['en'], value)
 
-    def test_lang_good_and_bad_and_territory(self):
+    def test_lang_missing_and_with_territory(self):
         option = search.LangOption()
         value = option.validate(['zh_CN', 'pt_BR', 'fr'])
         self.assertEqual(['fr', 'en', 'pt'], value)


mkdocs/contrib/search/__init__.py
```diff
@@ -14,19 +14,29 @@ class LangOption(config_options.OptionallyRequired):
...
         if isinstance(value, str):
             value = [value]
         elif not isinstance(value, (list, tuple)):
             raise config_options.ValidationError('Expected a list of language codes.')
-        for lang in value:
-            if lang != 'en' and not self.lang_file_exists(lang):
-                raise config_options.ValidationError(
-                    f'"{lang}" is not a supported language code.'
-                )
+        for lang in list(value):
```
oprypin, line 27: What's the reason to call `list(value)` here? From basic understanding, there is no difference at all vs not calling it.

mkdocs/contrib/search/__init__.py
```diff
@@ -14,19 +14,29 @@ class LangOption(config_options.OptionallyRequired):
     """ Validate Language(s) provided in config are known languages. """
 
     def lang_file_exists(self, lang):
```
oprypin: This function is no longer about existence, should rename it

mkdocs/contrib/search/__init__.py
```diff
@@ -14,19 +14,29 @@ class LangOption(config_options.OptionallyRequired):
...
-        path = os.path.join(base_path, 'lunr-language', f'lunr.{lang}.js')
-        return os.path.isfile(path)
+        for lang_part in lang.split("_"):
+            lang_part = lang_part.lower()
+            if os.path.isfile(os.path.join(base_path, 'lunr-language', f'lunr.{lang_part}.js')):
+                return lang_part
+        return False
```
oprypin: Perhaps return None now

ultrabug:
Good for me now :+1:

mkdocs/tests/search_tests.py
```diff
@@ -60,11 +60,18 @@ def test_lang_bad_type(self):
...
+        value = option.validate(['foo'])
+        self.assertEqual(['en'], value)
 
     def test_lang_good_and_bad_code(self):
         option = search.LangOption()
-        self.assertRaises(ValidationError, option.validate, ['en', 'foo'])
+        value = option.validate(['en', 'foo'])
+        self.assertEqual(['en'], value)
+
+    def test_lang_good_and_bad_and_territory(self):
```
oprypin: Uhhh let's not accidentally imply that some particular territory is bad :scream: 

What would the word be... "missing"?

ultrabug:
Thanks for the review @oprypin

oprypin:
Ah thanks for the great fix.

@ultrabug so you probably saw that I mentioned that a release is coming on Sunday [today]
but it seems like there's still some language PRs that we could merge now. So let's not rush with the release.

ultrabug:
Yes mate indeed, I'll check the pt_BR PR out later today, thanks

PR #2606, opened 2021-10-10
Depend on 'colorama' on Windows to fix color output in terminal

### LINKED ISSUE #2570: Support terminal coloring on Windows
E.g. we might use the [colorama](https://pypi.org/project/colorama/) library, otherwise the output looks like this in cmd.exe:

```
←[33mWARNING  -  ←[0mmkdocs_autorefs.plugin: cookbook-reference.md: Could not find cross-reference target '['a']'
←[33mWARNING  -  ←[0mmkdocs_autorefs.plugin: cookbook-reference.md: Could not find cross-reference target '['a']'
```

oprypin:
Adding the explicit dependency, because

1. 'click' prior to v8.0 did not have this dependency, it was used "only if present"
2. We *are* importing it, let's declare it as such

Fixes #2570

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -12,6 +12,15 @@ from mkdocs import utils
 from mkdocs import config
 from mkdocs.commands import build, gh_deploy, new, serve
 
+
+if sys.platform.startswith("win"):
+    try:
+        import colorama
+    except ImportError:
+        pass
+    else:
+        colorama.init()
+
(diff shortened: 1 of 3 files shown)


oprypin:
Related code in Click: https://github.com/pallets/click/commit/d7fac6e16099005c642f4dcb71d84a091a787073#diff-46505dbb054df22f8aa1fc5113faee7178f8ceaf4b6909a1d62cc528f381eb8dR496

PR #2612, opened 2021-10-11
Update codebase to Python >= 3.6

mondeja:
- Replaced some `.format` calls with f-strings.
- Removed `'r'` mode argument for `open` function as it is uneeded, but maybe maintainers want to keep it explicit?
- Removed `__future__` import and `coding: utf-8` comment from a file.


oprypin:
Note that in https://github.com/mkdocs/mkdocs/pull/2301 I already went over all the occurrences of `format` in the codebase and changed to f-strings only in the cases where I thought it would be an improvement. Most of these I think are harder to read with this change.

mkdocs/commands/serve.py
```diff
@@ -42,7 +42,7 @@ def builder():
             **kwargs
         )
         # Override a few config settings after validation
-        config['site_url'] = 'http://{}{}'.format(config['dev_addr'], mount_path(config))
+        config['site_url'] = f'http://{config["dev_addr"]}{mount_path(config)}'
```
oprypin: I don't like this change

mkdocs/config/config_options.py
```diff
@@ -426,15 +425,15 @@ def post_validation(self, config, key_name):
                 ("The 'docs_dir' should not be within the 'site_dir' as this "
                  "can mean the source files are overwritten by the output or "
                  "it will be deleted if --clean is passed to mkdocs build."
-                 "(site_dir: '{}', docs_dir: '{}')"
-                 ).format(config['site_dir'], config['docs_dir']))
+                 f"(site_dir: '{config['site_dir']}',"
+                 f" docs_dir: '{config['docs_dir']}')"))
```
oprypin: I don't like this change

mkdocs/config/config_options.py
```diff
@@ -426,15 +425,15 @@ def post_validation(self, config, key_name):
...
         elif (config['site_dir'] + os.sep).startswith(config['docs_dir'].rstrip(os.sep) + os.sep):
             raise ValidationError(
                 ("The 'site_dir' should not be within the 'docs_dir' as this "
                  "leads to the build directory being copied into itself and "
                  "duplicate nested files in the 'site_dir'."
-                 "(site_dir: '{}', docs_dir: '{}')"
-                 ).format(config['site_dir'], config['docs_dir']))
+                 f"(site_dir: '{config['site_dir']}',"
+                 f" docs_dir: '{config['docs_dir']}')"))
```
oprypin: I don't like this change

mkdocs/config/config_options.py
```diff
@@ -515,9 +514,7 @@ def run_validation(self, value):
         if config_types.issubset({str, dict}):
             return value
 
-        raise ValidationError("Invalid pages config. {} {}".format(
-            config_types, {str, dict}
-        ))
+        raise ValidationError(f"Invalid pages config. {config_types} {{str, dict}}")
```
oprypin: I don't like this change.
Also actually quite a strange message in the first place, maybe we should check its history.

mkdocs/structure/nav.py
```diff
@@ -62,7 +62,7 @@ def ancestors(self):
         return [self.parent] + self.parent.ancestors
 
     def _indent_print(self, depth=0):
-        ret = ['{}{}'.format('    ' * depth, repr(self))]
+        ret = [f'{"    " * depth}{repr(self)}']
```
oprypin: I don't like this change

mkdocs/structure/nav.py
```diff
@@ -92,7 +92,7 @@ def ancestors(self):
         return [self.parent] + self.parent.ancestors
 
     def _indent_print(self, depth=0):
-        return '{}{}'.format('    ' * depth, repr(self))
+        return f'{"    " * depth}{repr(self)}'
```
oprypin: I don't like this change

Also, huh, are there 2x `_indent_print`?

mkdocs/structure/pages.py
```diff
@@ -54,10 +54,10 @@ def __ne__(self, other):
 
     def __repr__(self):
         title = f"'{self.title}'" if (self.title is not None) else '[blank]'
-        return "Page(title={}, url='{}')".format(title, self.abs_url or self.file.url)
+        return f"Page(title={title}, url='{self.abs_url or self.file.url}')"
```
oprypin: I don't like this change.

```suggestion
        url = self.abs_url or self.file.url
        return f"Page(title={title}, url='{url}')"
```

mkdocs/structure/pages.py
```diff
@@ -54,10 +54,10 @@ def __ne__(self, other):
...
 
     def _indent_print(self, depth=0):
-        return '{}{}'.format('    ' * depth, repr(self))
+        return f'{"    " * depth}{repr(self)}'
```
oprypin: I don't like this change

mkdocs/theme.py
```diff
@@ -55,10 +55,9 @@ def __init__(self, name=None, **user_config):
         self._vars['locale'] = localization.parse_locale(self._vars['locale'])
 
     def __repr__(self):
-        return "{}(name='{}', dirs={}, static_templates={}, {})".format(
-            self.__class__.__name__, self.name, self.dirs, list(self.static_templates),
-            ', '.join(f'{k}={v!r}' for k, v in self._vars.items())
-        )
+        return (f"{self.__class__.__name__}(name='{self.name}',"
+                f" dirs={self.dirs}, static_templates={list(self.static_templates)},"
+                f" {', '.join(f'{k}={v!r}' for k, v in self._vars.items())})")
```
oprypin: I don't like this change

mkdocs/utils/meta.py
```diff
@@ -92,7 +92,7 @@ def get_data(doc):
             m2 = META_MORE_RE.match(line)
             if m2 and key:
                 # Add another line to existing key
-                data[key] += ' {}'.format(m2.group('value').strip())
+                data[key] += f' {m2.group("value").strip()}'
```
oprypin: I don't like this change

setup.py
```diff
@@ -40,7 +40,8 @@ def get_packages(package):
     os.system("python setup.py sdist bdist_wheel")
     os.system("twine upload dist/*")
     print("You probably want to also tag the version now:")
-    print("  git tag -a {0} -m 'version {0}'".format(get_version("mkdocs")))
+    version = get_version("mkdocs")
+    print(f"  git tag -a {version} -m 'version {version}'")
```
oprypin, line 44: Nice change. Good example of guideline (3) that I mention.

oprypin:
The last commit - I don't think it's good to clump it together with this pull request. Please drop it.

Also not sure about "Rewrite Nav config type validation error message" - ideally would be separate

ultrabug:
I'm stupid and merged in the wrong order @mondeja , I apologize for the conflict here...

PR #2616, opened 2021-10-12
Add hint about `-f/--config-file` in configuration documentation

### LINKED ISSUE #2581: consider adding reference to `--config-file` to configuration documentation
Much like https://github.com/mkdocs/mkdocs/issues/1954#issuecomment-582487448 I had no idea this option even existed. I checked the top-level help and I checked https://www.mkdocs.org/user-guide/configuration/ and neither makes any mention of this option which is very helpful! I'm not sure where the best place is for it, I only want to point out that this was not something myself (nor another group who uses mkdocs I was recently talking to) had any idea existed and a small change to documentation would likely improve this.

mondeja:
Closes #2581

Note that the current statement says "...are **always** configured...", which is not true.

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -8,7 +8,7 @@ Guide to all available configuration settings.
 
 Project settings are configured by default using a YAML configuration file in
 the project directory named `mkdocs.yml`. You can specify another path for it
-by using the `-f/--config-file` option (see `mkdocs build --help`).
+by using the `-f`/`--config-file` option (see `mkdocs build --help`).
 
 As a minimum, this configuration file must contain the `site_name` and
 `site_url` settings. All other settings are optional.


docs/user-guide/configuration.md
```diff
@@ -6,8 +6,9 @@ Guide to all available configuration settings.
 
 ## Introduction
 
-Project settings are always configured by using a YAML configuration file in the
-project directory named `mkdocs.yml`.
+Project settings are configured by default using a YAML configuration file in
+the project directory named `mkdocs.yml`. You can specify another path for it
+by using the `-f/--config-file` option (see `mkdocs build --help`).
```
oprypin:
```suggestion
by using the `-f`/`--config-file` option (see `mkdocs build --help`).
```

Important distinction because, as written, this means just one thing to copy literally.

oprypin:
Thanks

PR #2618, opened 2021-10-12
Fix the code to abbreviate Python's version

oprypin:
This breaks for Python >= 3.10

Extracting just this part from #2617


oprypin:
Thanks @mondeja for finding and fixing this

PR #2621, opened 2021-10-12
Re-format translation files to pass a lint check

oprypin:


oprypin:
This is where I noticed that the file *mkdocs/themes/**readthedocs**/locales/pt_BR/LC_MESSAGES/messages.po* https://github.com/mkdocs/mkdocs/pull/2535 is just a copy of the one from **mkdocs** theme.
The reformat isn't faulty, it's doing its best to match those translations to the other theme.
So the translation is indeed incomplete.

PR #2626, opened 2021-10-15
Remove line length limit from Markdown documents

oprypin:
Also try it out on some part of one document.

Note that I do *not* intend to automatically change all documents like that.

---

Disadvantages of forced line length:

* I really hate manually wrapping and re-wrapping stuff
* It makes `git blame` confusing

Advantages of forced line length:

?????

### DIFF
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -2,8 +2,8 @@
   // Enable all markdownlint rules
   "default": true,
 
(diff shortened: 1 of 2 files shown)


oprypin:
FYI @Andre601 from https://github.com/mkdocs/mkdocs/discussions/2442#discussioncomment-836487

Issue #2632, opened 2021-10-23
Support custom directories to watch when running `mkdocs serve` 


steven-terrana:
Hello! 

Thanks for your work on `mkdocs`. 

I'm looking for feedback on the following feature request before putting effort into the development. 

## Feature

Add a `-w` / `--watch-dir` command line option to `mkdocs serve`. 

The directories (or globs?) provided would be watched and changes would trigger a rebuild/reload.

## Context

The problem i'm trying to solve is that i'm currently using the [mkdocs-gen-files](https://github.com/oprypin/mkdocs-gen-files) plugin to add markdown files from outside the `docs_dir` into the site. 

This works great when building the documentation, but doesn't work for local development via `mkdocs serve`. 

The current hack i'm using to get around this is to add the extra directory to watch via the `theme.custom_dir` property and then adding `--watch-theme` to `mkdocs serve`. 

it works but.. ew. :) 

<hr>

If this idea sounds like a feature you support, I'd be happy to give the implementation a shot!

oprypin:
Yes, I think it's time to add this config.
It would probably be a config in YAML actually, named `watch`, with the flag `-w`/`--watch` augmenting it.

The watching implementation supports both files and directories seamlessly already.
So implementing it is easy, just adding a `watch` call. You could try doing that, yeah.

Note that there is "prior art": the "mkdocstrings" plugin basically decided to expose such functionality generically itself, even though it's not specific to the plugin.
https://github.com/mkdocstrings/mkdocstrings/search?q=watch
https://github.com/mkdocstrings/mkdocstrings/blob/0b4b2f5259f9607ecb82f11ae2372ddeb156bca9/src/mkdocstrings/plugin.py#L130-L132

So indeed, in the interest of not duplicating such plugin-specific configs, it's probably good to add this config generically.

PR #2642, opened 2021-10-25
Support custom directories to watch when running mkdocs serve

steven-terrana:
Implements #2632

* adds a `watch` property to the `mkdocs.yaml` schema. Accepts a list of directories to watch.
* adds a `-w`/`--watch` command line option to `mkdocs serve` that can be passed multiple times
* options from `mkdocs.yaml` and CLI flags are combined
* docs updated

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -371,6 +371,12 @@ watch:
 Allows a custom default to be set without the need to pass it through the `-w`/`--watch`
 option every time the `mkdocs serve` command is called.
 
+!!! Note
+
+    The paths provided via the configuration file are relative to the configuration file.
+
+    The paths provided via the `-w`/`--watch` CLI parameters are not. 
+
 ### use_directory_urls
 
 This setting controls the style used for linking to pages within the
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -355,6 +355,22 @@ extra:
 
 ## Preview controls
 
+## Live Reloading
+
+### watch
+
+Determines additional directories to watch when running `mkdocs serve`.
+Configuration is a YAML list.
+
+```yaml
+watch:
+- directory_a
+- directory_b
+```
+
+Allows a custom default to be set without the need to pass it through the `-w`/`--watch`
+option every time the `mkdocs serve` command is called.
+
 ### use_directory_urls
 
 This setting controls the style used for linking to pages within the
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -172,7 +172,7 @@ def cli():
 @click.option('--no-livereload', 'livereload', flag_value='no-livereload', help=no_reload_help)
 @click.option('--dirtyreload', 'livereload', flag_value='dirty', help=dirty_reload_help)
 @click.option('--watch-theme', help=watch_theme_help, is_flag=True)
-@click.option('-w', '--watch', help=watch_help, type=click.Path(), multiple=True, default=[])
+@click.option('-w', '--watch', help=watch_help, type=click.Path(exists=True), multiple=True, default=[])
 @common_config_options
 @common_options
 def serve_command(dev_addr, livereload, watch, **kwargs):
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -99,7 +99,7 @@ ignore_version_help = "Ignore check that build is not being deployed with an old
 watch_theme_help = ("Include the theme in list of files to watch for live reloading. "
                     "Ignored when live reload is not used.")
 shell_help = "Use the shell when invoking Git."
-watch_help = ("A directory to watch for live reloading. "
+watch_help = ("A directory or file to watch for live reloading. "
               "Can be supplied multiple times.")
 
 
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -99,6 +99,8 @@ ignore_version_help = "Ignore check that build is not being deployed with an old
 watch_theme_help = ("Include the theme in list of files to watch for live reloading. "
                     "Ignored when live reload is not used.")
 shell_help = "Use the shell when invoking Git."
+watch_help = ("A directory to watch for live reloading. "
+              "Can be supplied multiple times.")
 
 
 def add_options(opts):
@@ -170,11 +172,12 @@ def cli():
 @click.option('--no-livereload', 'livereload', flag_value='no-livereload', help=no_reload_help)
 @click.option('--dirtyreload', 'livereload', flag_value='dirty', help=dirty_reload_help)
 @click.option('--watch-theme', help=watch_theme_help, is_flag=True)
+@click.option('-w', '--watch', help=watch_help, multiple=True, default=[])
 @common_config_options
 @common_options
-def serve_command(dev_addr, livereload, **kwargs):
+def serve_command(dev_addr, livereload, watch, **kwargs):
     """Run the builtin development server"""
-    serve.serve(dev_addr=dev_addr, livereload=livereload, **kwargs)
+    serve.serve(dev_addr=dev_addr, livereload=livereload, watch=watch, **kwargs)
 
 
 @cli.command(name="build")
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -172,7 +172,7 @@ def cli():
 @click.option('--no-livereload', 'livereload', flag_value='no-livereload', help=no_reload_help)
 @click.option('--dirtyreload', 'livereload', flag_value='dirty', help=dirty_reload_help)
 @click.option('--watch-theme', help=watch_theme_help, is_flag=True)
-@click.option('-w', '--watch', help=watch_help, multiple=True, default=[])
+@click.option('-w', '--watch', help=watch_help, type=click.Path(), multiple=True, default=[])
 @common_config_options
 @common_options
 def serve_command(dev_addr, livereload, watch, **kwargs):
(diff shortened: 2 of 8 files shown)


mkdocs/commands/serve.py
```diff
@@ -74,6 +78,12 @@ def error_handler(code):
                 for d in config['theme'].dirs:
                     server.watch(d)
 
+            # convert to set in case CLI params match config file settings
+            watch_list = set(config["watch"])
+            for item in watch_list:
+                log.info(f"Watching additional directory: {item}")
```
oprypin: It's not necessarily a directory

mkdocs/commands/serve.py
```diff
@@ -74,6 +78,12 @@ def error_handler(code):
                 for d in config['theme'].dirs:
                     server.watch(d)
 
+            # convert to set in case CLI params match config file settings
+            watch_list = set(config["watch"])
```
oprypin: `set` has non-deterministic order. You might use `dict.fromkeys` instead of `set`.

mkdocs/commands/serve.py
oprypin: Yeah that's an easy solution

oprypin: I'm wondering if `info` log level is too noisy. Wouldn't those pop up every time and fill the screen just by themselves if there are many watched locations?

Or it could be turned into one log message for all directories. But not sure.

And should the `docs_dir` itself be announced as watched? Because why not?

mkdocs/commands/serve.py
oprypin: Ordering isn't important indeed, but I always have a hunch that it'll cause some bug that appears only randomly. Why do things non-deterministically if you can just as easily do deterministically.

```suggestion
            watch_list = dict.fromkeys(config["watch"])
```

mkdocs/commands/serve.py
oprypin: I've been meaning to try it out and see how each option feels. Haven't had the opportunity yet, though.

mkdocs/__main__.py
```diff
@@ -170,11 +172,12 @@ def cli():
 @click.option('--no-livereload', 'livereload', flag_value='no-livereload', help=no_reload_help)
 @click.option('--dirtyreload', 'livereload', flag_value='dirty', help=dirty_reload_help)
 @click.option('--watch-theme', help=watch_theme_help, is_flag=True)
+@click.option('-w', '--watch', help=watch_help, multiple=True, default=[])
```
oprypin:
```suggestion
@click.option('-w', '--watch', help=watch_help, type=click.Path(), multiple=True, default=[])
```

Improves help message

```diff
-  -w, --watch PATH                A directory to watch for live reloading. Can
+  -w, --watch TEXT                A directory to watch for live reloading. Can
                                   be supplied multiple times.
```

mkdocs/config/defaults.py
```diff
@@ -120,4 +120,7 @@ def get_schema():
         # A key value pair should be the string name (as the key) and a dict of config
         # options (as the value).
         ('plugins', config_options.Plugins(default=['search'])),
+
+        # a list of extra directories to watch while running `mkdocs serve`
+        ('watch', config_options.Type(list, default=[]))
```
oprypin: I think the path will have to be relativized to the location of `mkdocs.yml`, as here:

https://github.com/mkdocs/mkdocs/blob/988076794002394f12d5e393b0205f4f795087a3/mkdocs/config/config_options.py#L369-L370

Well, I *think* that's what would be the most logical as per what I've seen from MkDocs so far.
Feel free to argue otherwise.

---

This would probably require a new type in *config_optons.py*. `ListOfPaths` or something.
During validation it could reuse `FilesystemObject`.

I would recommend turning it non-abstract:
```diff
 class FilesystemObject(Type):
+    existence_test = staticmethod(os.path.exists)
+    name = 'path'
```

Relevant implementation example:

https://github.com/mkdocs/mkdocs/blob/988076794002394f12d5e393b0205f4f795087a3/mkdocs/config/config_options.py#L494

---

Note that we probably should *not* check for existence of the path (`exists=False` should be kept). Not sure though.

mkdocs/__main__.py
```diff
@@ -99,6 +99,8 @@ def __init__(self, log_name='mkdocs', level=logging.INFO):
 watch_theme_help = ("Include the theme in list of files to watch for live reloading. "
                     "Ignored when live reload is not used.")
 shell_help = "Use the shell when invoking Git."
+watch_help = ("A directory to watch for live reloading. "
```
oprypin:
```suggestion
watch_help = ("A directory or file to watch for live reloading. "
```

mkdocs/commands/serve.py
oprypin: I have decided on moving the logging to livereload. That way it can also report values that originate from plugins' calls to `server.watch`.

oprypin: And I wrote the implementation of this as well:
https://github.com/steven-terrana/mkdocs/pull/1

mkdocs/config/defaults.py
oprypin: Update: we should, in fact, check for existence (just set `exists=True`), otherwise it throws an `OSError` anyway after startup

oprypin: To clarify why this is important:

```yaml
watch:
    - foo
```
`mkdocs serve --config-file=subdir/mkdocs.yml -w bar`

I want this to watch the `foo` that is adjacent to `mkdocs.yml`.
The difference in the output being:

```diff
-Watching paths for changes: 'subdir/docs', 'subdir/mkdocs.yml', 'foo', 'bar'
+Watching paths for changes: 'subdir/docs', 'subdir/mkdocs.yml', 'subdir/foo', 'bar'
```

(`-w` flag not relativized intentionally)

oprypin: I'm leaving this to you btw.

And some tests for the new config option would be nice :D

mkdocs/commands/serve.py
```diff
@@ -43,7 +43,10 @@ def builder():
         )
 
         # combine CLI watch arguments with config file values
-        config["watch"].extend(watch)
+        if not config["watch"] is None:
```
oprypin: Have you observed a situation where this is `None`? I'm surprised. Maybe it's better to somehow prevent this situation

mkdocs/config/config_options.py
```diff
@@ -406,6 +407,37 @@ class File(FilesystemObject):
...
+        if len(value) == 0:
+            return
+        paths = []
+        for path in value:
+            if self.config_dir and not os.path.isabs(path):
+                path = os.path.join(self.config_dir, path)
+            if not os.path.exists(path):
+                raise ValidationError(f"The path {path} does not exist.")
+            path = os.path.abspath(path)
+            assert isinstance(path, str)
```
oprypin:
```suggestion
```

I don't know what this assert achieves, seems actually impossible 😀

mkdocs/tests/config/config_options_tests.py
```diff
@@ -494,6 +494,50 @@ def test_dir_is_config_dir_fails(self):
...
+
+    def test_valid_path(self):
+        paths = [os.path.dirname(__file__)]
+        option = config_options.ListOfPaths()
+        option.validate(paths)
+
+    def test_missing_path(self):
+        paths = [os.path.join("doesnt", "exist", "i", "hope")]
+        option = config_options.ListOfPaths()
+        self.assertRaises(config_options.ValidationError, option.validate, paths)
```
oprypin:
```suggestion
        with self.assertRaises(config_options.ValidationError):
            option.validate(paths)
```
as per https://github.com/mkdocs/mkdocs/pull/2653

mkdocs/tests/config/config_options_tests.py
```diff
@@ -494,6 +494,50 @@ def test_dir_is_config_dir_fails(self):
...
+
+    def test_empty_list(self):
+        paths = []
+        option = config_options.ListOfPaths()
+        option.validate(paths)
+
+    def test_non_list(self):
+        paths = os.path.dirname(__file__)
+        option = config_options.ListOfPaths()
+        self.assertRaises(config_options.ValidationError, option.validate, paths)
```
oprypin:
```suggestion
        with self.assertRaises(config_options.ValidationError):
            option.validate(paths)
```
as per https://github.com/mkdocs/mkdocs/pull/2653

mkdocs/tests/config/config_options_tests.py
```diff
@@ -494,6 +494,50 @@ def test_dir_is_config_dir_fails(self):
...
 
+class ListOfPathsTest(unittest.TestCase):
+
+    def test_valid_path(self):
+        paths = [os.path.dirname(__file__)]
+        option = config_options.ListOfPaths()
+        option.validate(paths)
+
+    def test_missing_path(self):
+        paths = [os.path.join("doesnt", "exist", "i", "hope")]
```
oprypin:
```suggestion
        paths = [os.path.join("does", "not", "exist", "i", "hope")]
```
the last linter failure 😀😖

oprypin:
Thank you!

docs/user-guide/configuration.md
```diff
@@ -355,6 +355,28 @@ extra:
...
+```
+
+Allows a custom default to be set without the need to pass it through the `-w`/`--watch`
+option every time the `mkdocs serve` command is called.
+
+!!! Note
+
+    The paths provided via the configuration file are relative to the configuration file.
+
+    The paths provided via the `-w`/`--watch` CLI parameters are not.
```
Andre601, line 378: Bit late to the party, but what exactly would it then be? What would it's "position" be based of? Where the command is executed? That should've been mentioned here IMO, so maybe in a future PR/Commit could this be changed?

PR #2643, opened 2021-10-25
Fix theme.js

darkdragon-001:
This script assumes document name being encoded in hash portion of the URI which is not the case here.

Pretty-print diff:
```diff
-var n=encodeURI(window.location.hash)||"#";
+var n="./";
```

Fixes #2010

PR #2650, opened 2021-10-27
Fix theme.js

### LINKED ISSUE #2010: ReadTheDocs theme expands everthing when opening a page
As soon as any page is selected, everything in the navigation gets expanded.

https://sphinx-rtd-theme.readthedocs.io/en/stable doesn't show this, it only expands the current page while everything else gets collapsed. This is the correct behaviour.

I was waiting for the new version exactly for this feature :(

darkdragon-001:
When no hash partition is given, use the current page instead of the empty hash which appears multiple times!

Pretty-print diff:
```diff
-var n=encodeURI(window.location.hash)||"#";
+var n=encodeURI(window.location.hash)||"./";
```

Multiple usage of `#` appears when
```jinja2
href="{% if not nav_item.is_section %}{{ nav_item.url|url }}{% else %}#{% endif %}"
```
in `mkdocs/themes/readthedocs/nav.html`.

Replaces #2643
Fixes #2010


oprypin:
The file "readthedocs/js/theme.js" must be an exact copy of the upstream theme. Custom edits to it will not be accepted.

Additionally, the solution you're suggesting is merely a workaround. 3-level-deep navigation is completely broken and you're just making it broken in a different (although less severe) way.

oprypin:
@darkdragon-001 Ah! That may actually be a useful direction to take. Sorry that I've been so harsh on this issue

PR #2667, opened 2021-11-06
Remove duplicated sentences in docs/deploying

fortierq:
Warning about gh-deploy overwriting gh-pages modifications appear twice. I don't think this it intended, is it?


oprypin (approved):

oprypin:
It could be intended (there's some logic to it), but seems better without the duplicate, so, thanks.

oprypin:
It could be intended (there's some logic to it), but seems better without the duplicate, so, thanks.

PR #2684, opened 2021-11-28
Migrate docs from "admonition" to "callouts" extension

oprypin:
The result is exactly equivalent, just different syntax -- one that doesn't look weird in vanilla Markdown parsers such as on GitHub.

Disclaimer: it's my own extension https://github.com/oprypin/markdown-callouts/


ultrabug:
This adds a new dependency to the project with a limited benefit since we still use admonition extension anyway... I'm not so sure this is worth it for the whole mkdocs user base.

facelessuser:
I kind of agree. I have no intentions to use callouts at this time. I do not suggest that callouts is inferior in any way, but simply I don't plan on changing all of my documentation for a new extension.

I feel there is nothing wrong plugging it in docs as an alternative, some may when prefer it,  but I'd like to include additional Markdown extensions if/when I deem that I need them.

If it was for some integral, new MkDocs feature, I could understand, but in this case, it doesn't seem to be doing that, just providing an alternative to existing solutions that most people are more than used to using at this point.

facelessuser:
Never mind, I misunderstood. This is required only for MkDocs documentation from what I see. This is not being forced on those installing MkDocs from what I can tell.

ultrabug:
Oh yeah? I was sure it did based on the requirements but I might have looked at it too fast I apologize.I'll check back when I get homeOn Feb 2, 2022 15:09, Isaac Muse ***@***.***> wrote:
Never mind, I misunderstood. This is required only for MkDocs documentation from what I see. This is not being forced on those installing MkDocs from what I can tell.

pawamoy:
Indeed, setup.py hardcodes the dependencies. The requirements file are not used. This change only impacts MkDocs own docs, nothing else.

ultrabug:
Alright sorry for the noise, I worked so much with those files when adding i18n support that I still thought they were related to the whole project...

oprypin:
I verified by diffing the HTML (before/after) that the output is the same, other than adding `<p class="admonition-title">Block</p>` which I hid with CSS.

PR #2701, opened 2021-12-07
Add support for caching builds between calls to "mkdocs serve"

kitanata:
Hello mkdocs team:

Thank you all for making such a wonderful documentation tool!

I recently joined a large project with extensive multi-language documentation. Compiling this documentation takes roughly 30 minutes each run for a single language. The `--dirtyreload` flag has been helpful in reducing some build redundancy during development, however stopping and starting the development server with the serve command causes the entire documentation to be rebuilt and this is slowing our doc development cycles way down.

mkdocs has hard-coded the use of a temporary directory for the site_dir configuration when the serve command is called. Since the entire temporary directory is new, when build is called by the serve command, then the entire project's docs gets recompiled. For most projects this probably isn't a problem and is even desirable. For our project it has been quite cumbersome and we'd like to optimize this! :D

This PR introduces a `--cached` argument to the serve command. It is turned off by default. But if the developer enables it when calling serve then mkdocs will use a `__mkdocs_cache__` directory in the root of the project instead of the temporary directory that is typically used. The `--cached` flag also disables the deleting of the `__mkdocs_cache__` directory when serve is exited from. This is a feature to allow the build to persist to disk across multiple calls to `mkdocs serve`.

So far I have tested this manually and it seems to work very well without needing to make any other changes besides modifying `__main__` and the `serve.py` file. It's a small change that has a large beneficial impact for some users, and it doesn't change any existing behavior. It's the best kind of PR. :-)

Example of use:

`mkdocs serve --dirtyreload --cached`

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -88,6 +88,7 @@ use_directory_urls_help = "Use directory URLs when building pages (the default).
 reload_help = "Enable the live reloading in the development server (this is the default)"
 no_reload_help = "Disable the live reloading in the development server."
 dirty_reload_help = "Enable the live reloading in the development server, but only re-build files that have changed"
+cached_help = "Use a cached build directory. Only re-build files that have changed."
 commit_message_help = ("A commit message to use when committing to the "
                        "Github Pages remote branch. Commit {sha} and MkDocs {version} are available as expansions")
 remote_branch_help = ("The remote branch to commit to for Github Pages. This "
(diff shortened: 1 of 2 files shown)


kitanata:
^ Please approve the workflow as soon as you can. I'd like to fix up the tests and documentation to get a pretty bow tied on top of this for ya as soon as possible. :)

pawamoy:
Thanks for the PR @kitanata, and for your patience. I'd be interested in such a caching mechanism as well. I could actually reuse it in one of my plugin/extension :thinking: 

Are you still interested in getting this merged @kitanata? We'd need you to rebase the PR, and also some tests I guess. Not sure what to test or how yet, though. I'll try to think about it.

PR #2710, opened 2021-12-16
SubConfig now catches failures and warnings

### LINKED ISSUE #2709: ` SubConfig` ignore configuration failures
While developing [mkdocs-gallery]() I found out that the `SubConfig` class does not retrieve the failures and warnings from the nested options.

This is because `run_validation` executes `Config.validate` but does not look at its results (`failures, warnings`):
https://github.com/mkdocs/mkdocs/blob/97eafb9094fdd56d19022ae8fda332e728df674b/mkdocs/config/config_options.py#L61

smarie:
Fixes #2709

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -65,6 +65,8 @@ class SubConfig(BaseConfigOption, Config):
             key, err = failed[0]
             raise ValidationError(f"Sub-option {key!r} configuration error: {err}")
 
+        return self
+
 
 class ConfigItems(BaseConfigOption):
     """
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -59,8 +59,11 @@ class SubConfig(BaseConfigOption, Config):
         return self.run_validation(value)
 
     def run_validation(self, value):
-        Config.validate(self)
-        return self
+        failed, self.warnings = Config.validate(self)
+        if len(failed) > 0:
+            # get the first failing one
+            key, err = failed[0]
+            raise ValidationError(f"Sub-option {key!r} configuration error: {err}")
 
 
 class ConfigItems(BaseConfigOption):


mkdocs/config/config_options.py
```diff
@@ -59,8 +59,11 @@ def validate(self, value):
         return self.run_validation(value)
 
     def run_validation(self, value):
-        Config.validate(self)
-        return self
+        failed, self.warnings = Config.validate(self)
+        if len(failed) > 0:
+            # get the first failing one
+            key, err = failed[0]
+            raise ValidationError(f"Sub-option {key!r} configuration error: {err}")
```
smarie, line 66:
```suggestion
            raise ValidationError(f"Sub-option {key!r} configuration error: {err}")

        return self
```

oprypin (approved):
Sorry for the delay

smarie:
Great, no worries @oprypin !

oprypin:
Oops, unfortunately this makes a test fail.

https://github.com/mkdocs/mkdocs/runs/5702600210?check_suite_focus=true#step:5:19
```
FAIL: test_extra_context (mkdocs.tests.build_tests.BuildTests)
  File "/home/runner/work/mkdocs/mkdocs/mkdocs/tests/build_tests.py", line 181, in test_extra_context
    cfg = load_config(extra={'a': 1})
  File "/home/runner/work/mkdocs/mkdocs/mkdocs/tests/base.py", line 40, in load_config
    assert(errors_warnings == ([], [])), errors_warnings
AssertionError: ([], [('extra', ('a', 'Unrecognised configuration name: a'))])
```

That makes me think that there are probably unforeseen negative implications from this change, so I'll have to revert it until further investigation.

smarie:
Indeed. The problem with CI not being executed on PRs is that we are not able to re-run it on an updated main HEAD before merging. Maybe something to consider in the future ?

smarie:
Update : I was able to test outside of tox, directly in PyCharm. 
This is what happens : 

 - the `extra` option in the default config schema is defined as an empty `SubConfig()` (without any options defined inside it) https://github.com/mkdocs/mkdocs/blob/17ff566d2a91703396e954efc521d3869472ba79/mkdocs/config/defaults.py#L111-L116
 - in the failing test , the configuration is created with `extra={'a': 1}`, https://github.com/mkdocs/mkdocs/blob/17ff566d2a91703396e954efc521d3869472ba79/mkdocs/tests/build_tests.py#L180-L183

Therefore there was an error in the subconfig `Unrecognised configuration name: a`.
This error was until now, hidden : **it was already there** but was not caught nor shown to the user.

The fix is now doing its job correctly, showing the error to the user.

Now the decision to take with this test is "should the 'extra' config option in the default schema accept any kind of options inside it ?". If the answer is yes, we should probably define a constructor flag in `SubConfig` to either allow for unvalidated options or not. If the answer is no, in that case the test should be modified so that extra is passed an empty dict.

Let me know !

oprypin:
@smarie If I understand correctly, the `extra` option is now widely used as a freeform container of undefined options that themes can access.
Example: https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#versioning
Whatever change you develop, make sure this still works without warnings.

smarie:
@oprypin I opened #2807 that fixes the failing test.

PR #2714, opened 2021-12-18
Don't import not used commands modules

mondeja:
As you can see in the next image, when you run `mkdocs build`, `serve`, `gh_deploy` and `new` modules commands are also imported, although they are not used:

not-used-commands

> Created with `pyinstrument -o out.html -r html -m mkdocs -- build`

These imports add some initialization time to the build because some dependencies like `watchdog` must be initialized. This PR solves it moving the imports to their respective functions.


oprypin (approved):
Nice!

PR #2740, opened 2022-01-04
Move livereload endpoint under the mount path

### LINKED ISSUE #2739: Custom subpath isn't applied for `livereload` feature
If using a `site_url` including a subpath for development (we're using it behind a proxy), this isn't applied when loading the `livereload.js` script, causing the whole auto-reload feature to break.

I expect it to be loaded from `https://test.local/my/subpath/js/livereload.js` if using `https://test.local/my/subpath` as `site_url`.

robertaistleitner:
Fixes #2739.


oprypin:
This is a good approach, something I was thinking towards myself.

I had a somewhat unwieldy style suggestion, so I have implemented it myself here:
https://github.com/mkdocs/mkdocs/compare/robertaistleitner:master...oprypin:robertaistleitner/master

As you didn't give me access to push to your pull request, I suggest you to apply it yourself:
```bash
git fetch origin fa0196138923cb0168cd5d16988947f7506a9fac
git merge fa0196138923cb0168cd5d16988947f7506a9fac
git push
```

robertaistleitner:
I merged your changes, weren't you able to push them yourself? I checked the `Allow edits by maintainers` box - this should be it, right?

oprypin:
I wasn't able to push, although I think I did it right. Maybe it's because you opened the PR from your master branch.

robertaistleitner:
I fixed another thing: the live reload script was injected even if live reloading was disabled.

mkdocs/livereload/__init__.py
```diff
@@ -176,7 +201,7 @@ def _serve_request(self, environ, start_response):
         # https://github.com/bottlepy/bottle/blob/f9b1849db4/bottle.py#L984
         path = environ["PATH_INFO"].encode("latin-1").decode("utf-8", "ignore")
 
-        m = re.fullmatch(r"/livereload/([0-9]+)/[0-9]+", path)
+        m = re.fullmatch(rf"{self.mount_path}livereload/([0-9]+)/[0-9]+", path)
```
oprypin: Ah I missed another aspect. This is a regex after all, so this would at least have to be `re.escape(self.mount_path)`

But it may be better to refactor it further because below we already have code that removes the mount_path prefix

oprypin:
>  @robertaistleitner
> I fixed another thing: the live reload script was injected even if live reloading was disabled

This is really a separate change. Could you revert it here and open another pull request?

oprypin (approved):

robertaistleitner:
Thanks for the test and refactoring!
Is there anything we can do about the failing macos test?

oprypin:
The test fails by itself, don't need to do anything within this PR.

oprypin:
Hm, this change causes quite a big annoyance :(

Steps:

1. `serve` one site (needs to have `site_url` customized)
2. open the site in a browser
3. stop serving it but keep the browser
4. `serve` another site with a different mount point
5. observe spam of "404" errors in the new terminal (also in the old browser but that's whatever)

image

---

As I don't see a nice way to mitigate this, I'm considering reverting the effects of this

PR #2741, opened 2022-01-04
Don't inject javascript poll script if livereload is disabled

robertaistleitner:


oprypin:
I had a less intrusive approach to this in mind.
You could simply detect that `watch` was never called and decide based on that

robertaistleitner:
True that 👍

PR #2751, opened 2022-01-23
Fix imports to get tests passing

cpcloud:
While bumping the version of this package from 1.2.1 to 1.2.3 for nixpkgs
I noticed the tests started failing.

I dug around a bit found that an import was incorrect (`import localization` in `theme_tests.py`)
and `config_options` and `defaults` were missing from `mkdocs/config/__init__.py`.

### DIFF
--- a/mkdocs/config/__init__.py
+++ b/mkdocs/config/__init__.py
@@ -1,4 +1,7 @@
 from mkdocs.config.base import load_config, Config
+from mkdocs.config import config_options, defaults
 
 __all__ = [load_config.__name__,
-           Config.__name__]
+           Config.__name__,
+           config_options.__name__,
+           defaults.__name__]
--- a/mkdocs/config/__init__.py
+++ b/mkdocs/config/__init__.py
@@ -1,7 +1,4 @@
 from mkdocs.config.base import load_config, Config
-from mkdocs.config import config_options, defaults
 
 __all__ = [load_config.__name__,
-           Config.__name__,
-           config_options.__name__,
-           defaults.__name__]
+           Config.__name__]
--- a/mkdocs/tests/base.py
+++ b/mkdocs/tests/base.py
@@ -6,6 +6,7 @@ from tempfile import TemporaryDirectory
 
 from mkdocs import config
 from mkdocs import utils
+from mkdocs.config import defaults as config_defaults
 
 
 def dedent(text):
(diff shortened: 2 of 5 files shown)


oprypin:
Could you explain why the tests are failing for you but not MkDocs' CI system?

> `config_options` and `defaults` were missing from `mkdocs/config/__init__.py`

And what exactly makes these as "missing", maybe they're not supposed to be there? It's just a re-export. What makes you think they're supposed to be there?

cpcloud:
> Could you explain why the tests are failing for you but not MkDocs' CI system?

Sure thing!

The CI is using the `unittest` module incorrectly.

This command:

```sh
python -m unittest discover -p 'theme_tests.py' mkdocs
```

Will treat `mkdocs` as the top level directory, per [the python docs](https://docs.python.org/3/library/unittest.html#cmdoption-unittest-discover-t).

This isn't correct, but it appears to be so, because the current directory is on `sys.path`, but so is `mkdocs`!

> 
> > `config_options` and `defaults` were missing from `mkdocs/config/__init__.py`
> 
> And what exactly makes these as "missing", maybe they're not supposed to be there? It's just a re-export. What makes you think they're supposed to be there?

This is just another consequence of the above.

oprypin:
Ah that's a great explanation! Thanks much.

Now,

> those reexports are used in tests

Could you instead fix the imports in tests rather than adding re-exports?

cpcloud:
> Ah that's a great explanation! Thanks much.
> 
> Now,
> 
> > those reexports are used in tests
> 
> Could you instead fix the imports in tests rather than adding re-exports?

Roger that, on it.

cpcloud:
@oprypin Done!

oprypin (approved):

cpcloud:
@oprypin Can this be merged?

oprypin:
OK.  :smile: 
Thanks!

oprypin:
Umm by the way, I don't know why you really needed this merge. Just in case: please don't use the current state of repository as the version 1.2.3, it has breaking changes compared to that now.

cpcloud:
The plan is to take just these commits as patches on top of 1.2.3, if possible.

cpcloud:
The merge wasn't necessary for what I need to do, I just don't like having outstanding pull requests around if it's not necessary.

PR #2755, opened 2022-01-25
Fix PyPy version in MacOS CI workflow

mondeja:
Github Actions now uses MacOS 11 as `macos-latest` and this version [does not includes PyPy 3.6](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#pypy). There are two options to fix this problem:

- a) Use `macos-10.15` instead of `macos-latest` in the CI. This allows to continue the testing against PyPy 3.6 (see [tests passing](https://github.com/mondeja/mkdocs/runs/4942876969?check_suite_focus=true)).
- b) Use PyPy 3.7, included in latest MacOS (see [tests passing](https://github.com/mondeja/mkdocs/actions/runs/1747596121)).

I've opted for option *b* because Python3.6 EOL has been reached some months ago and because the CI is currently using latest OS builds for Windows and Linux.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -8,8 +8,8 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


oprypin (approved):
Thanks!

PR #2777, opened 2022-02-23
Allow to unwatch files with `LiveReloadServer`

mondeja:
I'm working on a project that needs to watch files being served based in the content on other files. I've not found a way to stop watching files for changes. As far as I know, I can't access the watch object returned by `observer.schedule` needed to unwatch them, so crafted this `unwatch` method for the server.

Is working as expected in my local environment (Ubuntu 20.04). Mentioning @oprypin as creator and maintainer of the new livereload server.


oprypin:
Hmm by looking at the dependent pull request, I finally understand the problem that this tries to solve.
One might want to pick up new files to watch throughout the `serve`, and subsequently "un"-pick them up if they are no longer relevant.
Maybe there should be a better solution, such as building up the list of paths to watch from scratch every time (and calling `on_serve` every time).
But hm, this certainly can work, without many unnecessary changes.

To not give false hope that this can skip sub-paths of existing directories (the method doesn't even error in case it finds no match!), I think we should instead directly expose the result of `observer.schedule` and then operate based on it.

oprypin:
I looked into it more, and actually found it a bad experience to use those watch references directly. And it didn't even solve what I was trying to solve.

The main problem I was trying to avoid is for a plugin to *accidentally* unwatch paths that it didn't watch in the first place.
I solved it with a counter instead.
I pushed my changes directly. Sorry if you already started work.

oprypin:
Try it out if it works for you

mondeja:
It works, thank you! 👍🏼

PR #2807, opened 2022-03-27
SubConfig: allow to actually validate the nested configs

smarie:
Fixes https://github.com/mkdocs/mkdocs/issues/2709

This is a fix for previous PR (reverted) #2710

`SubConfig` now has an optional constructor argument named `ignore_validation`. By default it is `True` to preserve retrocompatibility with legacy. Setting it to `False` will now correctly collect and raise validation errors and warnings for subconfig items.


oprypin:
Thank you very much for the contribution. Finally I come back to ConfigOptions. And this makes perfect sense, particularly together with #2916 and #2938. I merged master (was needed after #2916) and added some of my own changes.

oprypin (approved):

smarie:
Thanks @oprypin !

Issue #2700, opened 2021-12-07
Links to sites using MkDocs


cpach:
Have you considered adding some links from the web pages to open source projects that use MkDocs?

That could be very useful for prospective new users who would like to get a feeling for how MkDocs works in practice.

CannonballSkippy:
I was just looking for the same.

squidfunk:
Here are some using Material for MkDocs 😉
https://github.com/squidfunk/mkdocs-material#trusted-by-

pawamoy:
There's this wiki page: https://github.com/mkdocs/mkdocs/wiki/MkDocs-Users

squidfunk:
Actually, I think that MkDocs could do some more marketing, so adding some prominent users to the website or README wouldn't hurt. It would show that the software is mature and trusted by many users and megacorps, which can be an important sign for new adopters. IMHO the GitHub wiki, albeit advertised on the homepage, is more or less within a blind spot. For example, Docusaurus has much better marketing materials: https://docusaurus.io/showcase

timvink:
> Actually, I think that MkDocs could do some more marketing.

I agree. That also extends to plugins. We now have a large amount of high quality, mature plugnis, but on the [plugins page](https://www.mkdocs.org/user-guide/configuration/#plugins) there is no link to the [wiki](https://www.mkdocs.org/dev-guide/plugins/). And there is no mention of plugins in the [getting started](https://www.mkdocs.org/user-guide/configuration/) guide.

I might be biased, but I think mkdocs + mkdocs-material + a subset of plugins is what makes the mkdocs ecosystem so powerful. I would love for more people to discover it :)

squidfunk:
Yup, absolutely. The plugin system is just awesome. The docs for Material for MkDocs include guidance for some plugins and Markdown extensions, mostly those that are (and need to be) natively supported. Listing them on the MkDocs page and having a curated list would lower entry barriers.

Also, and this is my personal opinion, redoing (or changing) the default theme to something more friendly might be something to consider. Of course, I'm biased because I do so much work on Material or MkDocs, which, for some users, is the only thing they know about MkDocs. However, from the eyes of a first-time user, the default theme might be off-putting as it looks pretty dated and has only very rudimentary capabilities.

Just my 2 cents.

timvink:
> redoing (or changing) the default theme to something more friendly might be something to consider

[...]

PR #2824, opened 2022-04-17
Add a link to best-of-mkdocs in the readme

pawamoy:
Following the advice of @squidfunk, I'm opening this PR to get feedback from MkDocs maintainers and open a discussion :slightly_smiling_face: 

This PR adds a link to the [best-of-mkdocs](https://github.com/pawamoy/best-of-mkdocs) list I've just created to the README.

Providing such curated data / galleries of examples in MkDocs own docs has been discussed previously in #2700.
I wouldn't mind transferring the best-of list to the MkDocs org and keep maintaining it, if that's something you'd like (it's pretty low maintenance as the only file to update is the YAML data, the rest is automatically rendered through a GitHub action).

One of the benefits of the best-of list is that it logs useful information during the rendering process. For example, I was able to fix several plugins URLs as they had been renamed or transferred to other GitHub accounts.

Another small benefit is that with a restricted access (PRs, contrary to the wiki), we won't have to remove spam contents added from time to time by users, or revert accidental/malicious deletions. And it's still very easy for the community to contribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Make sure to stick around to answer some questions as well!
 - [Official Documentation][mkdocs]
 - [Latest Release Notes][release-notes]
 - [MkDocs Wiki][wiki] (Third-party themes, recipes, plugins and more)
+- [Best-of-MkDocs][best-of] (Curated list of themes, plugins and more)
 
 ## Contributing to MkDocs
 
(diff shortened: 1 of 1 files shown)


oprypin:
Sorry for the total silence from me. I have finally gotten around to this.
For plugins this is definitely great.
Themes might be somewhat underrepresented, there's no image for them or anything.
Still though, certainly better than the mess in the wiki.
For now let's merge this addition of the link, and yes- perhaps this really should totally replace the Plugins page.
Not sure how to get that started, it's so hard for me to make any decision :sweat: - anyway if someone wants to open a new sub-thread at https://github.com/mkdocs/mkdocs/discussions/2823 , I'll be reading that :D

oprypin (approved):

pawamoy:
That's great, thank you @oprypin!

PR #2825, opened 2022-04-20
Clean up some code

Sea-n:
- Use `GitHub` in correct case
- Change `{% if foo%}` to `{% if foo %}`
- Convert tabs to spaces
- Change a few 2 or 3 spaces indent to 4 spaces


oprypin (approved):
Thanks! This is good.
I'll leave this for comments just in case. And not sure when I'll merge this, sorry.

Sea-n:
Sure, no rush 😊

PR #2894, opened 2022-07-18
When a Markdown extension fails to load, print its name and traceback

oprypin:


oprypin:
Example output:
```
ERROR    -  Config value: 'markdown_extensions'. Error: Failed to load extension 'mdx_gh_links'.
              File "/home/blaxpirit/repos/mkdocs/venv/lib/python3.10/site-packages/mdx_gh_links.py", line 35, in <module>
                from markdown.util import etree
            ImportError: cannot import name 'etree' from 'markdown.util' (/home/blaxpirit/.local/lib/py/markdown/util.py)
Aborted with 1 Configuration Errors!
```

ultrabug (approved):

ultrabug:
Thanks, the PR is good as-is

I was just wondering if we should add a mention that makes it explicit that mkdocs is not responsible for the breakage, like `Please check the traceback below to find out which extension module has a problem and report to them.`

oprypin:
Actually we don't know that that's true - that the extension is at fault. This could also be coming from configuring the extension incorrectly.

```yaml
extensions:
  - pymdownx.superfences:
      custom_fences: foo
```
```
ERROR    -  Config value: 'markdown_extensions'. Error: Failed to load extension 'pymdownx.superfences'.
              File "/home/blaxpirit/repos/mkdocs/mkdocs/config/config_options.py", line 641, in run_validation
                md.registerExtensions((ext,), self.configdata)
              File "/home/blaxpirit/repos/mkdocs/venv/lib/python3.10/site-packages/markdown/core.py", line 125, in registerExtensions
                ext.extendMarkdown(self)
              File "/home/blaxpirit/repos/mkdocs/venv/lib/python3.10/site-packages/pymdownx/superfences.py", line 273, in extendMarkdown
                name = custom.get('name')
            AttributeError: 'str' object has no attribute 'get'
```

Previously it was just this, btw:
```
ERROR    -  Config value: 'markdown_extensions'. Error: 'str' object has no attribute 'get'
```

Determining that this is for sure an extension's fault could be more complex or not even feasible.

Issue #2892, opened 2022-07-18
Support latest realise of Markdown library


Samreay:
I believe there has been some update to the `Markdown` library and how it internally records its version that is breaking things.

With a brand new environment and a fresh install of `mkdocs`, a `mkdocs build --strict --verbose` fails my project with this error:

```bash
DEBUG    -  Loading configuration file: /Users/sh/Projects/dataportalapiclient/mkdocs.yml
ERROR    -  Config value: 'markdown_extensions'. Error: module 'markdown' has no attribute 'version_info'
```
At this point, mkdocs has a dependency on `Markdown==3.4.1`, which was released [three days ago](https://github.com/Python-Markdown/markdown/tags). 

After running a `pip install Markdown==3.3.7` to downgrade the version, rerunning the build is successful:

```bash
DEBUG    -  Loading configuration file: /Users/sh/Projects/dataportalapiclient/mkdocs.yml
...
DEBUG    -  mkdocstrings: Tearing handlers down
INFO     -  Documentation built in 3.45 seconds
```

I notice in [this commit from May 27th on the Markdown repository](https://github.com/Python-Markdown/markdown/commit/a767b2daaad78ba32d45a4f1dabb7c5e218f030a), the deprecated `version_info` info object was removed, and replaced with the `__version_info__` object, as per this table:

| Deprecated Object                      | Replacement Object                  |
|----------------------------------------|-------------------------------------|
| `markdown.version`                     | `markdown.__version__`              |
| `markdown.version_info`                | `markdown.__version_info__`         |
| `markdown.util.etree`                  | `xml.etree.ElementTree`             |
| `markdown.util.string_type`            | `str`                               |
| `markdown.util.text_type`              | `str`                               |
| `markdown.util.int2str`                | `chr`                               |
| `markdown.util.iterrange`              | `range`                             |
| `markdown.util.isBlockLevel`           | `markdown.Markdown.is_block_level`  |
| `markdown.util.Processor().markdown`   | `markdown.util.Processor().md`      |
| `markdown.util.Registry().__setitem__` | `markdown.util.Registry().register` |
| `markdown.util.Registry().__delitem__` |`markdown.util.Registry().deregister`|
| `markdown.util.Registry().add`         | `markdown.util.Registry().register` |

[...]

PR #2901, opened 2022-07-25
config: try to detect extensions breaking with markdown >= 3.4.1

ultrabug:
see #2892 #2894 and Python-Markdown#1277

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import sys
 import traceback
@@ -9,6 +10,16 @@ import markdown
 
 from mkdocs import utils, theme, plugins
 from mkdocs.config.base import Config, ValidationError
+
+try:
+    import pkg_resources
+
+    markdown_dist = pkg_resources.get_distribution("markdown")
+    markdown_version = markdown_dist.version
+except Exception:
+    markdown_version = None
+
+log = logging.getLogger("mkdocs.config")
 
 
 class BaseConfigOption:
@@ -634,6 +645,24 @@ class MarkdownExtensions(OptionallyRequired):
 
         extensions = utils.reduce_list(self.builtins + extensions)
 
+        # List some of the Python-Markdown 3.4 deprecations
+        # https://python-markdown.github.io/change_log/release-3.4/
+        # see mkdocs#2892 and Python-Markdown#1277
+        md_34_deprecations = [
+                "Processor",
+                "Registry",
+                "etree",
+                "extendMarkdown",
+                "int2str",
+                "isBlockLevel",
+                "iterrange",
+                "markdown.util",
+                "markdown.version",
+                "markdown.version_info",
+                "md_globals",
+                "string_type",
+                "text_type",
+        ]
         # Confirm that Markdown considers extensions to be valid
         md = markdown.Markdown()
         for ext in extensions:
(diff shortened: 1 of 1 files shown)


ultrabug:
@oprypin I understand and agree that this is not ideal but I'd like your point of view on this approach.

mkdocs/config/config_options.py
```diff
@@ -647,6 +676,13 @@ def run_validation(self, value):
                     stack.insert(0, frame)
                 tb = ''.join(traceback.format_list(stack))
 
+                if markdown_version >= "3.4":
+                    if isinstance(e, ImportError) or isinstance(e, TypeError):
+                        for word in tb.split():
```
oprypin, line 681: `split` isn't robust against punctuation around words, so this probably could be done better without using it. Maybe a `\b(foo|bar)\b` regex.

Actually... seeing as almost all error messages contain quotes, this all code will not even detect this error, as far as I can tell.

```
ImportError: cannot import name 'etree' from 'markdown.util'
```

mkdocs/config/config_options.py
```diff
@@ -647,6 +676,13 @@ def run_validation(self, value):
                     stack.insert(0, frame)
                 tb = ''.join(traceback.format_list(stack))
 
+                if markdown_version >= "3.4":
```
oprypin, line 679: The value of `markdown_version` can be `None` which is not comparable.

It could also be these:
```python
"3.10" <= "3.4"
"10.0" <= "3.4"
```
-which is not correct

mkdocs/config/config_options.py
```diff
@@ -634,6 +645,24 @@ def run_validation(self, value):
...
+                "Processor",
+                "Registry",
+                "etree",
+                "extendMarkdown",
+                "int2str",
+                "isBlockLevel",
+                "iterrange",
+                "markdown.util",
+                "markdown.version",
+                "markdown.version_info",
```
oprypin, line 661: Never detects this error:
```python
>>> from markdown import version_info
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'version_info' from 'markdown'
```

mkdocs/config/config_options.py
```diff
@@ -647,6 +676,13 @@ def run_validation(self, value):
                     stack.insert(0, frame)
                 tb = ''.join(traceback.format_list(stack))
 
+                if markdown_version >= "3.4":
+                    if isinstance(e, ImportError) or isinstance(e, TypeError):
```
oprypin, line 680: Never detects this error:
```python
>>> import markdown.util
>>> markdown.util.string_type
  File "<stdin>", line 1, in <module>
AttributeError: module 'markdown.util' has no attribute 'string_type'
```

oprypin (commented):
I am against this approach :(
MkDocs can't become a collection of guesses about error messages. (Jinja2 could certainly also fall under such a circumstance too.) If the used package isn't willing to keep some old aliases, how can we be expected to maintain an equivalent set of checks for them?
There are way better things we could be doing instead of maintaining this.

This also isn't a significant upgrade compared to the general error message implementation, because it's not going to actually fix the problem.

And it's really error-prone to try to understand errors.
I added review comments regarding the implementation, but addressing them will not change my opinion on the idea.

oprypin:
+The funny thing is, it wouldn't even be much of a stretch from this  to *actually fixing* the errors just by monkeypatching the library. But I'm not suggesting that at all, either.

ultrabug:
We merged #2893 for users' sake but I'm not giving up on dropping it as fast as possible.

I do not want to leave us in the stage where we keep that Markdown version pin for too long so I'm trying to find something *reasonable* and **temporary** that we can remove in say, 1 year max.

Maybe just checking the Markdown version and adding to the current error message would be enough? How would you feel about that?

>  I added review comments regarding the implementation, but addressing them will not change my opinion on the idea.

Sorry I should have been clearer that this PR was mostly about the approach.

> +The funny thing is, it wouldn't even be much of a stretch from this to actually fixing the errors just by monkeypatching the library. But I'm not suggesting that at all, either.

I'm not either, not one bit.

oprypin:
I wrote a new comment here:
https://github.com/mkdocs/mkdocs/issues/2892#issuecomment-1200518008

Unlocking the version and then writing a warning that says "oh you probably got the broken version" still doesn't make sense to me.

ultrabug:
Ok, let's do that then.

PR #2913, opened 2022-08-06
Re-format code with `black -l100` and enforce it

oprypin:
- Remove spaces at the ends of docstrings, normalize quotes
- Some manual changes ahead of formatting code with Black
- Format code with `black -l100 --skip-string-normalization`
- Enforce Black formatter

Closes #2398 by @ultrabug


oprypin:
Note: please review this commit-by-commit

* https://github.com/oprypin/mkdocs/commit/black~3
* https://github.com/oprypin/mkdocs/commit/black~2?w=1
* https://github.com/oprypin/mkdocs/commit/black~?w=1
* https://github.com/oprypin/mkdocs/commit/black

ultrabug (commented):
Could you provide your rationale for sticking with py36 and skipping string normalization for reference please?

It sure does make the diffs smaller.

oprypin:
py36 is what we have so I choose that. That can be dropped separately. Anyway it probably makes no change on the behavior.

String normalization isn't such a big win. And it was a huge sticking point in the previous discussion.
And indeed it would be a huge diff and a lot of noise if combined with this PR. We can always follow up on that part separately.

oprypin:
I have personally looked at the whole diff at least twice and it seems good to me.

I plan to merge this today.

oprypin:
-or raise your concerns and then I won't merge :)

ultrabug (approved):
I'm fine, nothing struck me and tests are OK

PR #2916, opened 2022-08-07
Rework SubConfig so it doesn't share mutable state

### LINKED ISSUE #2909: ConfigItems overwrites all data with last entry in the list of config items
`ConfigItems` does not work as expected and results in a list of items that contains only the same values.

### Reproduciton

`plugin.py`
```python
class TestPlugin(BasePlugin):
    config_scheme = (
        (
            "items", mkdocs.config.config_options.ConfigItems(
                ("value", mkdocs.config.config_options.Type(str)),
            ),
        ),
    )

    def on_config(self, config):
        print(self.config)

```

`mkdocs.yml`
```yaml
site_name: Test Site

plugins:
    - test:
        items:
            - value: a
            - value: b
```

This results in:
```
{'items': [{'value': 'b'}, {'value': 'b'}]}
INFO     -  Documentation built in 0.05 seconds
INFO     -  [12:21:39] Watching paths for changes: 'docs', 'mkdocs.yml'
INFO     -  [12:21:39] Serving on http://127.0.0.1:8000/
```

Note that both "value" values are `b` rather than the first one being `a`.

### Cause

`ConfigItems` creates a single `SubConfig` instance when it is instantiated:
https://github.com/mkdocs/mkdocs/blob/a17a25c60eb6622556bd74735fab84403f319dd5/mkdocs/config/config_options.py#L77

It then uses this instance to do validation:
https://github.com/mkdocs/mkdocs/blob/a17a25c60eb6622556bd74735fab84403f319dd5/mkdocs/config/config_options.py#L94

The `SubConfig` instance uses `load_dict` to populate `self.data` before validation.
https://github.com/mkdocs/mkdocs/blob/a17a25c60eb6622556bd74735fab84403f319dd5/mkdocs/config/config_options.py#L60

The `load_dict` function uses `update` to make changes to `self.data` on the `SubConfig` instance:
https://github.com/mkdocs/mkdocs/blob/a17a25c60eb6622556bd74735fab84403f319dd5/mkdocs/config/base.py#L132

Since there is only one `SubConfig` instance, when this dictionary gets updated it gets updated across all references to it.

### Fix

A number of popular plugins already create their own `ConfigItems` to address this by making `item_config` a property that returns a separate `SubConfig` instance each time it is used. This seems like a reasonable approach.

https://github.com/pieterdavid/mkdocs-doxygen-plugin/blob/49b841b649acf58b668c0b805ab27f9e454920f4/mkdocsdoxygen/configitems.py#L24-L26

oprypin:
* Fixes #2909

 @pjbull  let me know if this works for you, or if you're not willing to try it out, also let me know :D


pjbull:
@oprypin I tested both before and after, and this fixes what I saw with #2909. Thanks!

ultrabug (approved):
nice catch on the tests

PR #2917, opened 2022-08-07
CI: stop checking diffs in translation files

oprypin:
This check has only caused a lot of spurious failures. I never check it. In my view this approach is still a non-starter, and my comments had predicted the problems quite spot-on: https://github.com/mkdocs/mkdocs/pull/2299#issuecomment-804362611

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -86,31 +86,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Python ${{ matrix.python-version }}
+      - name: Setup Python
         uses: actions/setup-python@v2
         with:
           python-version: 3.7
       - name: Install Python dependencies
         run: |
-          python -m pip install -r requirements/project.txt
-      - name: Check if Portable Object Templates should have been updated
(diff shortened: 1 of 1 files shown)


.github/workflows/tox.yml
```diff
@@ -86,31 +86,14 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Python ${{ matrix.python-version }}
+      - name: Setup Python
         uses: actions/setup-python@v2
         with:
           python-version: 3.7
       - name: Install Python dependencies
         run: |
-          python -m pip install -r requirements/project.txt
```
ultrabug, line 95: doing this to be faster I guess?

ultrabug (approved):
We knew this was not optimal and I agree with you that it was not useful enough.

.github/workflows/tox.yml
oprypin: Actually I'm more closely following deploy-release.yml.  This way we can be more certain that that workflow won't fail because this will continuously check it in a very similar way

ultrabug: Ok, thanks

PR #2927, opened 2022-08-13
Add `edit_uri_template` config

### LINKED ISSUE #1378: Improve edit_uri flexability

oprypin:
Resolves #1378

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -186,6 +186,7 @@ class Page:
                 noext = posixpath.splitext(src_uri)[0]
                 edit_uri = edit_uri_template.format(path=src_uri, path_noext=noext)
             else:
+                assert edit_uri is not None and edit_uri.endswith('/')
                 edit_uri += src_uri
             if repo_url:
                 # Ensure urljoin behavior is correct
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -41,7 +41,9 @@ class Page:
         self.update_date = get_build_date()
 
         self._set_canonical_url(config.get('site_url', None))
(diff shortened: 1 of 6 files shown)


ultrabug (commented):
I guess we're missing some docstring updates and maybe some documentation about this added feature please

mkdocs/structure/pages.py
```diff
@@ -172,10 +174,20 @@ def _set_canonical_url(self, base: Optional[str]) -> None:
...
+        edit_uri_template: Optional[str] = None,
+    ) -> None:
+        if edit_uri or edit_uri_template:
             src_uri = self.file.src_uri
-            edit_uri += src_uri
+            if edit_uri_template:
+                noext = posixpath.splitext(src_uri)[0]
+                edit_uri = edit_uri_template.format(path=src_uri, path_noext=noext)
+            else:
+                assert edit_uri is not None and edit_uri.endswith('/')
```
ultrabug, line 189: I usually don't like assertions like this in the code but since it's used some other places in this file, I guess I can live with it

ultrabug (approved):

mkdocs/structure/pages.py
oprypin: Assertions like what 😅
Also other assertions in this file are probably recently added by me 😅
Anyway my approach towards Python assert is "it can actually literally never happen"; this is not error checking. In this case it's to satisfy the type checker, it looks at assertions

PR #2937, opened 2022-08-16
Revert to the explicit way of registering plugins' methods

oprypin:
See prior: https://github.com/mkdocs/mkdocs/commit/f79b34d174e41084391868e7b503f5c61b8b1bdf#diff-da2086192f0b49a8ae9a67eba6ec5e7572aaa0c0db458b1d9deea21005fc7655L84

But now we need to also skip non-overridden methods. That's also the reason why I needed to use `staticmethod`, so these are not concealed by `bound method` functionality.


ultrabug (approved):
I like that we get staticmethods now

PR #2959, opened 2022-09-01
Only depend on `importlib-metadata` under Python 3.10

### LINKED ISSUE #2958: No python version marker on importlib-metadata
### Discussed in https://github.com/mkdocs/mkdocs/discussions/2956

<div type='discussions-op-text'>

<sup>Originally posted by **pawamoy** August 31, 2022</sup>
The `importlib-metadata` dependency was bumped to `>=4.3` in https://github.com/mkdocs/mkdocs/commit/ec9b583b02376b909c10889fef780cdf2f25a23d.

We're now getting resolution conflicts with Flake8 because they set an upper bound on `importlib-metadata`: `<4.3`. From their answers in different issues and PRs (starting point here: https://github.com/PyCQA/flake8/issues/1503), it seems there has been quite some friction regarding this, and they likely won't change their upper bound.

Back to MkDocs: shouldn't the dependency have a marker like `python_version < '3.8'`?

Actual conflict we get (PDM output, URLs stripped):

```
Conflicts detected: 
  importlib-metadata>=4.3 (from <Candidate mkdocs 1.3.1)
  importlib-metadata; python_version < "3.8" (from <Candidate flake8-comprehensions 3.10.0)
  importlib-metadata; python_version < "3.8" (from <Candidate flake8-tidy-imports 4.8.0)
  importlib-metadata>=0.12; python_version < "3.8" (from <Candidate pytest 7.1.2)
  importlib-metadata>=3.6.0; python_version < "3.10" (from <Candidate pytest-randomly 3.12.0)
  importlib-metadata; python_version < "3.8" (from <Candidate wps-light 0.16.1)
  importlib-metadata<4.3,>=1.1.0; python_version < "3.8" (from <Candidate flake8 5.0.4)
```

We're using Python 3.8 so I believe adding the marker would help in our case at least. Unless the right marker to use is `python_version < '3.10'`.</div>

pawamoy:
Fixes #2958 

We always try to import `importlib_metadata` first.
I can switch to a condition on the current Python version, instead of a try/except, if you think that's more robust.

Without the `type: ignore` comments I was getting this:

```
mkdocs/utils/__init__.py:30: error: Incompatible import of "entry_points" (imported name has type "Callable[[], Dict[str, List[EntryPoint]]]", local name has type "Callable[[KwArg(Any)], Union[EntryPoints, SelectableGroups]]")  [misc]
mkdocs/plugins.py:22: error: Incompatible import of "entry_points" (imported name has type "Callable[[], Dict[str, List[EntryPoint]]]", local name has type "Callable[[KwArg(Any)], Union[EntryPoints, SelectableGroups]]")  [misc]
Found 2 errors in 2 files (checked 58 source files)
```

EDIT: might need to squash to reword the commit: *below* instead of *under* (pardon my french).


oprypin:
Thanks for the research.

Usually the try-except imports apply nicely to stdlib modules that have 3rd party shims and you want to prioritize stdlib if it exists. And you do **have to** try stdlib first. The reason is that otherwise the behavior will be affected by the external state of whether this package is installed: Confusing error if it's not installed when it should be  and just arbitrarily using the (hopefully) equivalent implementation when it happens to be installed but was not requested.

Here, the stdlib module existing is not enough, so that approach can't be used. So I think we indeed have to use an explicit version check. 

```python
if sys.version_info >= (3, 10):
    from importlib.metadata import entry_points
else:
    from importlib_metadata import entry_points  # type: ignore
```

pawamoy:
Thanks, I've applied your suggestion.

mkdocs/plugins.py
```diff
@@ -5,10 +5,10 @@
 from __future__ import annotations
 
 import logging
+import sys
 from collections import OrderedDict
 from typing import Any, Callable, Dict, Optional, Sequence, Tuple, TypeVar, overload
 
-import importlib_metadata
```
oprypin, line 11: I just had a nitpick that I like to still keep the order of import statements - this is kinda between stdlib and third party, so: https://github.com/oprypin/mkdocs/commit/55da0f892a0eeff731b849486f751c0596ca0506

oprypin (approved):
Thanks, this is good.

And yes, I'm seeing (3.10 locally and 3.9 in CI) that now apparently `# type: ignore` is not needed.

mkdocs/plugins.py
pawamoy: Sure, I put it below because that's usually how isort likes it. Suggestion applied :slightly_smiling_face:

Issue #2091, opened 2020-05-02
Create and use local plugins


aklajnert:
As the [documentation](https://www.mkdocs.org/user-guide/plugins/) mentions, the plugin has to be a separate Python package which has to be registered in the `entry_points`.  

In my case, I have a simple script that does some pre-build operations on the documentation sources. The only way to run it automatically, for now - is to create another package that would register itself as the plugin. 

How about allowing to load plugins that are just local files? E.g. `- docs.custom_plugin::MyPluginClass` would load `MyPluginClass` from the `docs/custom_plugin.py`. Of course, the `MyPluginClass` should follow the plugin API by inheriting the `BasePlugin` class.

waylan:
If we were to support such a feature, it would need to be in a separate directory from the documentation (just like a theme's `custom_dir`). However, it is not much work to make a local package, which ensures the code is on the PYTHONPATH. 

Personally, my vote is to not do this. It is much cleaner with all Python code being installed.

aklajnert:
Since the directory containing `mkdocs.yml` is already in `PYTHONPATH`, it should be fairly easy to import local plugins by referencing its path. I don't understand why it has to be a separate directory, I think it can be wherever the user want it.  

I know that creating an installable package is not much work, but it feels a bit unnatural. If you work on a library, you can't add plugin entry point to the `setup.py` of the package to avoid registering (and possibly shadowing) internal plugin in people's environment.

aklajnert:
For now, I've created a generic plugin that allows calling any code from within the project as a hook. In case anybody needs it, you can find it here: https://pypi.org/project/mkdocs-simple-hooks/

waylan:
Awesome. I love this being a plugin. Be sure to add it to the [list on the wiki][1] (if you haven't already).

[1]: https://github.com/mkdocs/mkdocs/wiki/MkDocs-Plugins

GeneralZero:
For those that want to do a plugin locally here are the steps to get it to work. First make sure you add the folder with the extension into the PYTHONPATH. I did this through the following command. ```export PYTHONPATH=`pwd` ```

**mkdocs.yml:**
```yml
markdown_extensions:
[...]
  - common_fix:CommonFixExtention
```

**common_fix.py:**
```python
from markdown.preprocessors import Preprocessor
from markdown.extensions import Extension
import re

class CommonFixExtention(Extension):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)

[...]

PR #2978, opened 2022-09-15
Hooks feature: run plugin events from local Python files

oprypin:
Add a new `hooks` config to `mkdocs.yml`:

A list of paths to Python scripts (relative to `mkdocs.yml`) that are loaded and used as [plugin](https://www.mkdocs.org/dev-guide/plugins/) instances.

For example:

```yaml
hooks:
    - my_hooks.py
```

Then the file *my_hooks.py* can contain any [plugin event handlers](https://www.mkdocs.org/dev-guide/plugins/#events) (without `self`), e.g.:

```python
def on_page_markdown(markdown, **kwargs):
    return markdown.replace('a', 'z')
```

This does not enable any new abilities compared to plugins, it only simplifies one-off usages, as these don't need to be *installed* like plugins do.

Note that for `mkdocs serve` the hook module will *not* be reloaded on each build.

You might have seen this feature in the [mkdocs-simple-hooks plugin](https://github.com/aklajnert/mkdocs-simple-hooks) by @aklajnert. If using standard method names, it can be directly replaced, e.g.:

```diff
-plugins:
-  - mkdocs-simple-hooks:
-      hooks:
-        on_page_markdown: 'my_hooks:on_page_markdown'
+hooks:
+  - my_hooks.py
```

### DIFF
--- a/docs/css/extra.css
+++ b/docs/css/extra.css
@@ -23,6 +23,11 @@ code.no-highlight {
 div.admonition.block>.admonition-title {
     display: none;
 }
+div.admonition.new {
+    color: #15654a;
+    background-color: #e4f7f1;
+    border-color: #bcf1e8;
+}
 
 /* Definition List styles */
 
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -577,6 +577,41 @@ This alternative syntax is required if you intend to override some options via
 
 **default**: `[]` (an empty list).
 
+### hooks
+
+NEW: **New in version 1.4.**
+
+A list of paths to Python scripts (relative to `mkdocs.yml`) that are loaded and used as [plugin](#plugins) instances.
+
+For example:
+
+```yaml
+hooks:
+    - my_hooks.py
+```
+
+Then the file *my_hooks.py* can contain any [plugin event handlers](../dev-guide/plugins.md#events) (without `self`), e.g.:
+
+```python
+def on_page_markdown(markdown, **kwargs):
+    return markdown.replace('a', 'z')
+```
+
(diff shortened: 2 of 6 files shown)


ultrabug:
> You might have seen this feature in the [mkdocs-simple-hooks plugin](https://github.com/aklajnert/mkdocs-simple-hooks) by @aklajnert. If using standard method names, it can be directly replaced, e.g.:

Maybe as a courtesy for @aklajnert you could share your rationale in implementing this straight in MkDocs (other than not having to *install* something)?

Some users may argue that we often push back such ideas to the responsibility of plugins, so I'd like it to be clear here.

Thanks

oprypin:
Sure, you have a point there.

I think it's been too many times that I wanted to suggest a solution to someone in Discussions which was just to throw a few lines of code into a file, at which point installing that plugin constituted more than half of the remaining complexity.

And this plays better with (mainly new) MkDocs features: priorities also work, persisting state within `mkdocs serve` works, paths are handled correctly with `mkdocs build -f foo/mkdocs.yml`.

And pushing back to plugins is something that I want to reduce, particularly when a feature is unlikely to invite "configuration creep", that is, when people would each start to request a tiny tweak to the configuration for their use case.
Here I expect to require no configuration.

aklajnert:
FYI, initially I asked to put it into the core (see #2091), but I decided it will be faster, and easier to implement as a plugin.

ultrabug:
> Sure, you have a point there.

And you have a point too, thanks for clarifying that for everyone.

sreenivasulureddysura:
@oprypin this feature is not working mkdoc verison 1.4.2

ultrabug:
@sreenivasulureddysura please fill a complete bug report with a clear procedure to reproduce the problem you are facing.

Dropping a comment on a closed PR is not helpful I'm afraid.

Thanks

PR #2988, opened 2022-10-02
Switch to pyproject.toml and Hatch build

oprypin:
<del>

Biggest remaining issues:

* I can't actually remove `setup.py` because we have commands that require it:
  E.g. `python setup.py compile_catalog -t mkdocs`

    * I can't set the license because during the above command, `setuptools` says:
        ```
        configuration error: `project.license` must be valid exactly by one definition (2 matches found):

            - keys:
                'file': {type: string}
              required: ['file']
            - keys:
                'text': {type: string}
              required: ['text']
        ```

</del>


facelessuser:
> * I can't actually remove `setup.py` because we have commands that require it:
>   E.g. `python setup.py compile_catalog -t mkdocs`

What about something like this? https://github.com/facelessuser/Rummage/blob/master/pyproject.toml#L98

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+[project.entry-points."mkdocs.plugins"]
+search = "mkdocs.contrib.search:SearchPlugin"
+
+[tool.hatch.version]
+path = "mkdocs/__init__.py"
+
+[tool.hatch.build]
+artifacts = ["/mkdocs/**/*.mo"]
+[tool.hatch.build.targets.sdist]
+include = ["/mkdocs", "/LICENSE", "/pyproject.toml", "/README.md"]
```
ofek, line 76: https://hatch.pypa.io/latest/plugins/builder/sdist/#default-file-selection

```suggestion
include = ["/mkdocs"]
```

pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Environment :: Console",
+    "Environment :: Web Environment",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: BSD License",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.6",
```
mondeja, line 22: Python3.6 support was removed in #2988

```suggestion
```

.github/workflows/deploy-release.yml
```diff
@@ -14,20 +14,20 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip setuptools wheel babel
+          python -m pip install -U wheel hatch babel
```
ofek, line 20: I'd recommend

```suggestion
          python -m pip install -U build babel
```

.github/workflows/deploy-release.yml
```diff
@@ -14,20 +14,20 @@ jobs:
...
       - name: Compile localization message files
         run: |
           pybabel compile --statistics --directory mkdocs/themes/mkdocs/locales
           pybabel compile --statistics --directory mkdocs/themes/readthedocs/locales
       - name: Build
         run: |
-          python setup.py bdist_wheel sdist --formats gztar
+          hatch build
```
ofek, line 27:
```suggestion
          python -m build
```

.github/workflows/deploy-release.yml
facelessuser: I agree. Using `build` is the best way forward.

oprypin: But hold on, what about the `hatch publish` part?

facelessuser: That's a good question. I haven't personally used `hatch publish`. I still use `twine` based solutions IIRC. I've only dipped my toes into `hatch. @ofek can probably answer better.

ofek: Oh right, then you'd need Hatch still

oprypin: Anyway, why do I need `build`? Instead of trusting both `hatch` and `build` I could just be relying on `hatch`. In any case I trust `hatch` to continue doing the right thing going forward, more so than `build`

ofek: I didn't notice publish step, you can revert (but still you don't need `wheel`)

ofek (approved):

oprypin:
Sigh, still really scared about losing the "Used by" section, but let's hope for the best

ofek:
Success, it's still there 🙂

oprypin:
For now, anyway 🙂

PR #2991, opened 2022-10-03
Migrate from Tox to Hatch as test runner

oprypin:


oprypin:
Regarding needing to write `+type=` or `+py=` (with empty value) - that seems to be a bug

* https://github.com/pypa/hatch/issues/526

ofek:
You might prefer `hatch.toml` for environment config https://hatch.pypa.io/latest/intro/#configuration

oprypin:
I considered it briefly, but prefer one file. The main factor was that I want `dependencies` all to be in one file.

PR #3001, opened 2022-10-13
URL-encode paths when serving a redirect

blueswen:
When the `site_url` contains utf-8 characters (e.g. http://fake.domain/測試) and uses `mkdocs server` to serve doc, visiting localhost:8000 will cause wsgi encoding error when doing 302 redirects. The error log is as follows:

```
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/wsgiref/handlers.py", line 138, in run
    self.finish_response()
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/wsgiref/handlers.py", line 185, in finish_response
    self.finish_content()
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/wsgiref/handlers.py", line 324, in finish_content
    self.send_headers()
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/wsgiref/handlers.py", line 347, in send_headers
    self._write(bytes(self.headers))
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/wsgiref/headers.py", line 142, in __bytes__
    return str(self).encode('iso-8859-1')
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 11-12: ordinal not in range(256)
```

URL encoding those 302 paths could solve this issue. A minimal reproducible `mkdocs.yaml` is as follows:

```yaml
site_name: My Docs
site_url: http://fake.domain/測試
```

### DIFF
--- a/mkdocs/livereload/__init__.py
+++ b/mkdocs/livereload/__init__.py
@@ -15,6 +15,7 @@ import socketserver
 import string
 import threading
 import time
+import urllib.parse
 import warnings
 import wsgiref.simple_server
 from typing import Any, Callable, Dict, Optional, Tuple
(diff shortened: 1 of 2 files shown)


oprypin:
Thanks

PR #3008, opened 2022-10-16
Deprecate mkdocs.utils.warning_filter

oprypin:


ultrabug (approved):

mkdocs/utils/__init__.py
```diff
@@ -459,7 +459,13 @@ def get_counts(self) -> List[Tuple[str, int]]:
...
-# For backward compatibility as some plugins import it.
-# It is no longer necessary as all messages on the
-# `mkdocs` logger get counted automatically.
-warning_filter = logging.Filter()
+def __getattr__(name: str):
+    if name == 'warning_filter':
+        warnings.warn(
+            "warning_filter doesn't do anything since MkDocs 1.2 and will be removed soon. "
+            "All messages on the `mkdocs` logger get counted automatically.",
+            DeprecationWarning,
```
pawamoy, line 467: Should `stacklevel=2` be used? Not sure how this would behave when devs do `from mkdocs.utils import warnings_filter`.

oprypin: I've seen `stacklevel=2` more often as a cargo-culting mistake. No, it's basically just for warning *helpers*, which I don't have any.
Anyway MkDocs' warning printing code ignores stacklevel. Maybe you can check how this warning looks outside of MkDocs `__main__`. I only looked how it looks within MkDocs (and it was correct).

pawamoy (approved):

PR #3017, opened 2022-10-21
Make fields of File computed on demand

oprypin:
This means that plugin developers can change e.g. `dest_uri` without needing to manually update `abs_dest_path` as well. But they still can, if they wanted to. The field remains computed on demand until someone directly sets a value.


oprypin:
Sadly, this approach totally kills performance. These fields can be accessed a million times on a large site, they really have to stay plain fields.

PR #3018, opened 2022-10-21
 Implement cache busting: append a suffix to CSS and JS URLs 

### LINKED ISSUE #1979: Cache Busting
Is there any way to make mkdocs build the site with some hash string built into the names of the CSS and JS files to do some cache busting?

oprypin:
Fixes #1979

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -4,7 +4,7 @@ import gzip
 import logging
 import os
 import time
-from typing import Any, Dict, Optional, Sequence, Set, Union
+from typing import Any, Dict, Optional, Set
 from urllib.parse import urlsplit
 
 import jinja2
@@ -14,7 +14,7 @@ import mkdocs
 from mkdocs import utils
 from mkdocs.config.defaults import MkDocsConfig
 from mkdocs.exceptions import Abort, BuildError
-from mkdocs.structure.files import File, Files, get_files
+from mkdocs.structure.files import Files, get_files
 from mkdocs.structure.nav import Navigation, get_navigation
 from mkdocs.structure.pages import Page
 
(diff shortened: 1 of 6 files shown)


oprypin:
I slightly reworked the PR and I'm planning to move forward with it, after some polishing.

SagenKoder (approved):

oprypin:
How the options are currently named and what they are trying to express:

* `hash_rename_assets`: Patterns of filenames of assets that will be renamed by inserting a hash to the filename
* `hash_append_assets`: Patterns of filenames of assets that will always be linked by appending a query parameter with a hash

These names are **really bad**. Does anyone have ideas?

pawamoy:
`cache_busting_hash/query`? :man_shrugging:

ultrabug:
If we consider that we `add` the hash to either the `filename` or the `url`, here are a few names that come in mind:

- `use_hash_filename`
- `use_hash_url`

- `add_hash_filenames`
- `add_hash_urls`

- `hash_add_filenames`
- `hash_add_urls`

or simply

- `hashed_filenames`
- `hashed_urls`

Not sure which one I prefer yet...

kamilkrzyskow:
I like Ultrabug's last 2 options, however I fear that using `url` could be a bit confusing because users might input urls...

Since it's a list of patterns of paths to extras it might be something like:
`extras_hashed_via_rename` and (`extras_hashed_via_query` or `extras_hashed_via_param`) 😄 
or make it shorter removing the `extras_` part, 
`hashed_via_rename` and (`hashed_via_query` or `hashed_via_param`)
or even shorter
`hash_via_rename` and (`hash_via_query` or `hash_via_param`)

But then could it be confused for hashing the files themselves? 😵
Anyways, just my 2 cents ✌️

kchr:
How about the following?

- `hash_filenames`
- `hash_urls`

PR #3052, opened 2022-11-29
increase version range for Markdown

snpschaaf:
Currently I am trying bump the version of mkdocs for nixpkgs (NixOS/nixpkgs/pull/203592). 

The version of the `Markdown` package on [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) is slightly higher (v3.4.1) than the dependency range allows. 

Is it possible to increase it by one minor version? 

Currently I increase this range via patch file. I can tell that it works, at least for my documentation project.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ requires-python = ">=3.7"
 dependencies = [
     "click >=7.0",
     "Jinja2 >=2.11.1",
-    "Markdown >=3.2.1, <3.4",
+    "Markdown >=3.2.1, <=3.4.1",
     "PyYAML >=5.1",
     "watchdog >=2.0",
     "ghp-import >=1.0",


pawamoy:
Python-Markdown 3.4 removed deprecated objects, see https://python-markdown.github.io/change_log/release-3.4/
These removals probably break a lot of extensions. I believe MkDocs maintainers decided to prevent that by not allowing Markdown 3.4 in the version range.

oprypin:
That reply is correct. The version restriction shouldn't be there at all but it was a pragmatic decision.
It might be dropped after some time (~months).

snpschaaf:
My final thoughts on this:

I'm not an extension developer and don't know how they would pull in `Markdown`.
But my suggested change only would have impact if they use it directly from `mkdocs` (assumption).  
The previous Version range allowed only major version 3.x of `Markdown`. By the release of version  3.x the objects are already marked as deprecated. So extension developers had ignored the deprecated warning explicitly. 
So like foreseen by semver my suggestion would not be a breaking change (only updating the minor version).

pawamoy:
The removal of objects, whether they were deprecated for 1 day, 6 month, 10 years, or were not deprecated at all, *is* a breaking change.

I don't understand how [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) works: is there really only one version of Markdown available, and always the latest? Is that the same for every other project? What's the use of semantic versioning if only latest versions are available? There probably are many other examples where a project limits the accepted version range of a dependency? How does nixpkgs handle that usually?

snpschaaf:
> The removal of objects, whether they were deprecated for 1 day, 6 month, 10 years, or were not deprecated at all, is a breaking change. This probably should have been a major bump in Python-Markdown though, to respect semver.

I agree, this would have been more intuitive.

Regarding [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs): 
For one revision of nixpkgs there is only one version of any package (e.g. the pkg [python310Packages.markdown](https://search.nixos.org/packages?channel=22.11&from=0&size=50&sort=relevance&type=packages&query=python310Packages.markdown)).  

The tool nix has mechanisms to overwrite the version of the inner python package, the whole package as such or adding a new package (e.g. `python39Packages.markdown-legacy`). 

Then alternatively there is the option of fixing the package which is incompatible. This is the way I choose by adding a patch for mkdocs within nixpkgs. 

IMHO this is the cleanest option. Once mkdocs update its dependencies, the patch would not apply any longer. The next person updating the nix packages [python310Packages.mkdocs](https://search.nixos.org/packages?channel=22.11&show=python310Packages.mkdocs&from=0&size=50&sort=relevance&type=packages&query=python310Packages.mkdocs) and [python39Packages.mkdocs](https://search.nixos.org/packages?channel=22.11&show=python39Packages.mkdocs&from=0&size=50&sort=relevance&type=packages&query=python39Packages.mkdocs) could remove the patch.

pawamoy:
I see, thank you for the explanation, it's clearer now. I guess you'll have to stay patient indeed for now :smile:

PR #3070, opened 2022-12-14
Fix call to untyped `__init_subclass__` in typed context

mondeja:
I'm getting the next mypy error for `plugins.BasePlugin` trying to annotate a plugin:

```
Call to untyped function "__init_subclass__" in typed context
```

### DIFF
--- a/mkdocs/plugins.py
+++ b/mkdocs/plugins.py
@@ -85,7 +85,7 @@ class BasePlugin(Generic[SomeConfig]):
         name = f'{cls.__name__}[{config_class.__name__}]'
         return type(name, (cls,), dict(config_class=config_class))
 
-    def __init_subclass__(cls):
+    def __init_subclass__(cls) -> None:
         if not issubclass(cls.config_class, Config):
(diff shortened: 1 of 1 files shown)


oprypin:
I am guessing that you're using --disallow-untyped-calls. Please don't. It will not be supported.

mondeja:
I'm just using `strict = true` and `python_version = '3.7'`. Sorry I'm not an advanced mypy user, could we document this? How should I configure it?

oprypin:
https://github.com/mkdocs/mkdocs/blob/56b235a8ad43f2300d17f87e6fa4de7a3d764397/pyproject.toml#L208-L211

(particularly only the middle two are strictness configs)

I found this configuration to be useful. They are nice because they impose restrictions only on the main project, not on dependencies. I think most mypy flags are like that, you can pick and choose.

Just `--disallow-untyped-calls` stands out to me. It really should at least be isolated to the main project, not cross project boundaries. But even so, I don't know what it achieves.

Sorry for my tone 😞

PR #3129, opened 2023-02-24
Fix accessibility issue with top-level breadcrumb in ReadTheDocs theme

### LINKED ISSUE #2827: Top-level "Docs" breadcrumb disappeared in default theme
When we updated to mkdocs 1.3, our accessibility tests started failing on all pages. It turns out the top-level "Docs" breadcrumb has disappeared in the default theme.

Before:
image

After:
image

It looks like the [A tag inside wy-breadcrumbs was changed so that it no longer has any content](https://github.com/mkdocs/mkdocs/commit/fa52d3ae71251ae1c760cbe9b31caced816573d1#diff-cfc3ac6c9be30c04eecea2e3c295f14149b7ecdf159cdac641e6421673bd2847R3), which is invalid HTML for a link, hence why [our accessibility tests failed](https://github.com/jrnl-org/jrnl/runs/6142269783?check_suite_focus=true).

micahellison:
Fixes #2827.

This fix matches the change in the upstream ReadTheDocs theme by replacing an A tag's invalid "alt" attribute with an "aria-label" attribute that screen readers can consume.

### DIFF
--- a/mkdocs/themes/readthedocs/breadcrumbs.html
+++ b/mkdocs/themes/readthedocs/breadcrumbs.html
@@ -1,6 +1,6 @@
 <div role="navigation" aria-label="breadcrumbs navigation">
   <ul class="wy-breadcrumbs">
-    <li><a href="{{ nav.homepage.url|url }}" class="icon icon-home" alt="{% trans %}Docs{% endtrans %}"></a> &raquo;</li>
+    <li><a href="{{ nav.homepage.url|url }}" class="icon icon-home" aria-label="{% trans %}Docs{% endtrans %}"></a> &raquo;</li>
     {%- if page %}
       {%- for doc in page.ancestors[::-1] %}
         {%- if doc.link %}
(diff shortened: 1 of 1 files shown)


ultrabug:
@micahellison thanks for your proposal. I'd like to ask why you think necessary to remove the support for translation of the "Docs" term to a fixed "Home" tho?

Can't you just change the alt by an aria-label while preserving the localization feature?

micahellison:
> @micahellison thanks for your proposal. I'd like to ask why you think necessary to remove the support for translation of the "Docs" term to a fixed "Home" tho?
> 
> Can't you just change the alt by an aria-label while preserving the localization feature?

Yes, that was what I did at first, though I changed it to "Home" because I had the understanding that mkdocs wants to mirror the RTD Sphinx theme, and that theme uses "Home" instead of the translation. Line 25 here: https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/breadcrumbs.html

I'm happy to change it back, though.

oprypin (approved):

oprypin (approved):

PR #3156, opened 2023-03-15
Use f-string in another place

xeare:


xeare:
Looks like CI / lint is failing, similar to other PRs a head of me. I'm happy to fix something here if i need to to get that working in this branch.

pawamoy:
The title is a bit misleading: the code was already correct, you just change the %-formatted string to an f-string. I have nothing against using an f-string. Let see what the maintainers say :slightly_smiling_face:

xeare:
> The title is a bit misleading: the code was already correct, you just change the %-formatted string to an f-string. I have nothing against using an f-string. Let see what the maintainers say 🙂

I have nothing against % formatting either but i didn't think it was printing out as intended
```python

import time

start = time.monotonic()

print('Documentation built in %.2f seconds', time.monotonic() - start)
# outputs: Documentation built in %.2f seconds 0.0

print(f'Documentation built in {time.monotonic() - start:.2f} seconds')
# outputs: Documentation built in 0.00 seconds

```

I think the second one is how it was intended to be printed out. As for the f-string, it's being used throughout the file, so i didn't think that would be a problem.

pawamoy:
Oh, that's a feature of the logging system, you can pass inputs as positional arguments, so the current code is definitely correct. `print` doesn't do that indeed.

pawamoy:
Just checked the module and indeed, other log messages are built using f-strings, so I'd be keen to merge this. I'll do so in a few days if nobody objects to it.

oprypin (approved):

PR #3186, opened 2023-04-17
Fix year on footer

arsalanses:
I changed year from 2014 to 2023


pawamoy:
That's not really how copyright dates work. In this case, it says that MkDocs holds the copyright since 2014 up until now. Changing it to 2023 would be wrong. Closing, feel free to comment further!

PR #3202, opened 2023-04-30
Enable caching of compiled Jinja templates

oprypin:
This should reduce startup times a bit.

### DIFF
--- a/mkdocs/theme.py
+++ b/mkdocs/theme.py
@@ -114,8 +114,16 @@ class Theme:
         """Return a Jinja environment for the theme."""
 
         loader = jinja2.FileSystemLoader(self.dirs)
-        # No autoreload because editing a template in the middle of a build is not useful.
-        env = jinja2.Environment(loader=loader, auto_reload=False)
+        try:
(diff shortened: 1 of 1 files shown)


ultrabug:
This idea is cool and I already feel like it could benefit multi-language builds such as when used with the `mkdocs-static-i18n` tho I'd have to check.

While the `FileSystemBytecodeCache` implements a `clear()` method but it is not called by default which means that we leave potentially useful templating information in the `/tmp` directory of users after a build.

I'm unsure about where we should trigger a `clear()` to still allow users to benefit from caching but I'm under the impression that we should allow users to opt in or out this feature as well.

oprypin:
The only point of this cache is to persist across invocations. If it is cleaned up on exit, that's the same as not having it

ultrabug:
My point is that multi-build users could benefit from this while still clearing the cache at the very end (if we make it possible / tunable)

oprypin:
True, but that can also be achieved without writing anything to disk and instead just preserving the `jinja2.Environment` object.
Yeah worth thinking

PR #3203, opened 2023-04-30
Tiny optimization: stop Click looking for localization files

oprypin:


pawamoy:
Should we set it back to whatever value it had before, right after Click has finished loading? I fear it could have unwanted side effects for users who rely on this environment variable.

oprypin:
Hmm yes maybe some people are relying on the variable. Then it's not worth it

PR #3204, opened 2023-04-30
Copy warning about ignored default plugins in Using plugins section

pawamoy:
Closes https://github.com/mkdocs/mkdocs/pull/2319.

It's totally acceptable that we think this is just redundant, and risks to diverge at some point. I'll gladly close both PRs in that case.

### DIFF
--- a/docs/dev-guide/plugins.md
+++ b/docs/dev-guide/plugins.md
@@ -31,6 +31,16 @@ plugins:
     - search
 ```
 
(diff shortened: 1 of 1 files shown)


oprypin:
Hm this is exactly duplicated indeed, doesn't fit as well here, and anyway this page is under `dev-guide/`. I'd prefer to keep this as is.

pawamoy:
Fine by me! I'll close #2319 as well.

PR #3219, opened 2023-05-12
Modernize type annotations using `pyupgrade`

oprypin:
* https://peps.python.org/pep-0563/ - Postponed Evaluation of Annotations
* https://peps.python.org/pep-0585/ - Type Hinting Generics In Standard Collections
* https://peps.python.org/pep-0604/ - Allow writing union types as `X | Y`

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -4,7 +4,7 @@ import gzip
 import logging
 import os
 import time
-from typing import Any, Dict, Optional, Sequence, Set, Union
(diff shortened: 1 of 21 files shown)


pawamoy:
There's a flake8 plugin that tells you if an import should be put under `if TYPE_CHECKING`. Ruff does that as well. Could be a nice linting addition.

oprypin:
> Nice! Did you use pyupgrade, or a mix of search/replace?

pyupgrade and a bit of manual followup

PR #3224, opened 2023-05-19
New `exclude_docs` config: gitignore-like patterns of files to exclude

### LINKED ISSUE #1888: Disable navigation warning

### LINKED ISSUE #2139: Hidden files/directories
Revival of #1738.

Could you consider adding an option to serve hidden static files and templates? I cannot see any reason why one wouldn't want to serve these hidden files, and in particular the `.well-known` directory is used for a lot of things that one could want in their documentation, including:
- ACME challenges (for Let's Encrypt)
- [`security.txt`](https://securitytxt.org/)
- Keybase Site Verification
- Do-Not-Track policies
- HTTPS Opportunistic Encryption

see https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml

It would be very useful to have this as a built-in option. A plugin, which could work around this, seems completely unnecessary. Could you explain why this is not enabled, and reconsider the behaviour, or at least add an exception for the [well standardised](https://tools.ietf.org/html/rfc5785)  `.well-known` directory? Or is there a good reason I'm missing not to do this?

### LINKED ISSUE #2185: Files in a directory called templates are ignored
In a project where the documentation files are located under `docs/`, it seems that files inside a sub-directory called `templates/` (path being `docs/templates/`) are reported as missing when referring to them in the configuration file.

The configuration is like:

```
<redacted>

nav:
  - Introduction: 'index.md'
  - Templating:
    - Writing Templates: 'templates/index.md'
    - Arista EOS: 'templates/arista-eos.md'
    - Cisco IOS-XR: 'templates/cisco-iosxr.md'
    - Juniper JUNOS: 'templates/juniper-junos.md'
    - Peering Request E-mail: 'templates/peering-request-email.md'

<redacted>
```

When trying to `build` or `serve`, the following warning are raised (using `mkdocs >= 1.0`).

```
WARNING -  A relative path to 'templates/index.md' is included in the 'nav' configuration, which is not found in the documentation files 
WARNING -  A relative path to 'templates/arista-eos.md' is included in the 'nav' configuration, which is not found in the documentation files 
WARNING -  A relative path to 'templates/cisco-iosxr.md' is included in the 'nav' configuration, which is not found in the documentation files 
WARNING -  A relative path to 'templates/juniper-junos.md' is included in the 'nav' configuration, which is not found in the documentation files 
WARNING -  A relative path to 'templates/peering-request-email.md' is included in the 'nav' configuration, which is not found in the documentation files
```

Please note that the `docs/` directory is also at the same level in the file system than another directory called `templates/`. So from a file system perspective it looks like:

```
 docs/
  `---- index.md
  `---- templates/
somedir/
someotherdir/
templates/
```

Is it a normal behavior or is it a corner case?

### LINKED ISSUE #2316: Is there a way that I can exclude certain folders to not render on static site
I have a case where the repo A contains XYZ folders. I want the content in Y folder not to render on the static site. How can it be aheived?

### LINKED ISSUE #3110: Allow readme.md and index.md files
I have a use case having a `README.md` and an `index.md` file on my docs folder.

- The readme files content is repository specific and is the documentation that we want to show on github. 
- The index file is the project documentation homepage.

I tried to to define this on my `mkdocs.yml`: 
```yml
nav:
  - Introduction: index.md
  - Other: "other/index.md"
  - Repo Readme: "README.md"
```

I get that readme file is not considered as index if a `index.md` file is present, but why not allow still access to that file if we want to add that reference?

UPDATE
current vs expected behaviour
 
Current:
Add a file index.md to docs. A file index.html gets created in the site. The nav link leads there.
Add a file README.md to docs. A file index.html gets created in the site. The nav link leads there.
Add a file index.md to docs. Add a file README.md to docs. A file index.html gets created in the site. using the `index.md` file. The nav link leads there. the README.md file is ignored completely showing a warning about it

Expected
Add a file index.md to docs. A file index.html gets created in the site. The nav link leads there. (same)
Add a file README.md to docs. A file index.html gets created in the site. The nav link leads there. (same)
Add a file index.md to docs. Add a file README.md to docs. A file index.html gets created in the site. using the `index.md` file. The nav link leads there. a Readme.html file is created using the README.md file

oprypin:
These files will not be picked up into the `Files` collection and as such, also the final built site.

Also adds the ability to *un*-ignore the files that MkDocs forcibly ignores.

* Closes #2139
* Closes #2316
* Closes #2185
* Closes #1888
* Closes #3110

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,2 @@
 [flake8]
-max-line-length = 119
-extend-ignore = E203
+extend-ignore = E501, E203
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -310,7 +310,7 @@ exclude_docs: |
 
 This follows the [.gitignore pattern format](https://git-scm.com/docs/gitignore#_pattern_format).
 
-Note that `mkdocs serve` does *not* follow this setting and instead displays excluded documents but with a "DRAFT" mark.
+Note that `mkdocs serve` does *not* follow this setting and instead displays excluded documents but with a "DRAFT" mark. To prevent this effect, you can run `mkdocs serve --clean`.
 
 The following defaults are always implicitly prepended - to exclude dot-files (and directories) as well as the top-level `templates` directory:
 
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -310,6 +310,8 @@ exclude_docs: |
 
 This follows the [.gitignore pattern format](https://git-scm.com/docs/gitignore#_pattern_format).
 
+Note that `mkdocs serve` does *not* follow this setting and instead displays excluded documents but with a "DRAFT" mark.
+
 The following defaults are always implicitly prepended - to exclude dot-files (and directories) as well as the top-level `templates` directory:
 
 ```yaml
@@ -326,6 +328,31 @@ Otherwise you could for example opt only certain dot-files back into the site:
 exclude_docs: |
     !.assets  # Don't exclude '.assets' although all other '.*' are excluded
 ```
+
+### not_in_nav
+
+NEW: **New in version 1.5.**
+
+NOTE: This option does *not* actually exclude anything from the nav.
+
+If you want to include some docs into the site but intentionally exclude them from the nav, normally MkDocs warns about this.
+
+Adding such patterns of files (relative to [`docs_dir`](#docs_dir)) into the `not_in_nav` config will prevent such warnings.
+
+Example:
+
+```yaml
+nav:
+    - Foo: foo.md
+    - Bar: bar.md
+
+not_in_nav: |
+    /private.md
+```
+
+As the previous option, this follows the .gitignore pattern format.
+
+NOTE: Adding a given file to [`exclude_docs`](#exclude_docs) takes precedence over and implies `not_in_nav`.
 
 ## Build directories
 
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -291,6 +291,42 @@ server root and effectively points to `https://example.com/bugs/`. Of course, th
 list of all the Markdown files found within the `docs_dir` and its
 sub-directories. Index files will always be listed first within a sub-section.
 
+### exclude_docs
+
+NEW: **New in version 1.5.**
+
+This config defines patterns of files (under [`docs_dir`](#docs_dir)) to not be picked up into the built site.
+
+Example:
+
+```yaml
+exclude_docs: |
+    api-config.json    # A file with this name anywhere.
+    drafts/            # A "drafts" directory anywhere.
+    /requirements.txt  # Top-level "docs/requirements.txt".
+    *.py               # Any file with this extension anywhere.
+    !/foo/example.py   # But keep this particular file.
+```
+
+This follows the [.gitignore pattern format](https://git-scm.com/docs/gitignore#_pattern_format).
+
+The following defaults are always implicitly prepended - to exclude dot-files (and directories) as well as the top-level `templates` directory:
+
+```yaml
+exclude_docs: |
+    .*
+    /templates/
+```
+
+So, in order to really start this config fresh, you'd need to specify a negated version of these entries first.
+
+Otherwise you could for example opt only certain dot-files back into the site:
+
+```yaml
+exclude_docs: |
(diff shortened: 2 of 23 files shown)


ssbarnea (approved):

oprypin:
I'm wondering, maybe for the sake of plugins it would be better to keep these files in the `files` collection but then only exclude them from being copied to the destination.

oprypin:
Yet another idea I just had- for the sake of writing drafts, these pages would still be shown in `mkdocs serve` (perhaps with a forcibly inserted message about that on the page) but then excluded from `build`

pawamoy:
That would be really useful indeed. I wonder if the corresponding nav entry for a given excluded file/page should also be removed when building, because when I write something for my blog I have to comment/uncomment the post in the nav until it's ready.

oprypin:
I have made the changes from both ideas that I described.

See docs [in the diff](https://github.com/mkdocs/mkdocs/pull/3224/files#diff-546d5ce955be10a583e1890e3a17ae921c5e5166f30a67bf3635f8d697b0eda5).

If people could try out this code and see if it satisfies their wishes, that would be nice.
```bash
pip install git+https://github.com/mkdocs/mkdocs.git@refs/pull/3224/head
```

---

There's also the rather ugly matter of injecting raw HTML:

https://github.com/mkdocs/mkdocs/blob/89f15d9c6952a6ba28db4dae48d182d25699e90f/mkdocs/commands/build.py#L218-L223

The good thing is that all existing themes will support this out of the box (and still will be able to customize this if desired).
The bad things are kinda clear, and also it's not even translated. But how to do that it if translations are only supported for themes?

If you have better ideas how to achieve this, do let me know.

pawamoy:
Working fine :slightly_smiling_face: 
I'd still love *not* to have to comment items in the nav while they're being worked on, though this is probably another feature and could be implemented in a plugin.

What's the reason for multiline strings instead of arrays for the config settings?

oprypin:
> What's the reason for multiline strings instead of arrays for the config settings?

* Special characters such as `!\*?` are involved, this way we're avoiding weird YAML behavior.
* A text block is the original way to specify a gitignore file. I might add a feature to paste an external text file into mkdocs.yml, and with this approach it will fit perfectly.

oprypin:
> I'd still love _not_ to have to comment items in the nav while they're being worked on, though this is probably another feature and could be implemented in a plugin.

If someone adds a doc to the nav but the doc will be excluded, I suspect that more people would find behavior [1] more intuitive, not sure why you prefer [2].

1. Warn that an excluded file is being added to the nav, something should be done.
2. Silently obliterate the file also from the nav.

Imagine a situation where someone already has a wide exclude pattern but then forgets about it and really tries to add a file but nothing is happening (and they can even overlook this and push the site).

@pawamoy Could you provide arguments for your case?

pawamoy:
> Could you provide arguments for your case?

Hmmm I think my main argument would be that such a feature would allow me to exclude a file and remove the corresponding nav entry automatically, until I'm ready to publish it, and in which case I either move the file around or stop excluding it. In short: exclude it once when you start working on it, re-include it once when you're ready. Without such a feature, I constantly have to comment/uncomment the nav entry in mkdocs.yml, at least if I publish other things while working on the draft.

oprypin:
Hmm but you could just not add it to the nav until it's ready?

And what about my counterargument?

pawamoy:
> Hmm but you could just not add it to the nav until it's ready?

Yeah true, I guess I'm lazy and don't want to type the URL myself instead of clicking :thinking: 
Your counter argument is a good one, silently doing things will be create confusion.

I remember Jekyll had some first-class support for drafts. I'll check again how it works to see if there's something we can improve in MkDocs :slightly_smiling_face:

oprypin:
> Yeah true, I guess I'm lazy and don't want to type the URL myself instead of clicking 🤔

Hmm that gives me an idea
https://github.com/mkdocs/mkdocs/blob/89f15d9c6952a6ba28db4dae48d182d25699e90f/mkdocs/commands/build.py#L314-L316
what if this message would proactively print the localhost URLs of the pages 😃 rather than just the path

pawamoy:
That would be nice yes :)

PR #3225, opened 2023-05-23
fix MarkdownExtensions type annotation

rmorshea:
based on the validation function any of the following are possible:

- `list[str]`
- `dict[str, Any]`
- `list[dict[str, Any]]`

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -833,7 +833,7 @@ class Private(BaseConfigOption):
             raise ValidationError('For internal use only.')
 
 
-class MarkdownExtensions(OptionallyRequired[List[str]]):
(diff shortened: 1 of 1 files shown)


oprypin:
This annotation isn't about the input but about what `run_validation` returns. And it does seem to return `List[str]`, so correct.

What is the reason that you ran into this and wanted to change this?

rmorshea:
Hmm, perhaps when I'm

rmorshea:
Looking more closely it seems like those extra configuration options get stored under `mdx_configs`. The docstring says it's for internal use only, is it appropriate to mutate extension options in `on_config` in the way I suggested?

oprypin:
It might be the case that the types for configs are all messed up during `on_config`. I'll need to look later. Could you post your last message as a Discussion - how to set an extension's config

rmorshea:
https://github.com/mkdocs/mkdocs/discussions/3226

PR #3238, opened 2023-06-04
Hide instance-attribute labels in docs

pawamoy:
I've updated mkdocstrings' Python handler, and while checking the docs for MkDocs I noticed these labels were displayed. Since you already hide `class-attribute` labels, I figured you'd like to hide these ones as well.


oprypin:
Thanks!

PR #3245, opened 2023-06-06
Add plugin logger

pawamoy:
As discussed in https://github.com/mkdocs/mkdocs/discussions/3241.
Not sure if tests are needed, the code is quite trivial.

### DIFF
--- a/mkdocs/plugins.py
+++ b/mkdocs/plugins.py
@@ -510,3 +510,50 @@ class PluginCollection(dict, MutableMapping[str, BasePlugin]):
             if result is not None:
                 item = result
         return item
+
+
+class _PrefixLogger(logging.LoggerAdapter):
+    """A logger adapter to prefix log messages."""
+
+    def __init__(self, prefix: str, logger: logging.Logger):
+        """
+        Initialize the logger adapter.
+
+        Arguments:
+            prefix: The string to insert in front of every message.
+            logger: The logger instance.
+        """
+        super().__init__(logger, {})
+        self.prefix = prefix
+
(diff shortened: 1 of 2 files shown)


mkdocs/plugins.py
```diff
@@ -510,3 +510,50 @@ def run_event(self, name: str, item=None, **kwargs):
...
+        name: The name to use with `logging.getLogger`.
+
+    Returns:
+        A logger configured to work well in MkDocs,
+            prefixing each message with the plugin package name.
+
+    Examples:
+        >>> from mkdocs.plugins import get_plugin_logger
+        >>> logger = get_plugin_logger(__name__)
+        >>> logger.info("My plugin message")
```
oprypin: Hm MkDocs never uses `>>>` in examples, I'd prefer to keep it that way

mkdocs/plugins.py
```diff
@@ -510,3 +510,50 @@ def run_event(self, name: str, item=None, **kwargs):
             if result is not None:
                 item = result
         return item
+
+
+class _PrefixLogger(logging.LoggerAdapter):
```
oprypin: Could you rename `_PrefixLogger` to `PrefixedLogger` (also no underscore)

mkdocs/plugins.py
pawamoy: Python code block then?

oprypin: Yes

oprypin (approved):
Thanks

pawamoy:
Docs screenshot:
pluginlogger

PR #3255, opened 2023-06-11
Don't abort `mkdocs serve` for build errors during rebuilds

### LINKED ISSUE #3166: mkdocs serve should keep running when compilation crashes
I often break the structure of a markdown while editing it. I expect `mkdocs serve` to tell me about a compilation error but to continuously try to recompile on new saved changes (this is also the default behavior of [`mdbook serve`](https://github.com/rust-lang/mdBook/)).

I cannot find an option in mkdocs for such continuous compilation so I often have to redo the `serve` operation. Please let me know if I am missing something.

oprypin:
* Closes #3166

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -353,7 +353,7 @@ def build(
         counts = warning_counter.get_counts()
         if counts:
             msg = ', '.join(f'{v} {k.lower()}s' for k, v in counts)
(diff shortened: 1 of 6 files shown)


pawamoy (approved):
Nice :+1:

sgleske-ias:
Any way to have an option where mkdocs serve intentionally crashes if reload fails.  It would help me with a workaround for #3972 by putting it into a while/until loop.

Issue #3264, opened 2023-06-16
Dir in SubConfig is not evaluated relative to mkdocs.yaml


couling:
### What I expected vs what happened:
Placing a `Dir` type field in a `SubConfig` does not behave as expected:
- I expected: `dir` to be evaluated relative to the dir containing mkdocs.yaml (`./foo`)
- What happend: `dir` was evaluated relative to CWD (`.`)

### Steps to reproduce
Create a mkdocs plugin with a config like this:

```python
class MyConfigDir(c.Config):
    action: str = c.Type(str)
    dir: Path = c.Dir(exists=True)

class MyConfig(base.Config):
    search: list[MyConfigDir] = c.ListOfItems(c.SubConfig(MyConfigDir))
```
Configure it with this:

```yaml
- my-plugin:
    search:
    - dir: ../bar
      action: something
```

Then run from another directory:

```shell
mkdir bar
mkdocs build --config-file foo/mkdocs.yaml
```

### The error

```
Config value 'plugins': Plugin 'my-plugin' option 'search': Sub-option 'dir': The path '../source' isn't an existing directory.
```

oprypin:
Thanks for the great report. I confirm this, there is indeed an error with this diff

```diff
diff --git a/mkdocs.yml b/mkdocs.yml
index b01297467..6b177e30f 100644
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -8,2 +8,6 @@ edit_uri: blob/master/docs/
 
+foo:
+    - dir: docs
+      action: something
+
 theme:
diff --git a/mkdocs/config/defaults.py b/mkdocs/config/defaults.py
index ba6681d3d..6ffc56f0a 100644
--- a/mkdocs/config/defaults.py
+++ b/mkdocs/config/defaults.py
@@ -12,2 +12,7 @@ def get_schema() -> base.PlainConfigSchema:
 
+class MyConfigDir(base.Config):
+    action: str = c.Type(str)
+    dir: Path = c.Dir(exists=True)
+
+
 # NOTE: The order here is important. During validation some config options
@@ -18,2 +23,4 @@ class MkDocsConfig(base.Config):
 
+    foo = c.ListOfItems(c.SubConfig(MyConfigDir))
+
     config_file_path: str = c.Optional(c.Type(str))  # type: ignore[assignment]```

PR #3265, opened 2023-06-16
Pass config file path through SubConfig to Config

couling:
This ensures FilesystemObject fields in Config are correctly evaluated relative to mkdocs.yaml.

See #3264

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -58,6 +58,8 @@ class SubConfig(Generic[SomeConfig], BaseConfigOption[SomeConfig]):
     `config_options` are ignored (`validate=False`). Users should typically
     enable validation with `validate=True`.
     """
+    
+    __config_file_path = None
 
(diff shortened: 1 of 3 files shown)


oprypin:
Thanks for the PR.

Well, the more important thing here would be adding a test that would've failed before. Also I'll need to think for a bit whether there are any implications for passing `config_file_path` to all subclasses of `Config` or whether it can be considered a breaking change by now

couling:
@oprypin Thanks!  Please confirm that my legacy unit test is actually a legacy unit test, I've not used that style before. Hopefully this will be non-breaking.

oprypin:
Hi sorry, I'm overthinking how to deal with a pull request. Would it be OK for me to just push my edits to this pull request?

couling:
Yeah should be fine. I don't recall if I have a local copy so please don't force push.

oprypin (approved):

PR #3272, opened 2023-06-22
Automatically generate the list of locales for docs

oprypin:


ultrabug (approved):
elegant, will avoid mistakes

PR #3282, opened 2023-07-01
Make colorful/indented output optional, decrease indent

### LINKED ISSUE #2862: Disable color output in terminal when NO_COLOR is set
It would be nice to turn off terminal colors sometimes, but there currently is no option to do that.

https://no-color.org/ is the closest thing to a standard I could find, and they recommend disabling color if the `NO_COLOR` environment variable is set (to any value).

oprypin:
* These disable it: `--no-color` flag, `NOCOLOR=` var, or `| cat`.
* This force-enables it: `--color`

---

* Closes #2862

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -67,20 +67,20 @@ class ColorFormatter(logging.Formatter):
         replace_whitespace=False,
         break_long_words=False,
         break_on_hyphens=False,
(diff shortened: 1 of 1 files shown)


pawamoy (approved):

evanrichter:
Thank you so much!!

PR #3287, opened 2023-07-09
Use YAML indentation width of 2 spaces

oprypin:


ultrabug:
I wonder if a tool could ensure this rule in the future on the CI

ultrabug (approved):

oprypin:
> I wonder if a tool could ensure this rule in the future on the CI

There are solutions for stepping into Markdown blocks but I don't know any solution offhand that would validate YAML indentation in the first place. Though of course statistically there has to be one.

PR #3334, opened 2023-08-11
Add new supported languages by updating lunr-languages to 1.12.0

ultrabug:
Bump from 1.4.0 to 1.12.0

This adds support for:

- hi
- hy
- kn
- ko
- sa
- ta
- te
- zh


oprypin (approved):
Thanks! Just double-checked that this matches the upstream :)

PR #3380, opened 2023-09-09
Run `on_env` plugin events sooner

RealOrangeOne:
This ensures any modifications to the env are always used, and it's fully configured as early as possible.

Note: I see there's a `plugin-events.py` script to edit the SVG diagram, but I'm not sure what edits need to be made there - help appreciated!

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -294,12 +294,13 @@ def build(
             if dirty and site_directory_contains_stale_files(config.site_dir):
                 log.info("The directory contains stale files. Use --clean to remove them.")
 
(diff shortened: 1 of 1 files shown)


oprypin:
This will not be accepted. It is absolutely certain to break many plugins.

oprypin:
Running `on_files` and then `on_env` is a hard guarantee by now, that cannot be changed

RealOrangeOne:
@oprypin Good point. Perhaps moving it directly below `on_files` would be viable? Notably, I think running it above `get_page_markdown` is a good idea.

PR #3381, opened 2023-09-10
Keep the prior behavior of overwriting File.page

oprypin:
unless the page value is a *strict* subclass of Page

* Addresses https://github.com/mkdocs/mkdocs/pull/3367#issuecomment-1710729408

@squidfunk @ultrabug

### DIFF
--- a/mkdocs/structure/nav.py
+++ b/mkdocs/structure/nav.py
@@ -201,12 +201,22 @@ def _data_to_navigation(data, files: Files, config: MkDocsConfig):
                 f"A reference to '{file.src_path}' is included in the 'nav' "
                 "configuration, but this file is excluded from the built site.",
             )
-        if file.page is not None:
-            if not isinstance(file.page, Page):
-                warnings.warn(  # type: ignore[unreachable]
-                    "File.page should not be set to any type other than Page", DeprecationWarning
+        page = file.page
+        if page is not None:
+            if isinstance(page, Page):
+                if type(page) is not Page:  # Strict subclass
+                    return page
+                warnings.warn(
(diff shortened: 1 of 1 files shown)


squidfunk:
LGTM. So, the regressions that we've seen should be gone, but judging from the code, you decided to issue warnings if a plugin overwrites `Page`. Are we planning to switch to the behavior proposed in #3367 in MkDocs 1.6?

squidfunk (approved):

oprypin:
You'll be able to move forward with your plugin in 1.5.3 because you're working with strict subclasses and they're supported without warning. For other cases I kept old behavior and added warnings.

squidfunk:
Thanks, yes, I was asking what the plans for 1.6 were, because [the warning](https://github.com/mkdocs/mkdocs/pull/3381/files#diff-de72283007c995e5aa484d7f2e25d4a0c65ac42baeaf46c94db80147fee72beaR211) is not entirely clear to me:

``` py
warnings.warn(
  "A plugin has set File.page to an instance of Page and it got overwritten. "
  "The behavior of this will change in MkDocs 1.6.",
  DeprecationWarning,
)
```

oprypin:
Here's the list of cases:
* Strict subclass: 1.5.3 just accepts it and 1.6.0 still just accepts it
* Page: 1.5.3 ignores/overwrites it with warning, 1.6.0 just accepts it &nbsp; &larr; what you're asking about
* Other type: 1.5.3 ignores/overwrites it with warning, 1.6.0 errors
* None: always ignore/overwrite

squidfunk:
Perfect, thanks!

ultrabug (approved):
Awesome thanks!

PR #3383, opened 2023-09-10
Fix breadcrumb display in readthedocs theme

### LINKED ISSUE #3382: Something is wrong with readthedocs theme breadcrumbs
It probably should look like the upstream theme:

image

But it looks like this right now:

image

oprypin:
- Fix broken styling of breadcrumbs in readthedocs theme

  Commit https://github.com/mkdocs/mkdocs/commit/3c3670d92c972d874d7fe927df5abf728e003a66 left it half way between the old and new style

  Fixes #3382

- Fix linkified breadcrumb items in readthedocs theme

  A doc can never have a `link` attribute so this branch never kicked in.
  Actually even now, this branch will still not kick in, only mkdocs-section-index makes this state possible.


ultrabug (approved):
Nice catch o_O

PR #3394, opened 2023-09-14
Revert to pre-1.5 title rendering behavior

### LINKED ISSUE #3357: Header and navigation sidebars display incorrectly arbitrary markup
Hello!

```md
```

## External links:

oprypin:
MkDocs 1.5 changed so that title detection is based fully on Markdown parsing, which makes it more reliable. But that also brought in another change - that Markdown actually gets partly rendered.

This way we can keep *only* the H1 tag detection and yank the title as soon as possible - before processing any inline elements.

* Fixes #3357

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -454,5 +454,5 @@ class _ExtractTitleTreeprocessor(markdown.treeprocessors.Treeprocessor):
         md.treeprocessors.register(
             self,
             "mkdocs_extract_title",
-            priority=-1,  # After the end.
+            priority=40,  # First, or at least much earlier than `inline`.
         )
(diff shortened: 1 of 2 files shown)


ultrabug:
Just to clarify and illustrate what this PR allows and fixes.

Using [@Enn83 's demo repo](https://github.com/squidfunk/mkdocs-material/issues/5893#issue-1866480620), this is what current (breaking) MkDocs looks like:

2023-09-15-103848_952x899_scrot

This PR fixes it and looks like this:

2023-09-15-103458_781x413_scrot

2023-09-15-103510_944x896_scrot

oprypin:
Per https://github.com/mkdocs/mkdocs/issues/3357#issuecomment-1925438350 I am now fully sure that the old behavior was complete nonsense, so going back to it is not the way

PR #3400, opened 2023-09-18
Change PyPI deploy to rely on "trusted publishers"

oprypin:
https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
-no longer relies on any token


ultrabug (approved):
TIL the newest way of release CD :package: 

I kind of like relying only on hatch but I guess sticking to pypa is better

PR #3421, opened 2023-10-12
minor changes

haleemas311297:


oprypin:
Aight found the source of the garbage, thanks for the pointer

https://youtu.be/aaaZ-sjnjZ0?t=17459

pawamoy:
Hmmm, they use MkDocs repo to show `git blame` commands. The video does not explicitly show how to create a pull request on GitHub using MkDocs repo :thinking: But yeah that's probably one of the sources (some PRs were created before the video was published).

oprypin:
* The video I linked is pirated content of the actual course, so its "publishing" time is not representative.
* I feel like the video skips over something, or maybe the "graded assessment" (definitely not represented in the video) has something more to do with it.

Or maybe students just like to test the depth of their knowledge by doing something slightly off-script but are guided by the part that mentions "MK docs"

pawamoy:
Yeah, dunno. Anyway, not much we can do to prevent PRs from being created here :confused:

ultrabug:
If they keep coming too much, we might ask GH for advise as well. I guess they have experience in this too.

PR #3429, opened 2023-10-16
Drop Python 3.7, officially support Python 3.12

oprypin:


pyproject.toml
```diff
@@ -210,7 +210,6 @@ dependencies = [
 
 [tool.black]
 line-length = 100
-target-version = ["py37"]  # 3.7
```
pawamoy, line 213: Shouldn't we target 3.8 explicitly?

oprypin: No, they now have a feature to infer it from requires-python

ultrabug (approved):
Cool!

PR #3437, opened 2023-10-21
Stop config validation after the first error

oprypin:
If one config option completely fails validation, that means its value remains "raw" from YAML, which can cause unpredictable errors when validating other options that might look at the value. Instead just decide one error is enough.

* See one such case: https://github.com/mkdocs/mkdocs/pull/3351


ultrabug (approved):
Indeed, I like fail fast better

PR #3448, opened 2023-10-29
Allow adding event handlers at multiple priorities

oprypin:
-within one plugin.

### DIFF
--- a/docs/dev-guide/plugins.md
+++ b/docs/dev-guide/plugins.md
@@ -429,6 +429,12 @@ Since MkDocs 1.4, plugins can choose to set a priority value for their events. E
 
 #### ::: mkdocs.plugins.event_priority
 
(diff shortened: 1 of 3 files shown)


squidfunk:
Why do you think this should be supported?

oprypin:
I think a better question is why shouldn't it be supported. Right now it's a limitation based only on how it's implemented. If someone was designing this system anew, surely they wouldn't say that it's *better* that you can't have an early on_files handler if you already have a late handler for it.

squidfunk:
Okay, so I read that there's no immediate need why this get's implemented, but it's a nice-to-have. From a plugin author's perspective, I'm not sure it's necessary, though. At least, I haven't seen one sound use case that would mandate for it.

ultrabug (approved):
I love the idea and can clearly see use cases, thanks!

PR #3450, opened 2023-10-29
Support docs from the root directory.

athackst:
Per discussion #3062 this change would allow users to build a mkdocs site from the root directory.

This change removes the validation checks for the site directory to not be within the docs directory.
If the site folder is within the docs directory, it will skip processing those files.

Side effects:
- The livereload server changes the site directory to a temp folder, so a previously built site directory would be copied to the temp folder.  This can be prevented by one of the following methods: 
    - (1) Setting exclude_docs to the original site directory and (2) Running `mkdocs serve --clean`
    - Deleting the previously built site directory previous to mkdocs serve
- The livereload server copies hidden files, such as those within the ".git" directory.  This can be prevented by running `mkdocs serve --clean`
- The livereload server reloads when any change is made in the root directory.  However in my testing this has resulted in a single reload.

The extra copies of files created by default using the server does not create any usability issues on the resulting site (site loads as expected, all links work as expected).  If users have many extra files in their root folder, they may experience additional time building due to the extra copies.


athackst:
It might be worth considering an "ignore" option for `InclusionLevel` which would additionally not copy to the server.

oprypin:
Thanks for pushing this forward. Sadly, I think this is too many steps that users would have to take in order to use this. If someone runs into this unprepared, confusion is fully guaranteed.
I think this `--clean` in particular is nasty for this, and maybe that "draft" feature was a mistake in general.
I want to more comprehensively solve all of these small issues before simply unlocking this

athackst:
To be clear, it's still usable as is. A normal user would not even notice the extra site folder since nothing links to it. It just copies more than it needs to into the temp directory that serves the site. 

It performs as expected on a normal build.

athackst:
@oprypin I updated this one after rebasing on top of #3476 

The site directory is now appropriately marked as excluded both in build and serve.

squidfunk:
Note that advertising that  `site_dir` can be contained within `docs_dir`, albeit it might be supported, could break a lot of plugins that do not source their files from the `Files` collection, but use something like `glob` or other ways of enumerating files in the `docs_dir`. One might argue that those approaches are not idiomatic, but I've seen them more than often (and am guilty as well), so this change has quite some potential for frustration.

athackst:
I wouldn't advocate for changing the default from the 'docs' directory.  I think there are legitimate use cases for wanting to call mkdocs from root though. It's a feature that's been requested a fair amount through mkdocs history.

oprypin:
@athackst We will get this working for sure. I would like to propose a different "explicit" approach though - see https://github.com/mkdocs/mkdocs/pull/3519

oprypin:
My last conclusion on this was that the feature request is good but the exact change I wanted to be done differently.

So, I was using this PR as if it's an **issue** and the actual **pull request** I wanted to merge is https://github.com/mkdocs/mkdocs/pull/3519.
However that one got stuck - didn't go into 1.6.0 and its status is unknown.
I'll leave it because I think it's the best direction overall.

To avoid confusion, maybe this pull request should be closed and an issue opened instead.

MatthewSteen:
This feature would be useful for documentation-only sites and personal sites deployed with pages. 

This can be done with https://github.com/jupyter-book/jupyter-book, which might be a good resource. Here's an example...
- https://gitlab.com/MatthewSteen/usa-politics

Note, another related feature (if not already implemented) would be to allow the `index.md` to be in the root dir (or `readme.md`).

PR #3456, opened 2023-11-01
Quoted pip arguments containing square brackets

vedranmiletic:
This is strictly necessary when using non-bash shells (e.g. tcsh).


oprypin:
Thanks, this is needed. Wouldn't single quotes be even more safe though?

oprypin (approved):
Thanks

vedranmiletic:
That was quick. Thank you!

Issue #3418, opened 2023-10-10
Security contact for vulnerability report


jorgectf:
Hello,

The [GitHub Security Lab team](https://securitylab.github.com/) has found a potential vulnerability in your project. Please create a [Security Advisory](https://docs.github.com/en/code-security/repository-security-advisories/creating-a-repository-security-advisory) and invite me in to further disclose and discuss the vulnerability details and potential fix. Alternatively, please add a [Security Policy](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) containing a security email address to send the details to, and/or enable [Public Vulnerability Reporting](https://docs.github.com/en/code-security/security-advisories/repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository) so we can submit the details.

If you prefer to contact us by email, please reach out to [securitylab@github.com](mailto:securitylab@github.com) with reference to `GHSL-2023-208`.

Thank you,

GitHub Security Lab

oprypin:
Thanks. "Private vulnerability reporting" enabled. Is that sufficient?

jorgectf:
Thank you @oprypin, that is great.

PR #3465, opened 2023-11-11
Stop executing YAML tags for mkdocs_theme.yml, warn about third-party projects

oprypin:
@jorgectf 
This is the outcome of https://github.com/mkdocs/mkdocs/issues/3418
Which warned about arbitrary code execution via installing a theme and that theme having malicious YAML tags.
I found that there are a lot more ways for arbitrary code execution if you install an untrusted package and import it.
As such, this change only slightly reduces the attack surface, but at the same time I don't really consider it a security issue in the first place, and I don't think it'd be useful for users to see all versions of MkDocs as having a security vulnerability and then also see that this change fixes it (but actually not really).

But I agree:
* executing YAML tags in `mkdocs_theme.yml` *is* pointless.
* the documentation *could* definitely use a note about the dangers of third-party packages.

So, the changes are in accordance to that:

- Stop allowing arbitrary YAML tags for mkdocs_theme.yml
- Docs: warn about installing third-party plugins

Bonus change:

- Cache the result of the function `get_themes()`

### DIFF
--- a/docs/dev-guide/plugins.md
+++ b/docs/dev-guide/plugins.md
@@ -15,6 +15,8 @@ appropriate package name and install it using `pip`:
 pip install mkdocs-foo-plugin
 ```
 
+WARNING: Installing an MkDocs plugin means installing a Python package and executing any code that the author has put in there. So, exercise the usual caution; there's no attempt at sandboxing.
+
 Once a plugin has been successfully installed, it is ready to use. It just needs
 to be [enabled](#using-plugins) in the configuration file. The [Catalog]
 repository has a large ranked list of plugins that you can install and use.
--- a/docs/user-guide/choosing-your-theme.md
+++ b/docs/user-guide/choosing-your-theme.md
@@ -196,6 +196,8 @@ theme supports the following options:
 
 A list of third party themes can be found at the [community wiki] page and [the ranked catalog][catalog]. If you have created your own, please add them there.
 
+WARNING: Installing an MkDocs theme means installing a Python package and executing any code that the author has put in there. So, exercise the usual caution; there's no attempt at sandboxing.
+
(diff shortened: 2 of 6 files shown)


docs/dev-guide/plugins.md
```diff
@@ -15,6 +15,8 @@ appropriate package name and install it using `pip`:
 pip install mkdocs-foo-plugin
 ```
 
+WARNING: Installing an MkDocs plugin means installing a Python package and executing any code that the author has put in there. So, exercise the usual caution; there's no attempt at sandboxing.
```
jorgectf, line 18:
```suggestion
WARNING: By installing an MkDocs plugin you are installing a Python package and executing any code embedded by the author. There's no attempt at sandboxing, so exercise the usual caution.
```

docs/user-guide/choosing-your-theme.md
```diff
@@ -196,6 +196,8 @@ theme supports the following options:
 
 A list of third party themes can be found at the [community wiki] page and [the ranked catalog][catalog]. If you have created your own, please add them there.
 
+WARNING: Installing an MkDocs theme means installing a Python package and executing any code that the author has put in there. So, exercise the usual caution; there's no attempt at sandboxing.
```
jorgectf, line 199:
```suggestion
WARNING: By installing an MkDocs plugin you are installing a Python package and executing any code embedded by the author. There's no attempt at sandboxing, so exercise the usual caution.
```

jorgectf (approved):
Thank you for this PR, especially for warning about the perils of using external plugins.

docs/dev-guide/plugins.md
oprypin: Sorry, I like my wording better :s

PR #3468, opened 2023-11-15
feat(reproducible-builds): preserve existing sitemap.xml.gz mtime if sitemap content did not change

ultrabug:
This is a proposal to address #3460 and #2689

(yes I know, tests are broken, this PR is just to demonstrate what I had in mind)

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import gzip
+import hashlib
 import logging
 import os
 import time
+from pathlib import PurePath
 from typing import TYPE_CHECKING, Sequence
 from urllib.parse import urljoin, urlsplit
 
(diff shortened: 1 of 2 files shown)


oprypin:
Hmm I think this misses the main usecases of reproducible builds.

What people usually deal with:

1. I have a CI job or a distro package with 100% constant inputs, it should produce 100% constant outputs.

2. I am validating that nothing in a site changed after a refactor, so I run 2 builds into 2 separate directories and compare them.

(Note: items 1 and 2 are already possible to satisfy by setting `SOURCE_DATE_EPOCH`)

3. I am annoyed that every single time I push to gh-pages, the file `sitemap.xml.gz` changes for no reason.

What people do **not** ask for:

4. I just built a site and I have the directory saved locally, the next rebuild should look at that file and preserve it.

---

This PR addresses only item 4 and partly item 3 if people deploy locally rather than from CI.

oprypin:
Closing per this and https://github.com/mkdocs/mkdocs/pull/3460#issuecomment-1836793168

PR #3477, opened 2023-11-20
Move code to external mkdocs-get-deps dependency

oprypin:
The "get-deps" command will still be provided inside MkDocs like before (and is not deprecated at all) but is now implemented in a separate repository and can be used directly from there as well.
https://github.com/mkdocs/get-deps

This separation of code was done because obtaining just the "get-deps" part with very few dependencies can be useful: one can run it in a main environment but then install actual MkDocs with dependencies in a virtual environment.

I have also already applied this extracted small library "mkdocs-get-deps" for another use case, which I'm very proud of:
### https://github.com/mkdocs/hatch-mkdocs :new: 

In the near future I will write a new recommended guide for how to manage dependencies for a MkDocs site. This is only one part of it though.


pawamoy (approved):

sillydan1:
This creates a problem with `pip<=22.0.2` (the version shipped with [ubuntu:22.04](https://hub.docker.com/_/ubuntu) is `22.02`)

Upgrading the pip version to `pip>=22.0.3` fixes the issue.

Recreation `requirements.txt`:
```
mkdocs
mkdocs-material
```

## Recreation steps:
Launch a new ubuntu 22.04 container:
```
docker run --rm -it -v .:/code -w /code ubuntu:22.04
```
Inside the container
```
apt update
apt install python3 python3-pip
python3 -m pip install mkdocs mkdocs-material
```

Observe that https://github.com/pypa/pip/issues/10851 occurs

pawamoy:
Hi @sillydan1, I'm not sure to understand why this resolved pip bug from 2022 is relevant to this PR.

oprypin:
Thanks for the report @sillydan1 . It should probably be opened as a new issue.

@pawamoy The problem, as was already described, is that users of Ubuntu 22.04 out of the box get an error for this command:
```bash
pip install mkdocs mkdocs-material
```

I confirm the finding, this was my repro:

```bash
virtualenv venv
venv/bin/pip install pip==22.0.2
venv/bin/pip install mkdocs mkdocs-material

```

<details><summary>Error:</summary>

```
ERROR: Exception:
  File "/tmp/a/venv/lib/python3.11/site-packages/pip/_internal/cli/base_command.py", line 165, in exc_logging_wrapper
    status = run_func(*args)
             ^^^^^^^^^^^^^^^
  File "/tmp/a/venv/lib/python3.11/site-packages/pip/_internal/cli/req_command.py", line 205, in wrapper
    return func(self, options, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/a/venv/lib/python3.11/site-packages/pip/_internal/commands/install.py", line 389, in run
    to_install = resolver.get_installation_order(requirement_set)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/a/venv/lib/python3.11/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 188, in get_installation_order
    weights = get_topological_weights(
              ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/a/venv/lib/python3.11/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 276, in get_topological_weights
    assert len(weights) == expected_node_count
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

</details>

The cause is the intersection of the bug in pip and the fact that mkdocs-material pinned the version of MkDocs.

https://github.com/squidfunk/mkdocs-material/blob/5f40a5337492c8e8578142bd39b29d938af73947/requirements.txt#L24

The best solution would be to unpin versions there. In the world of Python packaging, it is hostile to pin versions in a library, because it just doesn't work well. See my comment that I wrote for a different project:
* https://github.com/Guts/mkdocs-rss-plugin/issues/137#issuecomment-1271901582

Some workarounds in the meantime

1. ```bash
   pip install -U pip
   ```
2. ```bash
   pip install 'mkdocs' 'mkdocs-material<=9.5.2'
   ```
3. ```bash
   pip install 'mkdocs<=1.5.3' 'mkdocs-material'
   ```

oprypin:
Also @sillydan1 I'm not sure how you determined that this PR causes it. I was not able to confirm or deny that, but I think it's not related. I think the bug is purely due to the version number conflict. In any case definitely should be a separate issue.

sillydan1:
Sorry for reviving a closed PR. You're right, this should be an issue by itself - I will create one.

> I'm not sure how you determined that this PR causes it.

The only changes that happened to my CI build was that `mkdocs` was upgraded to `v1.6.0` and that `mkdocs-get-deps` would be installed as a consequence. In retrospect, it's just a red herring and the issue you mentioned @oprypin with `mkdocs-material` version pinning is the root cause.

Issue #2248, opened 2020-12-08
Dark mode support based on OS choice


Fryguy:
I'm not sure if it's already done, but I couldn't find anything in the docs.  If not, I think this would be a cool feature.  I did see `nav_style` in the theme support, but that seems to only allow the header to be switched at compile time, as opposed to styling the page at runtime based on user/OS choice.

out0xb2:
### Case Study: favicon
Can it be changed depending upon the browser's light or dark theme?

**GitHub.com** 
Open a browser in light theme, and the GitHub.com favicon is black on transparent.  Switch to dark theme, and GitHub's favicon switches to white on transparent.  It looks... good, is visible

**mkdocs.org**
In light theme, browse to https://www.mkdocs.org/ , observe the favicon is black on transparent.  Switch to dark theme, refresh, the MkDocs favicon is remains black on dark, barely discernable...

waylan:
Personally, I detest dark mode for anything. But that is just my personal opinion. While I won't force that opinion on others, I have absolutely no motivation to spend time on this. If anyone wants to work on this, they are welcome to provide a PR. I would be willing to review the PR. 

Any PR needs to address all of the following concerns:

- All built-in themes should be addressed in the same manner. In other words, a single PR should address both built-in themes, not just the one you care about (although... see next item).
- The `readthedocs` theme is a clone of the Sphinx theme of the same name. Any dark mode in this theme much reflect the upstream theme. I haven't checked if the upstream theme even supports a dark mode, so this may be moot. I'll leave it to someone else to do the research and provide links to back it up.
- The `mkdocs` theme is based on bootstrap/bootswatch. There may be a simple way to switch between light/dark there. However, we provide various overrides and additional non-bootswatch CSS/images which would need to all be addressed as well.
- Any switch to enable/disable dark mode should be a [theme config setting][1]. No changes need to be made to MkDocs outside of the theme. As an aside, any third-party themes should be able to implement dark mode without any changes to MkDocs at all.
- Before accepting any PR, someone will need to promise to continue to support dark mode. In other words, they will need to remain available to address bug reports and adapt future changes to MkDocs to work with dark mode.

[1]: https://www.mkdocs.org/user-guide/custom-themes/#theme-configuration

squidfunk:
[...]

Issue #3189, opened 2023-04-20
[FR] Alternative setting for highlightjs


redtide:
Current:
```yml
highlightjs: true
hljs_languages:
- ...
hljs_style: ...
```
Alternative:
```yml
highlightjs:
  languages:
     builtin:
     - cpp
     extra (or custom or whatever):
     - relative or abs url to customlanguage.min.js
  style: ...
```
or if #2171 would be (finally) a thing (or not, needs some strategy to support also only 1 mode by defining 1 of  2):
```yml
highlightjs:
  ...
  style:
    light: same as hljs_style
    dark: dark hljs theme to use on dark mode
```

oprypin:
That's a fine suggestion for a refactor (just makes it looks nice), but what's the practical benefit that would justify incompatibility or duplication?

tomchristie:
Really it'd be neater if highlightjs (or whatever other alternative) was included through overriding the template `styles`/`libs` blocks. Perhaps something to move towards as part of #3190.

redtide:
> Really it'd be neater if highlightjs (or whatever other alternative) was included through overriding the template `styles`/`libs` blocks, rather than through config. Perhaps something to move towards as part of #3190.

in general you say "less integrations, more modularity", I totally agree that that's a better choice.

PR #3493, opened 2023-12-01
Update mkdocs theme to bootstrap 5.3 and add support for dark mode.

waylan:
This is related to #2248 but does not resolve that issue. That issue is requesting the option for the reader to select dark/light from the website itself. However, this only implements the option to select light/dark mode at build time from within the config file. Presumably, #2248 could be built on top of this.

As of version 5.3, Bootstrap supports color modes and provides light and dark modes out-of-the-box. The `mkdocs` theme has always used the [Bootswatch Cerulean](https://bootswatch.com/cerulean/) theme. Therefore, this updates the CSS and JS to version 5.3 of Bootstrap/Bootswatch and also updates most of the color definitions in the local CSS (`base.css` and `extera.css`) to use bootstrap variables (for example, admonitions now use Bootstrap colors). That way, switching between light and dark modes provides consistent results.

The challenge is with code blocks and the default syntax highlighter used by the theme ([Highlight.js](https://highlightjs.org/). Highlight.js defines isn't own background colors for code blocks in its CSS files. And Highlight.js provides 496 themes, not all of which have light/dark counterparts. This presents multiple challenges.

First of all, the background colors do not exactly match Bootstrap's colors. However, so long as every code block is processed by Highlight.js, then they all look the same. That means the `nohighlight` code blocks will get a different background color. Therefore, `text` should be used as the 'language' to maintain consistently styled code blocks (I have included an update to the docs in this regard).

At some point in the past, the `hljs_style` config option was added to the `mkdocs` theme. This option accepts the string name of any of Highlight.js' many themes and the template simply inserts that string into the URL pointing to the CSS file hosted in a CDN. In fact, this is why the theme uses a CDN. Otherwise, all 496 CSS files would need to be included with the theme files on MkDocs. We could pick two Highlight.js themes (one each for light and dark modes) and host them locally, but that would be a breaking change. Therefore, whenever the color mode is changed, the theme needs to be changed to match. There are a few different ways we could do this:

1. Require the user to change both the `color_mode` and `hljs_style` independently. One would have no effect on the other.
2. Provide a second "dark" config option (i.e.: `hljs_style_dark`) which defines the corresponding dark theme for Highlight.js. Then the relevant option could be used depending on `color_mode`.
3. We could abandon Highlight.js and instead rely on the `codehilite` Markdown extension, which would require hosting our own CSS styles.
4. Something else I haven't considered...

While option 1 would be the simplest, is would likely result in many complaints from users as the two config options would presumably be connected. Therefore, I have implemented option 2 here. Option 3 would be a breaking change, but would be much easier to work with in any future implementation of #2248.

Any input would be appreciated on the above issues.

Finally, I will note that as of Bootstrap version 5, jQuery was dropped as a dependency. Bootstrap now uses plain Javascript with the exception of Popper.js (which is included in the Bootstrap bundle) for popups/modals/etc. However, most of the Javascript in the `mkdocs` theme still uses jQuery. Presumably we could refactor all of the code in `base.js` to not depend on jQuery, which would have the protentional to reduce load times, etc. That said, many users likely have custom JavaScript of their own (in extra.js`) which expects jQuery, so removing jQuery completely would be a breaking change. Therefore, I have left that for a separate use to deal with. The only changes I made to the Javascript code were to update calls to Bootstrap's API.


waylan:
See #3189 which suggests a refactor of the Highlight.js config options.

redtide:
> However, this only implements the option to select light/dark mode at build time from within the config file.

Most websites I visited and also some desktops give the possibility to choose dark/light versions of a theme at runtime, I don't see why to set it at build time instead making a dark version of the existing themes and choose if to use the system/dark/light color mode at runtime like the former do.
Other than that, if it's possible to use js (I was told it's possible via css but I wasn't able to make it work) you can also [swap highlightjs at runtime ](https://redtide.github.io/bootstrap5-test-page/#code-block) (I used MKDocs to build this site).

redtide:
> 1. Require the user to change both the `color_mode` and `hljs_style` independently. One would have no effect on the other.

I think the user should choose the hljs style in the config and the color mode at runtime (e.g. via dropdown "Light/Dark/Auto" menu).
The hljs theme might be only light or dark, so swapping will do nothing, otherwise work as expected, so yes, make hljs depending on color mode if both are set, mkdocs should not care if they match with the site theme.

> 2. Provide a second "dark" config option (i.e.: `hljs_style_dark`) which defines the corresponding dark theme for Highlight.js. Then the relevant option could be used depending on `color_mode`.

Which IMO is the better option (see also below).

> 3 We could abandon Highlight.js and instead rely on the `codehilite` Markdown extension, which would require hosting our own CSS styles.

Personally I dislike this, but it's just my own opinion.

> Finally, I will note that as of Bootstrap version 5, jQuery was dropped as a dependency. Bootstrap now uses plain Javascript with the exception of Popper.js

IIRC popper.js is needed for the search and keyboard dialogs (I had troubles with them when making the test theme I linked above because of missing popper.js/jquery), so maybe jquery could be made optional if not using them?

> See https://github.com/mkdocs/mkdocs/issues/3189 which suggests a refactor of the Highlight.js config options.

That was a suggestion on how I did it, more or less, in my own theme but I guess it's not necessary a refactor if your `hljs_style_dark` addition idea would do the job without breaking compatibility.

waylan:
> > However, this only implements the option to select light/dark mode at build time from within the config file.
> 
> Most websites I visited and also some desktops give the possibility to choose dark/light versions of a theme at runtime, I don't see why to set it at build time instead making a dark version of the existing themes and choose if to use the system/dark/light color mode at runtime like the former do.

Well, we have to start with something. Do we start with light as the default or dark? Different users may have different opinions about that. So we are giving them the option to choose. But once that is complete, then we can add the JS to do the in-browser switching. So its not like I'm wasting my time implementing a build-time config option. It will need to exist anyway. The in-browser switch is simply an additional add-on which can be added later.

> I think the user should choose the hljs style in the config and the color mode at runtime (e.g. via dropdown "Light/Dark/Auto" menu).
> The hljs theme might be only light or dark, so swapping will do nothing, otherwise work as expected, so yes, make hljs depending on color mode if both are set, mkdocs should not care if they match with the site theme.

I considered that. But then I tried it and wow is it hideous. Leaving the two unconnected would be a disservice to our users IMO.

> > See #3189 which suggests a refactor of the Highlight.js config options.
> 
> That was a suggestion on how I did it, more or less, in my own theme but I guess it's not necessary a refactor if your `hljs_style_dark` addition idea would do the job without breaking compatibility.

Personally, I actually prefer your suggestion in #3189. However, it breaks compatibility. Ideally, when a user upgrades to a new version of MkDocs, they should not need to make edits to their site config before building the first time. Sure, they may want to make edits to take advantage of new features, but their site should generally render the same without any edits. Of course, at times that is unavoidable, but we want to be cautious about when we choose to do so.

> > 3 We could abandon Highlight.js and instead rely on the `codehilite` Markdown extension, which would require hosting our own CSS styles.
> 
> Personally I dislike this, but it's just my own opinion.

What is it that you dislike? Hosting our own CSS? The specific styles provided by Pygments? The lack of ability to chose from multiple options?

redtide:
> Well, we have to start with something. Do we start with light as the default or dark? Different users may have different opinions about that. So we are giving them the option to choose. But once that is complete, then we can add the JS to do the in-browser switching. So its not like I'm wasting my time implementing a build-time config option. It will need to exist anyway. The in-browser switch is simply an additional add-on which can be added later.

I would say OK for light, you can check how it works on my test page link. If you support a system theme preference it will switch automatically so who cares about the default?
Sites around (and so desktops) have light theme as default AFAICS.

> I considered that. But then I tried it and wow is it hideous. Leaving the two unconnected would be a disservice to our users IMO.

Is that terrible in the link I posted? Again, I don't think MKDocs should care about the style an user sets. HLJS provides a lot, and you can't match all of them, so can't you just make your own css and let the user eventually override it?
I'm not sure what you mean with unconnected, if to switch both the site theme with the HLJS together is the same thing I prefer, if unconnected in the sense because not same colors that IDK what you can do about it, make a "MKDocs style" for both looks to me a waste of time to make it and also maintain it, users should choose what they like. As start I used a "GitHub dark palette" for mine, which seems to me working well with the HLJS GH versions.

> Personally, I actually prefer your suggestion in https://github.com/mkdocs/mkdocs/issues/3189. However, it breaks compatibility. Ideally, when a user upgrades to a new version of MkDocs, they should not need to make edits to their site config before building the first time.

I haven't got this, can't you still support the old method and also add a new one (not great but I can't see a better option to avoid breakage), no?

> What is it that you dislike? Hosting our own CSS? The specific styles provided by Pygments? The lack of ability to chose from multiple options?

Abandon HLJS and use codehilite. In general I like (and use) some of those extensions but there are some, like syntax highlight that I don't, but it's my opinion and by chance I started to use HLJS before trying to switch to MKDocs from Jekill, where previously I was using Google Prettify.

waylan:
I added an `auto` mode, which follows the local system setting of the browser.

```yaml
theme:
    name: mkdocs
    color_mode: auto
```

Color Mode switch

The trick was working out how to have both light and dark Highlight.js themes available for live switching (see 6485ff9). As I understand it, browsers do not download a `disabled` `<link>`, so no unnecessary resources are used with this solution. 

One can also do either `color_mode: light` or `color_mode: dark` which builds with the one option only and does not do any switching (although, it could with a JavaScript call to `setColorMode`). Now it should be easy to add on-page switching.

oprypin:
Wow that's great stuff!

redtide:
Good job, looks great!
And by looking at commits and the preview I like the coding and visual styles!

waylan:
I have now added a user option which is enabled with `user_color_mode_toggle: true` in the config.

Untitled

However, I needed to update fontawesome to version 6 as v3 did not have all of the appropriate icons. Personally, I would have likely used Bootstrap Icons, but I'm trying to maintain the existing dependencies for backward compatibility.

So now this one PR updates both bootstrap and fontawesome to the latest releases and adds a feature (color modes). I'm wondering if this is too much for one PR. If so, let me know and I'll break it up into a few.

Regardless, I still need to update the documentation.

oprypin:
This seems like the perfect opportunity to upgrade fontawesome. But yes, if we intend to squash this pull request into one, that would be too much. It might be prudent to switch the approach for this pull request to developing 1 ready stack of commits to be merged without squashing. Then a few other changes such as "Add user_color_mode_toggle" could also get their own commit. Anyway just seems much easier than multiple pull requests. Also I'm not sure when we'll actually get to merging and releasing this; I certainly would want to merge the fontawesome change right before merging the theme change (which the "stacked commits" approach would solve perfectly).

redtide:
> This seems like the perfect opportunity to upgrade fontawesome

Not sure if you know about [forkawesome](https://forkaweso.me/Fork-Awesome/icons/), might be another alternative in case?

waylan:
I believe this is ready to go. I have refactored the commits into 4 separate logical commits, which should be able to be committed as-is (not squashed). The last commit resolves #2248. See the commit messages and documentation updates for details.

docs/css/extra.css
```diff
@@ -69,7 +48,9 @@ body.homepage>div.container>div.row>div.col-md-9 {
 
 .doc-object {
     padding-left: 10px;
-    border-left: 4px solid rgba(230, 230, 230);
+    /* csslint ignore:start */
```
oprypin: So csslint just straight up errors on every variable, ok... It should just be dropped. It is an abandoned project.

I will drop it for sure. You can do as you like:
* either drop csslint already and rebase onto that (commit 31c033babfa4a988f9064f963ceafc65dda37105 does so, you can `git fetch` it if you like)
* or keep the ignore comments and I'll drop them later

mkdocs/themes/mkdocs/base.html
```diff
@@ -65,16 +69,16 @@
     </head>
 
     <body{% if page and page.is_homepage %} class="homepage"{% endif %}>
-        <div class="navbar fixed-top navbar-expand-lg navbar-{% if config.theme.nav_style == "light" %}light{% else %}dark{% endif %} bg-{{ config.theme.nav_style }}">
-            <div class="container">
+        <div class="navbar fixed-top navbar-expand-lg bg-{{ config.theme.nav_style }}" data-bs-theme="{% if config.theme.nav_style == 'light' %}light{% else %}dark{% endif %}">
+            <div class="container-fluid">
```
oprypin: I'm noticing that this makes the navigation take up the whole page width. I think it's better to keep it like it was - aligned with the page content. More in line with the original idea of the theme. **Also** Bootstrap already has a wider page width anyway.

```suggestion
            <div class="container">
```

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -42,32 +44,41 @@ body > .container {
...
+    border: 1px solid var(--bs-secondary-border-subtle);
+    /* csslint ignore:end */
     border-radius: 4px;
     margin: 20px auto 30px auto;
 }
 
 h1 {
-    color: #444;
+    /* csslint ignore:start */
+    color: var(--bs-secondary-text-emphasis);
```
oprypin: I think this color is not appropriate, it's really grey. On Bootstrap's own site, headings are just the default black color :thinking: 

Based on this page I was confused
https://getbootstrap.com/docs/5.3/customize/css-variables/
It says `  --bs-heading-color: inherit;` but .. that's not the actual value.
Maybe we just do `inherit` directly - that looks good.

```suggestion
    color: inherit;
```

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -42,32 +44,41 @@ body > .container {
...
+    /* csslint ignore:end */
     font-weight: 400;
     font-size: 42px;
 }
 
 h2, h3, h4, h5, h6 {
-    color: #444;
+    /* csslint ignore:start */
+    color: var(--bs-secondary-text-emphasis);
```
oprypin:
```suggestion
    color: inherit;
```

mkdocs/themes/mkdocs/css/bootstrap.min.css.map
oprypin, line 1: Maybe we shouldn't add the map file because it refers to a file "dist/cerulean/bootstrap.css" which does not exist so is not usable

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -226,24 +240,110 @@ blockquote {
...
+.admonition.example h1, .admonition.example h2, .admonition.example h3,
+.admonition.example h4, .admonition.example h5, .admonition.example h6,
+details.example h1, details.example h2, details.example h3,
+details.example h4, details.example h5, details.example h6 {
+    color: var(--bs-light-text-emphasis);
+}
+
+.admonition.example a, details.example a {
+    color: var(--bs-light-text-emphasis);
+}
```
oprypin: I think changing the color of links is invalid, should probably be dropped for all admonition types.

Normally they are blue, this makes them grey

```suggestion
```

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -226,24 +240,110 @@ blockquote {
...
+    background-color: var(--bs-light-bg-subtle);
+    border-color: var(--bs-light-border-subtle);
+}
+
+.admonition.example h1, .admonition.example h2, .admonition.example h3,
+.admonition.example h4, .admonition.example h5, .admonition.example h6,
+details.example h1, details.example h2, details.example h3,
+details.example h4, details.example h5, details.example h6 {
+    color: var(--bs-light-text-emphasis);
+}
```
oprypin: We could keep the "example" admonition type out of the theme; I would like it to remain an ugly purple for my own documentation-writing purposes.

But how about applying this style as a fallback for all admonitions, which is currently missing?

```suggestion
.admonition, details {
    color: var(--bs-light-text-emphasis);
    background-color: var(--bs-light-bg-subtle);
    border-color: var(--bs-light-border-subtle);
}

.admonition h1, .admonition h2, .admonition h3,
.admonition h4, .admonition h5, .admonition h6,
details h1, details h2, details h3,
details h4, details h5, details h6 {
    color: var(--bs-light-text-emphasis);
}
```

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -226,24 +240,110 @@ blockquote {
...
+.admonition.new h1, .admonition.new h2, .admonition.new h3,
+.admonition.new h4, .admonition.new h5, .admonition.new h6,
+details.new h1, details.new h2, details.new h3,
+details.new h4, details.new h5, details.new h6 {
+    color: var(--bs-success-text-emphasis);
+}
+
+.admonition.new a, details.new a {
+    color: var(--bs-success-text-emphasis);
+}
```
oprypin: Could we perhaps keep the `new` admonition type **out** of the theme? There is no other theme that also has this style. I just wanted this for my own purposes.

```suggestion
```

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -42,32 +44,41 @@ body > .container {
...
 
 pre, .rst-content tt {
     max-width: 100%;
-    background: #fff;
-    border: solid 1px #e1e4e5;
-    color: #333;
+    /* csslint ignore:start */
+    background: var(--bs-secondary-bg-subtle);
+    border: solid 1px var(--bs-secondary-border-subtle);
+    color: var(--bs-secondary-text-emphasis);
```
oprypin: The colors have super little contrast - just 3 shades of grey. This also deviates from the prior theme.

And it deviates very far from highlightjs which has high-contrast colors - in places where it does not kick in, colors will randomly be grey.

I have devised that these colors match the prior colors very well, as well as being close to highlightjs:

```suggestion
    background-color: var(--bs-body-bg);
    border: solid 1px var(--bs-border-color);
    color: var(--bs-body-color);
```


And it seems that you've noticed this effect yourself, because you replaced `nohighlight` with `text` to avoid this

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -77,16 +88,17 @@ code.code-large, .rst-content tt.code-large {
 
 code {
     padding: 2px 5px;
-    background: #fff;
-    border: solid 1px #e1e4e5;
-    color: #333;
+    /* csslint ignore:start */
+    background-color: var(--bs-secondary-bg-subtle);
+    border: solid 1px var(--bs-secondary-border-subtle);
+    color: var(--bs-secondary-text-emphasis);
```
oprypin: And let's match it for `code`.

**But** with subtler background because it's too noisy inside admonitions otherwise

```suggestion
    background-color: rgba(var(--bs-body-bg-rgb), 0.75);
    border: solid 1px var(--bs-border-color);
    color: var(--bs-body-color);
```

mkdocs/themes/mkdocs/css/base.css
```diff
@@ -226,24 +240,110 @@ blockquote {
     text-align: left;
```
oprypin, line 222: (detached comment)

Hmm headerlinks look quite ugly - the chain logo is always underlined now, so it's hard to see that it's even supposed to be a chain. 

Also it's more heavily misaligned from center.

How about this addition:

```
.headerlink {
    text-decoration: none;
    vertical-align: middle;
}
```

docs/css/extra.css
```diff
@@ -20,17 +20,6 @@ div.admonition.block>.admonition-title {
     display: none;
 }
```
oprypin, line 22: To bring back my admonitions :)

```suggestion

.admonition.new, details.new {
    color: var(--bs-success-text-emphasis);
    background-color: var(--bs-success-bg-subtle);
    border-color: var(--bs-success-border-subtle);
}
.admonition.example, details.example {
    color: var(--bs-info-text-emphasis);
    background-color: var(--bs-info-bg-subtle);
    border-color: var(--bs-info-border-subtle);
}

```

oprypin (commented):
Publishing my comments for commit "Update mkdocs theme to Bootstrap/Bootswatch 5.3.2"

All my edits are available as a commit 17168b9cdb5d3d75204551126254eceb168f9660

Sorry for big delay :(

mkdocs/themes/mkdocs/js/base.js
```diff
@@ -18,8 +18,72 @@ function applyTopPadding() {
...
+                    setColorMode(mql.matches ? 'dark' : 'light');
+                    mql.addEventListener('change', onSystemColorSchemeChange);
+                } else {
+                    setColorMode(mode);
+                    mql.removeEventListener('change', onSystemColorSchemeChange);
+                }
+                updateModeToggle(mode);
+            });
+        });
+
```
oprypin: Let's not put this code into `$(document).ready`. That causes the site to always briefly blink white before turning to dark. Just put all new code at the end of this file at the top level. It is fine because the `<script>` is put after the `body`, so all elements are available.

mkdocs/themes/mkdocs/js/base.js
```diff
@@ -18,8 +18,72 @@ function applyTopPadding() {
...
+function updateModeToggle(mode) {
+    // Update icon and toggle checkmarks of color mode selector.
+    var menu = document.getElementById('theme-menu');
+    document.querySelectorAll('[data-bs-theme-value]')
+       .forEach(function(toggle) {
+            if (mode == toggle.getAttribute('data-bs-theme-value')) {
+                toggle.lastElementChild.classList.remove('d-none');
+                menu.firstElementChild.setAttribute('class', toggle.firstElementChild.getAttribute('class'));
+            } else {
+                toggle.lastElementChild.classList.add('d-none');
```
oprypin: If we're populating `aria-pressed`, let's keep it up to date
```suggestion
                toggle.setAttribute('aria-pressed', 'true');
                toggle.lastElementChild.classList.remove('d-none');
                menu.firstElementChild.setAttribute('class', toggle.firstElementChild.getAttribute('class'));
            } else {
                toggle.setAttribute('aria-pressed', 'false');
                toggle.lastElementChild.classList.add('d-none');
```

mkdocs/themes/mkdocs/js/base.js
```diff
@@ -18,8 +18,72 @@ function applyTopPadding() {
...
+        mql.addEventListener('change', onSystemColorSchemeChange);
+    } else {
+        setColorMode(defaultMode);
+        updateModeToggle(defaultMode);
+    }
+
+    document.querySelectorAll('[data-bs-theme-value]')
+       .forEach(function(toggle) {
+            toggle.addEventListener('click', function (e) {
+                var mode = e.target.getAttribute('data-bs-theme-value');
```
oprypin: Important: need to refer to the original button element, otherwise you can click on the text `span` of the button and then it thinks nothing is clicked.

```suggestion
                var mode = e.currentTarget.getAttribute('data-bs-theme-value');
```

mkdocs/themes/mkdocs/js/base.js
```diff
@@ -18,8 +18,72 @@ function applyTopPadding() {
...
+                localStorage.setItem('colormode', mode);
```
oprypin: Let's name this one 'mkdocs-colormode'. Some namespacing...

mkdocs/themes/mkdocs/base.html
```diff
@@ -164,6 +168,37 @@
...
+                                </li>
+                                <li>
+                                  <button type="button" class="dropdown-item d-flex align-items-center" data-bs-theme-value="dark" aria-pressed="{% if config.theme.color_mode == 'dark' %}true{% else %}false{% endif %}">
+                                    <i class="fa-solid fa-moon fa-fw"></i>
+                                    <span class="ms-2">Dark</span>
+                                    <i class="fa-solid fa-check ms-auto{% if config.theme.color_mode != 'dark' %} d-none{% endif %}"></i>
+                                  </button>
+                                </li>
+                                <li>
+                                  <button type="button" class="dropdown-item d-flex align-items-center" data-bs-theme-value="auto" aria-pressed="{% if config.theme.color_mode == 'auto' %}true{% else %}false{% endif %}">
```
oprypin: Don't need to set type="button", it's default

```suggestion
                                  <button class="dropdown-item d-flex align-items-center" data-bs-theme-value="light" aria-pressed="{% if config.theme.color_mode == 'light' %}true{% else %}false{% endif %}">
                                    <i class="fa-solid fa-sun fa-fw"></i>
                                    <span class="ms-2">Light</span>
                                    <i class="fa-solid fa-check ms-auto{% if config.theme.color_mode != 'light' %} d-none{% endif %}"></i>
                                  </button>
                                </li>
                                <li>
                                  <button class="dropdown-item d-flex align-items-center" data-bs-theme-value="dark" aria-pressed="{% if config.theme.color_mode == 'dark' %}true{% else %}false{% endif %}">
                                    <i class="fa-solid fa-moon fa-fw"></i>
                                    <span class="ms-2">Dark</span>
                                    <i class="fa-solid fa-check ms-auto{% if config.theme.color_mode != 'dark' %} d-none{% endif %}"></i>
                                  </button>
                                </li>
                                <li>
                                  <button class="dropdown-item d-flex align-items-center" data-bs-theme-value="auto" aria-pressed="{% if config.theme.color_mode == 'auto' %}true{% else %}false{% endif %}">
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -87,29 +91,29 @@
                         <ul class="nav navbar-nav">
                         {%- for nav_item in nav %}
                         {%- if nav_item.children %}
-                            <li class="dropdown{% if nav_item.active %} active{% endif %}">
-                                <a href="#" class="nav-link dropdown-toggle" data-toggle="dropdown">{{ nav_item.title }} <b class="caret"></b></a>
+                            <li class="nav-item dropdown" data-bs-theme="{% if config.theme.nav_style == 'dark' %}dark{% else %}light{% endif %}">
```
oprypin:
```suggestion
                            <li class="nav-item dropdown">
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -164,6 +168,37 @@
                             </li>
                         {%- endif %}
                       {%- endblock %}
+                      {%- if config.theme.user_color_mode_toggle %}
+                            <li class="nav-item dropdown" data-bs-theme="{% if config.theme.nav_style == 'dark' %}dark{% else %}light{% endif %}">
```
oprypin:
```suggestion
                            <li class="nav-item dropdown">
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -164,6 +168,37 @@
...
+                              <a class="nav-link dropdown-toggle d-flex align-items-center" href="#" id="theme-menu" aria-expanded="false" data-bs-toggle="dropdown" data-bs-display="static" aria-label="Toggle theme">
+                                <i class="fa-solid fa-circle-half-stroke fa-fw"></i>
+                                <span class="d-lg-none ms-2">Toggle theme</span>
+                              </a>
```
oprypin: The dropdown is off-center vertically compared to other nav items. Let's drop `d-flex` from it, that's what causes it. Also change to button.

```suggestion
                              <button id="theme-menu" aria-expanded="false" data-bs-toggle="dropdown" data-bs-display="static" aria-label="Toggle theme" class="nav-link dropdown-toggle">
                                <i class="fa-solid fa-circle-half-stroke fa-fw"></i>
                                <span class="d-lg-none ms-2">Toggle theme</span>
                              </button>
```

oprypin (commented):
Publishing comments for all color mode selector changes

Final state available in commit cf13620f8d6c9c12612126f023ebe2d53f99fbec

If you want to be able to apply it directly, the 'dark mode' and 'user_color_mode_toggle' commits would have to be squashed first. Otherwise you have to recreate and apply my comments manually

oprypin:
Great stuff though. These are all the comments I have :)

mkdocs/themes/mkdocs/base.html
```diff
@@ -65,16 +69,16 @@
     </head>
 
     <body{% if page and page.is_homepage %} class="homepage"{% endif %}>
-        <div class="navbar fixed-top navbar-expand-lg navbar-{% if config.theme.nav_style == "light" %}light{% else %}dark{% endif %} bg-{{ config.theme.nav_style }}">
-            <div class="container">
+        <div class="navbar fixed-top navbar-expand-lg bg-{{ config.theme.nav_style }}" data-bs-theme="{% if config.theme.nav_style == 'light' %}light{% else %}dark{% endif %}">
```
oprypin: Hm this comment somehow got lost from the previous batch.

I see this `{% if config.theme.nav_style` in three places and I finally get that it tries to achieve inverted colors for the navbar. But these will not work, they aren't being updated dynamically.
Instead let's keep using the deprecated "navbar-dark" class. (That deprecation seems misguided for sure...)

```suggestion
        <div class="navbar fixed-top navbar-expand-lg navbar-{% if config.theme.nav_style == "light" %}light{% else %}dark{% endif %} bg-{{ config.theme.nav_style }}">
```

mkdocs/themes/mkdocs/css/base.css
waylan: I was just trying to maintain what was there using Bootstrap variables. The existing style defined a gray color so I used the closest variable color I could find. But if you want to make the change that is fine with me.

mkdocs/themes/mkdocs/css/base.css
waylan: I debated this with myself. But as it turns out, Bootstrap does change link colors within alerts. As I matched the admonition styles to Bootstrap alerts, I included the link styles. But I can back that out if you want.

mkdocs/themes/mkdocs/css/base.css
waylan: Good call having a fallback. Hadn't considered that for some reason.

mkdocs/themes/mkdocs/css/base.css
waylan: So the `new` and `example` admonition styles were previously defined in `extra.css` which unique to MkDocs' documentation. I moved them to the theme because I though they made sense there. After all, new and example are the sorts of blocks one might want in technical documentation, which is MkDocs primary use case. And, they are using two of the otherwise unused alert styles defined by Bootstrap. Guess I don't see why they can't be included with the theme. But if that is what you really want I can move them back. 😕

mkdocs/themes/mkdocs/css/base.css
waylan: > And it seems that you've noticed this effect yourself, because you replaced `nohighlight` with `text` to avoid this

I did, but I would have actually preferred the colors I chose over highlight.js colors. In fact, I actually considered setting `!important` on those styles so they override highlight.js. But then I realized a user may prefer a different highlight.js theme which those colors would not work with.

mkdocs/themes/mkdocs/base.html
waylan: Ah, I noticed that off-center, but couldn't figure out what was causing it. So thanks.

Regarding changing to `button`, why? All the other nav items use `a` links.

mkdocs/themes/mkdocs/base.html
waylan: > But these will not work, they aren't being updated dynamically.

I know they aren't updated dynamically. `nav_style` is not affected by `color_mode` at all and I am not making any reference to color mode here. So there is nothing to be updated dynamically. My suggested change implements the existing behavior with Bootstraps new API as near as we can.

You seem to be suggesting we revert to Bootstraps now-deprecated old API. Why?

waylan:
Thanks for the feedback. I'll need to sort through it all when I have some time, which may be a while.

docs/css/extra.css
oprypin: Dropped csslint on master and rebased

mkdocs/themes/mkdocs/base.html
oprypin: Applied this suggestion - reverted this in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/css/base.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

docs/css/extra.css
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

mkdocs/themes/mkdocs/js/base.js
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/a83615554940f681b9c292b21de78e84616017c8

mkdocs/themes/mkdocs/js/base.js
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/a83615554940f681b9c292b21de78e84616017c8

mkdocs/themes/mkdocs/js/base.js
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/a83615554940f681b9c292b21de78e84616017c8

mkdocs/themes/mkdocs/js/base.js
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/a83615554940f681b9c292b21de78e84616017c8

mkdocs/themes/mkdocs/base.html
oprypin: Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/a83615554940f681b9c292b21de78e84616017c8

mkdocs/themes/mkdocs/base.html
oprypin: Why button: because unlike other items, it's just a button that doesn't lead anywhere :)
Good for semantic browsing / accessibility.

Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/a83615554940f681b9c292b21de78e84616017c8

mkdocs/themes/mkdocs/base.html
oprypin: Hmm something was causing problems for me but now I'm not fully sure that it was this. But sorry, I probably won't have the energy to come back to this and try to understand why this complex logic always works correctly, when the simpler deprecated setup works in a more straightforward way. *And* is also what the preexisting code does - basically I'm reducing the diff. We can reconsider this edit in the future and double-check everything again from a known working state.

Applied this suggestion in https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae

oprypin:
I have made a new branch that is ready to be force-pushed to here. (Shall I just do that?) It contains all my suggested edits applied. Maybe it will be easier for you to just re-review it in the final shape, rather than checking my individual edits?

This is https://github.com/mkdocs/mkdocs/compare/master...oprypin:mkdocs:themeup

1.  "Update mkdocs theme to Bootstrap/Bootswatch 5.3.2"
    * Before: https://github.com/mkdocs/mkdocs/commit/b80b22b31429a07afd307fec59804b262f202c7f
    * After: https://github.com/mkdocs/mkdocs/commit/351b87939ddc61bc767e2b869e2a2673fe9af5ae
      (applied all my code suggestions, rebased without csslint)
1. "Update FontAwesome to version 6"
    * Before: https://github.com/mkdocs/mkdocs/commit/70e38273b7c9f28858c63f398a57292bdea3056a
    * After: https://github.com/mkdocs/mkdocs/commit/694a6028f649de6881ddc13496077856dcb3abc6
      (rebased without csslint)
1. "Add support for a 'dark mode' to mkdocs theme" + "Add 'user_color_mode_toggle' to mkdocs theme"
    * Before: https://github.com/mkdocs/mkdocs/commit/7265192995df75c232913fecb40553bb9b300f6e + https://github.com/mkdocs/mkdocs/commit/02564cc03cd7cc1567a264198e2c3bf81f9bd077
    * After: https://github.com/mkdocs/mkdocs/commit/a83615554940f681b9c292b21de78e84616017c8
       (squashed 2 commits, rebased with updated markdownlint style, applied all my code suggestions)

waylan:
If that is what you want to do. For the reasons stated in my individual responses, I actually think your changes are not improvements, but I don't have the time or energy to try to convince anyone of that. I also don't have the time to review your changes and likely won't for some months. So, you can merge your changes with this or just close this and open a new PR. Whatever works best for you.

dagnelies:
This really looks great! Thanks for the effort you put into this. 👍 
I also think that it should be merged ASAP because it's a great update, even if there are some details where your opinions diverge. 
After all, many of such small details could still be updated later on . As long as it's "stable", it's all good! Happy Easter! 🥚 🐰 🥚

oprypin:
I plan to merge this for the 1.6.0 release **with my edits**, I hope that's OK.

I have checked the theme extensively for any deficiencies in Firefox/Linux, and I think any edits are for the better.

https://github.com/mkdocs/mkdocs/pull/3631

oprypin (approved):

Fryguy:
I hadn't seen this work happening in the background after #2248 was closed, but this is awesome! Great work @waylan!

squidfunk:
It's awesome to see dark mode coming to MkDocs' core themes! I checked the linked preview in #3631, and observed that the dark mode solution suffers from occasional white flashing, especially on very long pages (see attached video). I tried to find the code that implements dark mode to learn why, but to no avail, so my guess is that the selected preference is checked and applied "too late", i.e., at the end of the `body`. When there's a lot of content, the end of the parser sees the end of the `body` only after a while has passed, but the page already has partially rendered, which explains the flashing.

https://github.com/mkdocs/mkdocs/assets/932156/602965f5-bee9-44e0-9623-cfb2f5661b45

The solution would be to move the code that decides what mode is to be applied to the beginning of `body`.

mkdocs/themes/mkdocs/js/base.js
oprypin: For the record, per this thread, the delay for dark mode was even longer prior to this edit. Need to further improve this, though.

squidfunk:
Working on #3649, I found some potential regressions:

- The search modal's close button looks different
- Search result titles are underlined
- The table of contents collapse button on mobile looks different

__Screenshots__

Bildschirm­foto 2024-04-17 um 08 39 46
Bildschirm­foto 2024-04-17 um 08 39 18

waylan:
There are a couple factors here. First, we updated to a more recent version of Bootstrap and Bootstrap changed the style of some items. The different look to the close and collapse buttons could be an intentional change on Bootstrap's part. That needs to be verified. The second factor is that we are not always using Bootstrap's defaults. IIRC links were changed in some way which caused our customizations to stop working. It could be that I missed the change to search results. In any event, this PR is already merged, seems like it would be appropriate to open a new issue to track this stuff. Also, unfortunately I do not have any time to work on this myself.

oprypin:
* Yeah that close button does look really weird 🤔

* All links are underlined, that's indeed changed

* Yes the nav collapse button looks different, not sure what's up with that but it's not bad

Thanks for checking

PR #3498, opened 2023-12-02
Clarify error about taken port in `serve`; allow to open a browser

### LINKED ISSUE #3496: With `mkdocs serve` switch to another port if the current port is busy
It is possible that a couple of instances of the MkDocs web server are running at the same time, for legitimate reasons. The second time, the server won't start, unless a different port is used. The solution is to use the `-a`/`--dev-addr` option with `mkdocs-server`, but it is tedious.

Programmatically checking that the port is free is easy, and this could be automated (rather than failing).

### Additional points
  1. For the port numbers, a simple approach would be to increment the port from 8000 up, let's say on a range from 8000 to 8019 (I do not see why one would want to open more than a couple; it would rapidly become difficult for the user to keep up with their own mind pointers/terminal sessions).
  2. Needless to say, a server that self-updates (after detecting a file change) should keep the same port.

See also discussion in #3494

oprypin:
- Add a flag to override the `serve` port
- Add diagnostics when `mkdocs serve` fails to bind a port

-  Closes #3496

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -1,15 +1,21 @@
 from __future__ import annotations
 
+import hashlib
+import json
 import logging
+import os.path
 import shutil
 import tempfile
+import urllib.request
 from os.path import isdir, isfile, join
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Mapping
+from urllib.error import HTTPError
 from urllib.parse import urlsplit
 
 from mkdocs.commands.build import build
 from mkdocs.config import load_config
-from mkdocs.livereload import LiveReloadServer, _serve_url
+from mkdocs.exceptions import Abort
+from mkdocs.livereload import LiveReloadServer, ServerBindError, _serve_url
 
 if TYPE_CHECKING:
     from mkdocs.config.defaults import MkDocsConfig
@@ -61,6 +67,12 @@ def serve(
     if port is None:
         port = DEFAULT_PORT
 
+    origin_info = dict(
+        path=os.path.dirname(config.config_file_path) if config.config_file_path else os.getcwd(),
+        site_name=config.site_name,
+        site_url=config.site_url,
+    )
+
     mount_path = urlsplit(config.site_url or '/').path
     config.site_url = serve_url = _serve_url(host, port, mount_path)
 
@@ -73,7 +85,12 @@ def serve(
         build(config, serve_url=None if is_clean else serve_url, dirty=is_dirty)
 
     server = LiveReloadServer(
-        builder=builder, host=host, port=port, root=site_dir, mount_path=mount_path
+        builder=builder,
+        host=host,
+        port=port,
+        root=site_dir,
+        mount_path=mount_path,
+        origin_info=origin_info,
     )
 
     def error_handler(code) -> bytes | None:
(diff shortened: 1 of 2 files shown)


oprypin:
I dropped port randomization and trying multiple ports, compared to https://github.com/mkdocs/mkdocs/pull/3497. I think it can actually be too confusing.

mkdocs/__main__.py
```diff
@@ -248,7 +252,9 @@ def cli():
 
 
 @cli.command(name="serve")
-@click.option('-a', '--dev-addr', help=dev_addr_help, metavar='<IP:PORT>')
+@click.option('-a', '--dev-addr', help=dev_addr_help, metavar='IP:PORT')
```
squidfunk, line 255: So, please don't get me wrong, but IMHO, this is not a very consistent design. You can specify a port via `--dev-addr`, but also via `--port`? That's two ways of achieving the same thing.

oprypin: OK what do you suggest?

squidfunk: Either remove `--port` or deprecate and split up `--dev-addrs` into `--host` and `--port`. I'm not sure what you're aiming here for however, because `--dev-addr` is working perfectly fine.

pawamoy:
What happens if I permanently set the port to 8010 with `serve_port: 8010` in the config file, and this port is already in use when I run `mkdocs serve`? Will I be able to override it with `-p 8020` or `-a 127.0.0.1:8020`?

pawamoy:
I wonder if this really brings value. Choosing a different ports among different projects will be cumbersome, and will not play well with project templates. Also it caters to a (I'd say) small fraction of users who run multiple live-reload servers concurrently. And IMO is only useful when you also use `-o` to automatically open the tab in the browser. So many users will likely keep using 8000 by default. Besides, even if you permanently sets ports 8007, 8020, 8055 and 8080 on *your* projects, nothing guarantees that *other* projects you contribute to won't use the same ports. In that case you'll again fall back to overriding with `-p`. The first few ports (and the multiples of 10) will probably be over-used across all projects.

I feel like anything more than a default port plus a way to override it doesn't bring any real guarantee or benefit, and only increases cognitive overhead and/or maintenance work. I do note that this is an opt-in feature, so I won't push against it.

oprypin:
Yea the feature kinda falls apart easily. A compromise where nobody will bother trying to benefit from it

mkdocs/__main__.py
pawamoy: I would find a `-p` option really convenient. I literally *never* change the default host, and if I had to, it would probably be in a container setup where I would set it to 0.0.0.0 and never change it again. Whereas in (rare) cases where I serve multiple sites, I need to change the port, and having to re-specify the default host 127.0.0.1 is a bit annoying.

So I wouldn't mind if `--dev-addr` was deprecated in favor of two `--host` and `--port` options. However I'm not the only MkDocs user and I fear it would be quite disruptive to deprecate `--dev-addr` for, again, rarely happening situations. Maybe they could both (dev-attr and host+port) live together, forever, as mutually exclusive options?

squidfunk:
> I wonder if this really brings value. 

I feel that we're wasting our time discussing something that provides so little value as opposed to other open issues that would need more attention. This won't add any new functionality besides what's already achievable.

oprypin:
For you perhaps time is money. And for me, not spending time on one interesting issue doesn't mean I'll be more likely to spend time on another issue (actually often quite the opposite)

fralau:
> For you perhaps time is money. And for me, not spending time on one interesting issue doesn't mean I'll be more likely to spend time on another issue (actually often quite the opposite)

That's a key point 🙂, for occasional or volunteer contributors. Both viewpoints exist and complement each other.

Of course, I am beating the drum for my own subjects of interest; though I can see and appreciate the other person's viewpoint.

I any case if`mkdocs` could assume that the default host was localhost (`mkdocs serve -a 8001`), I would consider myself completely vindicated!

PR #3504, opened 2023-12-03
Help? Just adding `import markdown.htmlparser` breaks CI

oprypin:
I'm trying to develop some feature and among other things it needs this import. But just adding the import entirely breaks tests, despite `markdown.htmlparser` definitely existing. And I have no idea why or how to reproduce it. Happens only on GitHub Actions on all platforms.


oprypin:
Ah I figured it out :facepalm: - the minimum required version of Markdown for this is 3.3 but I had an older version configured for the "minimum requirements" test

PR #3522, opened 2023-12-17
Exclude tests from the wheel distribution

oprypin:
I can't think of any other project that includes tests


pawamoy (approved):
Same experience, I think it's common to include them in sdists and exclude them from wheels.

PR #3525, opened 2023-12-21
Remove mention of site_url being required option

Andre601:
The Getting started page mentions `site_url` being a required option. This can't be true however, since making a new MkDocs project using `mkdocs new` only generates a `mkdocs.yml` file containing the `site_name` option, which works fine when using the serve and build commands, meaning site_url is optional.

This is further confirmed by the fact that the Configuration page mentions that only site_name is required and all other options being optional.

This PR removes the `site_url: https://example.com/` from the code snippets while also updating the NOTE box about this option being required, by removing it altogether.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -117,7 +117,7 @@ Now change the configuration file to alter how the documentation is displayed by
 changing the theme. Edit the `mkdocs.yml` file and add a [`theme`][theme] setting:
 
 ```yaml
-: MkLorum
+site_name: MkLorum
 nav:
   - Home: index.md
   - About: about.md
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -123,8 +117,7 @@ Now change the configuration file to alter how the documentation is displayed by
 changing the theme. Edit the `mkdocs.yml` file and add a [`theme`][theme] setting:
 
 ```yaml
-site_name: MkLorum
-site_url: https://example.com/
+: MkLorum
```
oprypin: There's some error in this edit

Andre601: fixed

oprypin (approved):
Thanks

PR #3526, opened 2023-12-21
Don't trigger rebuilds for docs in exclude_docs

athackst:


oprypin:
Hi. This pull request will not be accepted at this time, it needs discussion for the nuance.
It is a non-obvious, and thus insidious, breaking change. For example, for "mkdocs-awesome-pages-plugin", people surely rely on the fact that the site rebuilds whenever they edit a ".pages" file.
At the very least someone will need to investigate whether that plugin is the only common use case for hidden files or not. Usually I'm the one doing such investigations but I'm not putting that on my plate at the moment.

athackst:
Ok, if it bubbles up your priority queue, feel free to re-open!

PR #3533, opened 2023-12-30
Expose a `page.content_title` attribute

### LINKED ISSUE #3532: Consider changing the precedence of title sources
Specifically, on this page https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data I would prefer that the level 1 headers take priority over the navigation name, or an option to do so.

I'm realizing that deriving the title from the navigation sidebar does not provide a good user experience and I would like to avoid having to add meta-content to every page. I will use a real example: https://hatch.pypa.io/1.9/config/static-analysis/

The sidebar already represents clearly defined sections and in this case the entire sidebar relates to config so it would be very verbose to add the word "configuration" to every name.

Here is the full(ish) picture showing the desired title as the level 1 header:

image

A consequence is that when you scroll down with Material for MkDocs the title on top is now no longer specific:

image

As well as the social card that shows up when you post a link to a page in Slack or Twitter or Discord:

slack-imgs

oprypin:
This is only a specific sub-case of `title` but may make more sense to directly use in some cases.

* Closes #3532

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -52,9 +52,6 @@ class Page(StructureItem):
         )
 
         # Placeholders to be filled in later in the build process.
-        self.markdown = None
-        self._title_from_render: str | None = None
-        self.content = None
         self.toc = []  # type: ignore
         self.meta = {}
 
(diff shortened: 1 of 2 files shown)


squidfunk:
This was closed without comment. @oprypin could you please explain why?

ofek:
Is this feature blocked on #3357 or are they uncoupled?

oprypin:
Uncoupled

ofek:
Can this be merged?

oprypin:
I didn't merge this because not sure if next maintainers will want to own this code.

PR #3546, opened 2024-01-10
Ensure page titles from rendered Markdown do not contain any markup

### LINKED ISSUE #3357: Header and navigation sidebars display incorrectly arbitrary markup

waylan:
While the `_ExtractTitleTreeprocessor` already extracts text from the etree, postprocessors may add more markup. Therefore, markupsafe's `striptags`  is used to ensure all markup is removed. Conveniently, `markupsafe` is already a dependency and available. Fixes #3357.

### DIFF
--- a/mkdocs/tests/structure/page_tests.py
+++ b/mkdocs/tests/structure/page_tests.py
@@ -366,7 +366,24 @@ class PageTests(unittest.TestCase):
         pg = Page(None, fl, cfg)
         pg.read_source(cfg)
         pg.render(cfg, fl)
-        self.assertEqual(pg.title, '*Hello &mdash; beautiful world')
+        self.assertEqual(pg.title, '*Hello — beautiful world')
+
+    _RAW_CONTENT = dedent(
+        '''
+        # Hello <span>world</span>
+
+        Hi.
+        '''
+    )
+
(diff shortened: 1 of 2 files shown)


waylan:
Hmm, the failing lint check seems nonsensical to me. Not sure what to do there.

ofek:
The suggestion is correct because that is third party package and currently it is in the standard library block.

mkdocs/tests/structure/page_tests.py
```diff
@@ -366,7 +366,24 @@ def test_page_title_from_markdown_strip_formatting(self, docs_dir):
         pg = Page(None, fl, cfg)
         pg.read_source(cfg)
         pg.render(cfg, fl)
-        self.assertEqual(pg.title, '*Hello &mdash; beautiful world')
+        self.assertEqual(pg.title, '*Hello — beautiful world')
```
waylan, line 369: I'm not convinced this behavior is desired here. I expect we do not want HTML entities to be unescaped. If so, we will need to clone the striptags function and eliminate the call to `unescape`. See [this comment](https://github.com/mkdocs/mkdocs/issues/3357#issuecomment-1887767130) for more.

PR #3550, opened 2024-01-16
Fetch remote inherited config via http

garryod:
Allows inherited configurations, specified via the `INHERIT` entry in `mkdocs.yml` to be retrieved from a remote server via HTTP(S). See #3552 for rationale

E.g.
```yaml
INHERIT: https://example.com/parent-config.yml
```

Where a scheme cannot be parsed out of the URL the entry is treated as a local file, meaning this change is backwards compatible (except in the instance where a user has a file named `http://something`)

### DIFF
--- a/mkdocs/tests/utils/utils_tests.py
+++ b/mkdocs/tests/utils/utils_tests.py
@@ -23,6 +23,16 @@ deep1:
     deep2-1:
         deep3-1: replaced
 """
+BASEHTTPYML = """
+INHERIT: http://localhost:8181/parent.yml
+foo: bar
+baz:
+    sub1: replaced
+    sub3: new
+deep1:
+    deep2-1:
+        deep3-1: replaced
+"""
 PARENTYML = """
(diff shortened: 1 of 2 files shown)


pawamoy:
Hello, thanks for your contribution. Has this been discussed prior to this PR?

garryod:
Hi Timothée, it hasn't, I wasn't exactly sure if it would be considered a "large" change. Where would be the best place to discuss the change?

pawamoy:
I see, the word "large" in contributing.md might not be the best choice. Your changes here are not large indeed, in terms of lines of code, but they are impactful nonetheless. Please open a new issue where you describe in details *why* you think MkDocs should be able to inherit configuration using HTTP(S) URLs :slightly_smiling_face:

garryod:
> Please open a new issue where you describe in details _why_ you think MkDocs should be able to inherit configuration using HTTP(S) URLs

Done, see #3552

PR #3556, opened 2024-01-21
Fix behavior overriding `self.meta` on `read_source` and add tests

### LINKED ISSUE #3555: [Change Request] Allow plugins to edit metadata before `on_page_markdown`
Hello.

I was building a plugin which tracks git history of a file **on non-metadata lines**. I wanted to track only content edit history, not metadata edit history. But I realized `self.meta` is overrided by `meta.get_date` in following lines;

https://github.com/mkdocs/mkdocs/blob/7186f4ce9c88e7c445ef90fee85c065ca297a7c8/mkdocs/structure/pages.py#L202-L215

Therefore I request `mkdocs` to change this behavior. I can upload pull requests myself if you want.

McDic:
Fixes #3555 .

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -19,6 +19,7 @@ from mkdocs import utils
 from mkdocs.structure import StructureItem
 from mkdocs.structure.toc import get_toc
 from mkdocs.utils import _removesuffix, get_build_date, get_markdown_title, meta, weak_property
+from mkdocs.utils.misc import dict_merge_inplace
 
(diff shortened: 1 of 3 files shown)


McDic:
Not sure why 3.10 test is failed in my local machine.

I looked at [plugin_tests.py](https://github.com/mkdocs/mkdocs/blob/master/mkdocs/tests/plugin_tests.py) and it seems that test file is not about actually running the plugin so I did not touch it. If I have to touch it, can you please elaborate how should I do this?

Issue #3213, opened 2023-05-06
Ampersands seen in  search results


jidanni:
I do not own this site. All I know is I did a search and I saw these raw ampersands in the results.
![Screenshot_20230506_115822_Chrome Beta.jpg](https://user-images.githubusercontent.com/1230959/236598854-8a610877-2b74-4421-9c47-b8fb40233035.jpg)

The site was https://vespucci.io/ and the search was for "copy".

theoneGUI:
It looks like this is the offending line. I'm new here, so I don't know the ins and outs of the program. 

https://github.com/mkdocs/mkdocs/blob/25f4ea153322357c31d15fe601c4ea743028aefa/mkdocs/contrib/search/templates/search/main.js#L32

I was able to replicate the problem by adding an ampersand to a title of a markdown document. I don't know if removing the 
`escapeHtml()` around `title` is _the_ answer, but it's an answer.

I see that the search result is getting formatted as if it's getting inserted directly into HTML with escapes, but is the `formatResult()` necessary for `title` and `summary` if it's putting the escape code in the document?

oprypin:
#3560 is an attempted fix but it's also rather confusing. Instead maybe this simpler fix could be applied.

```diff
diff --git a/mkdocs/contrib/search/search_index.py b/mkdocs/contrib/search/search_index.py
index 770f24fe2..c90eeb0dc 100644
--- a/mkdocs/contrib/search/search_index.py
+++ b/mkdocs/contrib/search/search_index.py
@@ -8,6 +8,8 @@ import subprocess
 from html.parser import HTMLParser
 from typing import TYPE_CHECKING
 
+from markupsafe import Markup
+
 if TYPE_CHECKING:
     from mkdocs.structure.pages import Page
     from mkdocs.structure.toc import AnchorLink, TableOfContents
@@ -90,7 +92,9 @@ class SearchIndex:
 
         text = ' '.join(section.text) if self.config['indexing'] == 'full' else ''
         if toc_item is not None:
-            self._add_entry(title=toc_item.title, text=text, loc=abs_url + toc_item.url)
+            self._add_entry(
+                title=Markup(toc_item.title).striptags(), text=text, loc=abs_url + toc_item.url
+            )
 
     def generate_search_index(self) -> str:
         """Python to json conversion."""
```

This fixes escaping in the search index and causes no other changes for MkDocs' own site at least

```diff
[...]

PR #3560, opened 2024-01-26
Sanitizing search entry titles

waylan:
MkDocs' search plugin indexes page sections by iterating over the sections of a page and adding an entry to the index for each section. It currently uses the title of a section's table of contents (TOC) entry as the title of the search index entry. As a reminder, the TOC of the page is provided by Markdown's `toc` extension which sanitizes a heading's content into plain text. In the typical case, it still works fine.

However, the Markdown `toc` extension later added support for the `data-toc-label` attribute, which a user could assign to a heading (h1-6) element (presumably via the `attr_list` extension). The content of that attribute would then be used as the label/title of the TOC item. The problem is that this allows users to introduce unsanitized (not plain) text into the title of a search index entry, which confuses search in weird ways.

It is also possible for a third-party plugin to alter the TOC of a page by inserting markup into a TOC entry. For example, the [`show_symbol_type_toc`](https://mkdocstrings.github.io/python/usage/configuration/headings/?h=show_symbol_type_toc#show_symbol_type_toc) option of mkdocsstrings-python specifically inserts spans into TOC titles. Again, this confuses search. Not only do search terms not return results as expected, but when the results are returned (due to a match in the body text), the title of the search result contains escaped HTML and is potentially confusing to the user (see Python-Markdown/markdown#1432).

Note that the same (or similar) issues exist with the Page title, but that is the subject of #3357 and will presumably be addressed with a fix to that larger issue. Therefore, I am not address that here. Although, it is possible we could run the Page title through `striptags` as extra insurance. So I could add that here if desired.

Given the above issues, a change needs to be made to ensure search is passed a sanitized version of the title of each section.  There are at least 2 ways this could be done:

1. Pass the TOC title through `striptags` and pass the plain text result to the search index.
2. Retrieve the actual heading (h1-6) element, and pass the plain text result to the search index.

While option 1 would be more consistent with current behavior, we could be indexing text that is different than what appears in the body of the page, which means section headings which should clearly be included in search results could be omitted (or visa-versa). Therefore, the search index should be passed a sanitized version of the actual heading, not the TOC title.

As it turns out, the existing code already strips tags from the headings of the document when building the search index, but then ignores those sanitized headings and uses the TOC title instead by using a section's `id` to associate the TOC title with the section. However, if we use the sanitized heading text, then we don't need to reference the TOC at all and any reference to it can be removed from the code.

Given the above reasoning, I have implemented option 2 in this PR.

Consider the following examples:

```markdown
# Some text {data-toc-label='different text'}
```

Both the previous behavior and option 1 above would be to index `different text`, and option 2 would index `Some text`. Clearly indexing `Some text` makes more sense in this instance.

```markdown
# Some text {data-toc-label='<span class="foo">Some text</span>'}
```

The previous behavior would be to index `<span class="foo">Some text</span>` and both options 1 and 2 above would index `Some text`.

```markdown
# `foo(bar=0)` {data-toc-label='foo'}
```

Both the current behavior and option 1 would index `foo`. The string `bar` would not be indexed at all. With option 2, the string `foo(bar=0)` would be indexed. But that isn't tokenized into separate words by the search index by default (in my testing the search term `foo(bar=0` was needed to get the entry returned as a search result). Of course, the [search.separator](https://www.mkdocs.org/user-guide/configuration/#separator) config option allows one to define their own set of word separator characters. That could potentially result in better search results, and maintains the unmodified title in the search results displayed to the user. That said, using the word separator alone may not be as effective as one might think. According to its developer, lunr.js is not particularly well suited to tokenizing code (see olivernn/lunr.js#424 for at least one discussion about that).

Therefore, I have added an additional new feature to my proposal:

Add an additional `keywords` field to the search index and add support for a `data-search-keywords` attribute which can be defined on a heading element. When the attribute is defined on a heading, the contents of that attribute are assigned to the `keywords` field in the entry in the search index. A heading without the attribute would have no keywords assigned to that section and would continue to behave as-is. Consider the following example:

```markdown
# `foo(bar=0)` {data-toc-label='foo' data-search-keywords='foo bar'}
```

With the proposed change, the string `foo bar` would be indexed as keywords for the above example, which would result in reasonable search results. Initially I used the keywords as the title for the search index entry, but that caused the keywords to be used as the title in returned search results, which would be confusing to a user. Therefore, adding a new keywords field seemed like the best way to address this.

I also discovered a minor bug where if a heading contained inline markup, then the entire text of the heading was not retained (which presumably went unnoticed for years because it wasn't actually being used). Specifically, each child element's text content would override/replace its previous sibling's and/or parent's text content. That bug has been fixed and a test added. Each child element's content is now appended to the the existing content of the `title`.

Finally, I added `boost=10` to the `keywords` field of the search index so that results with matching keywords would get sorted higher in the results. Although, in my limited testing, the boost doesn't seem to make much difference. I have a vague recollection that similar observations were made in the past about the ineffectiveness of boost. :shrug:

### DIFF
--- a/mkdocs/contrib/search/search_index.py
+++ b/mkdocs/contrib/search/search_index.py
@@ -31,12 +31,12 @@ class SearchIndex:
         self._entries: list[dict] = []
         self.config = config
 
-    def _add_entry(self, title: str | None, text: str, loc: str) -> None:
+    def _add_entry(self, title: str | None, text: str, keywords: str | None, loc: str) -> None:
         """A simple wrapper to add an entry, dropping bad characters."""
         text = text.replace('\u00a0', ' ')
         text = re.sub(r'[ \t\n\r\f\v]+', ' ', text.strip())
 
-        self._entries.append({'title': title, 'text': text, 'location': loc})
+        self._entries.append({'title': title, 'text': text, 'keywords': keywords, 'location': loc})
 
     def add_entry_from_context(self, page: Page) -> None:
         """
@@ -58,7 +58,7 @@ class SearchIndex:
 
         # Create an entry for the full page.
         text = parser.stripped_html.rstrip('\n') if self.config['indexing'] == 'full' else ''
-        self._add_entry(title=page.title, text=text, loc=url)
+        self._add_entry(title=page.title, text=text, keywords='', loc=url)
 
         if self.config['indexing'] in ['full', 'sections']:
             for section in parser.data:
@@ -72,7 +72,7 @@ class SearchIndex:
         create an entry in the index.
         """
         text = ' '.join(section.text) if self.config['indexing'] == 'full' else ''
-        self._add_entry(title=section.title, text=text, loc=f'{abs_url}#{section.id}')
+        self._add_entry(title=section.title, text=text, keywords=section.keywords, loc=f'{abs_url}#{section.id}')
 
     def generate_search_index(self) -> str:
         """Python to json conversion."""
@@ -104,7 +104,7 @@ class SearchIndex:
             if haslunrpy:
                 lunr_idx = lunr(
                     ref='location',
-                    fields=(dict(field_name='title', boost=10), 'text'),
+                    fields=('title', 'text', dict(field_name='keywords', boost=10)),
                     documents=self._entries,
                     languages=self.config['lang'],
                 )
@@ -132,16 +132,26 @@ class ContentSection:
         text: list[str] | None = None,
         id_: str | None = None,
         title: str | None = None,
+        keywords: str | None = None
     ) -> None:
         self.text = text or []
         self.id = id_
         self.title = title or ''
+        self.keywords = keywords or ''
 
     def __eq__(self, other):
-        return self.text == other.text and self.id == other.id and self.title == other.title
+        return (
+            self.text == other.text and
+            self.id == other.id and
+            self.title == other.title and
+            self.keywords == other.keywords
+        )
 
     def __repr__(self):
-        return f"{self.__class__.__name__}(text={self.text}, id='{self.id}', title='{self.title}')"
+        return (
+            f"{self.__class__.__name__}("
+            f"text={self.text}, id='{self.id}', title='{self.title}', keywords='{self.keywords}')"
+        )
 
 
 _HEADER_TAGS = tuple(f"h{x}" for x in range(1, 7))
@@ -182,8 +192,7 @@ class ContentParser(HTMLParser):
         self.section.id = attrs.get('id', None)
         if 'data-search-keywords' in attrs:
             # Override title with user defined search keywords
-            self.section.title = attrs['data-search-keywords']
-            self.is_header_tag = False
+            self.section.keywords = attrs['data-search-keywords']
         self.data.append(self.section)
 
     def handle_endtag(self, tag: str) -> None:
--- a/mkdocs/contrib/search/search_index.py
+++ b/mkdocs/contrib/search/search_index.py
@@ -140,6 +140,9 @@ class ContentSection:
     def __eq__(self, other):
         return self.text == other.text and self.id == other.id and self.title == other.title
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}(text={self.text}, id='{self.id}', title='{self.title}')"
+
 
 _HEADER_TAGS = tuple(f"h{x}" for x in range(1, 7))
 
--- a/mkdocs/contrib/search/search_index.py
+++ b/mkdocs/contrib/search/search_index.py
@@ -10,7 +10,6 @@ from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from mkdocs.structure.pages import Page
-    from mkdocs.structure.toc import AnchorLink, TableOfContents
 
 try:
     from lunr import lunr  # type: ignore
@@ -31,19 +30,6 @@ class SearchIndex:
     def __init__(self, **config) -> None:
         self._entries: list[dict] = []
         self.config = config
-
-    def _find_toc_by_id(self, toc, id_: str | None) -> AnchorLink | None:
-        """
-        Given a table of contents and HTML ID, iterate through
-        and return the matched item in the TOC.
-        """
-        for toc_item in toc:
-            if toc_item.id == id_:
-                return toc_item
-            toc_item_r = self._find_toc_by_id(toc_item.children, id_)
-            if toc_item_r is not None:
-                return toc_item_r
-        return None
 
     def _add_entry(self, title: str | None, text: str, loc: str) -> None:
         """A simple wrapper to add an entry, dropping bad characters."""
@@ -76,21 +62,17 @@ class SearchIndex:
 
         if self.config['indexing'] in ['full', 'sections']:
             for section in parser.data:
-                self.create_entry_for_section(section, page.toc, url)
+                self.create_entry_for_section(section, url)
 
     def create_entry_for_section(
-        self, section: ContentSection, toc: TableOfContents, abs_url: str
+        self, section: ContentSection, abs_url: str
     ) -> None:
         """
-        Given a section on the page, the table of contents and
-        the absolute url for the page create an entry in the
-        index.
+        Given a section of a page and the absolute url for the page
+        create an entry in the index.
         """
-        toc_item = self._find_toc_by_id(toc, section.id)
-
         text = ' '.join(section.text) if self.config['indexing'] == 'full' else ''
-        if toc_item is not None:
-            self._add_entry(title=toc_item.title, text=text, loc=abs_url + toc_item.url)
+        self._add_entry(title=section.title, text=text, loc=f'{abs_url}#{section.id}')
 
     def generate_search_index(self) -> str:
         """Python to json conversion."""
@@ -122,7 +104,7 @@ class SearchIndex:
             if haslunrpy:
                 lunr_idx = lunr(
                     ref='location',
-                    fields=('title', 'text'),
+                    fields=(dict(field_name='title', boost=10), 'text'),
                     documents=self._entries,
                     languages=self.config['lang'],
                 )
@@ -153,7 +135,7 @@ class ContentSection:
     ) -> None:
         self.text = text or []
         self.id = id_
-        self.title = title
+        self.title = title or ''
 
     def __eq__(self, other):
         return self.text == other.text and self.id == other.id and self.title == other.title
@@ -175,10 +157,17 @@ class ContentParser(HTMLParser):
         self.data: list[ContentSection] = []
         self.section: ContentSection | None = None
         self.is_header_tag = False
+        self.is_permalink = False
         self._stripped_html: list[str] = []
 
     def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None:
         """Called at the start of every HTML tag."""
+        attrs = dict(attrs)
+        # Check for permalink in header
+        if self.is_header_tag and tag == 'a' and 'class' in attrs and 'headerlink' in attrs['class']:
+            self.is_permalink = True
+            return
+
         # We only care about the opening tag for headings.
         if tag not in _HEADER_TAGS:
             return
@@ -187,14 +176,21 @@ class ContentParser(HTMLParser):
         # for it and assign the ID if it has one.
         self.is_header_tag = True
         self.section = ContentSection()
+        self.section.id = attrs.get('id', None)
+        if 'data-search-keywords' in attrs:
+            # Override title with user defined search keywords
+            self.section.title = attrs['data-search-keywords']
+            self.is_header_tag = False
         self.data.append(self.section)
-
-        for attr in attrs:
-            if attr[0] == "id":
-                self.section.id = attr[1]
 
     def handle_endtag(self, tag: str) -> None:
         """Called at the end of every HTML tag."""
+
+        # Check for permalinks
+        if self.is_permalink and tag == 'a':
+            self.is_permalink = False
+            return
+
         # We only care about the opening tag for headings.
         if tag not in _HEADER_TAGS:
             return
@@ -203,6 +199,11 @@ class ContentParser(HTMLParser):
 
     def handle_data(self, data: str) -> None:
         """Called for the text contents of each tag."""
+
+        # Do not retain permalink text.
+        if self.is_permalink:
+            return
+
         self._stripped_html.append(data)
 
         if self.section is None:
@@ -216,7 +217,7 @@ class ContentParser(HTMLParser):
         # Otherwise it is content of something under that header
         # section.
         if self.is_header_tag:
-            self.section.title = data
+            self.section.title = self.section.title + data
         else:
             self.section.text.append(data.rstrip('\n'))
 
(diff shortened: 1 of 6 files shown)


waylan:
I can't make out the issue with the 2 remaining mypy errors.

mkdocs/contrib/search/search_index.py
```diff
@@ -150,13 +132,26 @@ def __init__(
         text: list[str] | None = None,
         id_: str | None = None,
         title: str | None = None,
+        keywords: str | None = None,
     ) -> None:
         self.text = text or []
         self.id = id_
-        self.title = title
+        self.title = title or ''
+        self.keywords = keywords or ''
```
oprypin: Nit: seems conceptually more appropriate to keep `title` and `keywords` as `None`  if missing? Not sure though

mkdocs/contrib/search/search_index.py
```diff
@@ -150,13 +132,26 @@ def __init__(
...
+            self.text == other.text
+            and self.id == other.id
+            and self.title == other.title
+            and self.keywords == other.keywords
+        )
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}("
+            f"text={self.text}, id='{self.id}', title='{self.title}', keywords='{self.keywords}')"
```
oprypin:
```suggestion
            f"{type(self).__name__}("
            f"text={self.text!r}, id={self.id!r}, title={self.title!r}, keywords={self.keywords!r})"
```

mkdocs/contrib/search/search_index.py
```diff
@@ -175,10 +170,17 @@ def __init__(self, *args, **kwargs) -> None:
         self.data: list[ContentSection] = []
         self.section: ContentSection | None = None
         self.is_header_tag = False
+        self.is_permalink = False
         self._stripped_html: list[str] = []
 
     def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None:
         """Called at the start of every HTML tag."""
+        atts = dict(attrs)
```
oprypin: Creating a dict on every tag occurrence is likely to become a performance bottleneck. It should be created only if necessary.

mkdocs/contrib/search/search_index.py
```diff
@@ -216,7 +225,7 @@ def handle_data(self, data: str) -> None:
         # Otherwise it is content of something under that header
         # section.
         if self.is_header_tag:
-            self.section.title = data
+            self.section.title = self.section.title + data
```
oprypin: What is this "data" and why is it OK to concatenate these two strings? How do they each look that it's OK to concatenate them without even a space in between? Can you add a code comment?

mkdocs/contrib/search/search_index.py
```diff
@@ -175,10 +170,17 @@ def __init__(self, *args, **kwargs) -> None:
...
         self.section: ContentSection | None = None
         self.is_header_tag = False
+        self.is_permalink = False
         self._stripped_html: list[str] = []
 
     def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None:
         """Called at the start of every HTML tag."""
+        atts = dict(attrs)
+        # Check for permalink in header
+        if self.is_header_tag and tag == 'a' and 'headerlink' in (atts.get('class') or ''):
```
oprypin: 'class' is a string. `'headerlink' in 'class1 adfgfheaderlinkdhsgh class2'` is true, that's probably not intended.

oprypin (commented):
I have no knowledge of the search plugin so I'll mostly just trust you on this one. I did notice several things that stood out to me generally though.

mkdocs/contrib/search/search_index.py
waylan: Generally, I would agree. However, my understanding is that lunr.js expects to only accept strings. Therefore, while we accept `None`, we only use strings internally. I am open to changing this, but be aware that the existing behavior already worked this way. I simply maintained consistency with the new field (and cleanup up a few inconsistencies with the existing fields).

mkdocs/contrib/search/search_index.py
waylan: This is the Python Standard lib HTMLParser. Data is anything that is not part of a tag. So it includes all text and whitespace. For example, consider this HTML:

``` HTML
<p>
foo <span>bar</span> baz
</p>
```
This method would be called 3 times and in each respective call, `data` would contain the following values:

1. `\nfoo `
2. `bar`
3. ` baz\n`

Concatenate those together, and you get the plaintext content of the `p` element: `\nfoo bar baz\n`. Notice that whitespace is preserved. There is no need to add any.

Not sure why we need a code comment. This is standard use here.

mkdocs/contrib/search/search_index.py
waylan: Good call. We should definitely fix this.

mkdocs/contrib/search/search_index.py
waylan: By the way, this is the bug I mentioned at the end of my first comment. Previously, each call to the function would have overwritten the previous value so that the final value would be only ` baz\n`. This fixes that bug.

mkdocs/contrib/search/search_index.py
waylan: I agree. However, we need to retrieve one of the attributes in the first if statement. In other words, we need the dict before we know if we need it. I'm open to suggestions here.

waylan:
The current lint errors appear to be related to a new [markdownlint rule](https://github.com/DavidAnson/markdownlint/blob/main/doc/md055.md) that was [added](https://github.com/DavidAnson/markdownlint/commit/30d62f19acf237c3ec017e660960389881cbd06b) in the most [recent release](https://github.com/DavidAnson/markdownlint/blob/main/CHANGELOG.md#0330). Presumably the various tables in the documentation need to be cleanup up and/or the default for the new rule needs to be changed to match the style used. Regardless, that has no effect on the changes here and I am ignoring it in this PR.

mkdocs/contrib/search/search_index.py
waylan: Okay, I worked this out.

waylan:
Thanks for the feedback. I have made all of the changes... except for the one where we are meeting lunr.js' API. I think this is ready to go.

oprypin (approved):
This is good!

Just to note:

* I intend to merge this myself
* I intend to check https://github.com/mkdocs/mkdocs/pull/3546 first

oprypin (changes requested):
I've been looking at the "keywords" feature and I'm really having second thoughts on it.

1. I don't see how the changes in this pull request cause a need for this new feature. I don't think they do.
   For <code># &grave;foo(bar=0)&grave;</code> the behavior both before and after is to index `foo(bar=0)`.
   The fact that data-toc-label no longer affects anything is certainly a change, but I don't think it's a problem.

2. It bloats the size of search index files by adding `"keywords": ""` to every heading.

3. You're saying that "boost" doesn't really work anyway?

4. When using "material", the most widely used theme, people will be confused why this newly announced "keywords" feature doesn't work at all. That theme replaces the "search" plugin and simply will not have this feature.

oprypin:
Accordingly pushed a commit to revert the "keywords" feature. Let me know what you think.

waylan:
> 1. I don't see how the changes in this pull request cause a need for this new feature. I don't think they do.
>    For `` # `foo(bar=0)` `` the behavior both before and after is to index `foo(bar=0)`.
>    The fact that data-toc-label no longer affects anything is certainly a change, but I don't think it's a problem.

The problem is that without keywords the search term `foo` or even `foo bar` will not include the `` # `foo(bar=0)` `` section in the search results. The only search term which will get that section to be included in the search result is the search term `foo(bar=0`, which is not helpful to users. Keywords were added as a solution to that problem. With the keywords, then the search terms `foo` or `foo bar` will result in that section being included in the results.

> 3\. You're saying that "boost" doesn't really work anyway?

Correct. That has always been the case and is an issue with the upstream project. I had forgotten that when I added the boost. We can either remove it or leave it. Doesn't matter either way to me.

> 4\. When using "material", the most widely used theme, people will be confused why this newly announced "keywords" feature doesn't work at all. That theme replaces the "search" plugin and simply will not have this feature.

So we can't make MkDocs better because a third party addon doesn't make use of that part of it? That seems like a strange argument to me.

oprypin:
OK so there are several changes that this causes. They can be observed by building mkdocs' own site before & after.

Here's output from

```bash
diff -U3 <(jq <sitebefore/search/search_index.json) <(jq <siteafter/search/search_index.json) | delta --light
```

---

These diffs are for the positive:  - most importantly, it fixes the failure to escape angle brackets:

image

---

However, I think this diff is for the negative:

image

`data-toc-label` is indeed no longer used, as was intended by the author, but that affects mkdocstrings' results for the worse.  @pawamoy 

---

<details><summary>And there was also this one but it's a pretty easy fix, I'll push it now</summary>

image

</details>

---

With this, I am not confident that we really understand this change

oprypin:
Discussion can continue on https://github.com/mkdocs/mkdocs/issues/3213, I'll post another finding there

waylan:
> However, I think this diff is for the negative:

Yes, and that is why I added keywords. Lets take this diff:

``` diff
- documentation_pages
+ \n        documentation_pages(*, inclusion: Callable[[InclusionLevel], bool] = InclusionLevel.is_included) -> Sequence[File]\n\n
```

Both the before and after are bad for different reasons. Before is bad because that entry won't show up in a search for `InclusionLevel`, Callable` etc. After is also bad for the reasons explained in detail in previous comments. However, if a key of keywords are defined, then the entry will be returned for a search on any of they provided keywords. Quite frankly, I am tired of re-explaining this multiple times.

oprypin:
Here's what you get when searching for `InclusionLevel` with this change. (And it is not affected by my removal of the `keywords` feature, as it's entirely opt-in)

image

That is problematic for sure.

Even if it's an abuse of functionality by mkdocstrings. But it will need to be hashed out first.

---

Even though of course as you're saying in the current state it is difficult to find anything if you're searching for `InclusionLevel`

waylan:
> Here's what you get when searching for `InclusionLevel` with this change.

Let's run with that. Right now, today, without my proposed changes, on mkdocs.org, if you search for `InclusionLevel`, this is what you get:

MkDocs

My change clearly gives better results. I don't understand why you think that it is worse. In your screenshot, the first result has the heading text which exactly matches the heading text of the linked section in the docs. That is what I expect and what this PR is trying to accomplish. Yet, nowhere have you expressed the idea that you disagree that that is better. So I don't understand what it is you think I'm doing or what you expect the search results should look like. Because, to me, your screen shot proves that my fix does exactly what it is supposed to do.

And if mkdocstrings later adds support for keywords (by including search keywords in its output), then the search results would be even better.

pawamoy:
Thanks for tagging me!

> My change clearly gives better results.

I too find that @oprypin's screenshot looks good, even if the `title` values containing whitespace are a bit weird.

> And if mkdocstrings later adds support for keywords (by including search keywords in its output), then the search results would be even better.

Looks easy (defining the `data-search-keywords` HTML attribute), though with both the built-in `search` plugin and Material for MkDocs own `search plugin starting to diverge, I'm not sure how much work it will represent for mkdocstrings to support both. Tagging @squidfunk just to makes sure he gets a chance to learn about this keywords feature.

squidfunk:
@pawamoy I'm not sure I understand what `keywords` is supposed to be for, and how it is used and presented in the search. Something similar was proposed for Material for MkDocs own search plugin in https://github.com/squidfunk/mkdocs-material/issues/3174, and we're considering implementing it, albeit we've not yet spec'd out how we want the DX to be. For instance:

- https://github.com/squidfunk/mkdocs-material/issues/3174#issuecomment-1452564246 – here, keywords are basically like aliases, so not rendered
- https://github.com/squidfunk/mkdocs-material/issues/3174#issuecomment-1488398702 – here, keywords are used in metadata, albeit presentation is not clear

We'll probably allow both, because the new search will allow to configure what metadata is searchable and eventually, how it is presented. We still have way to go, but we're getting there! Tracked in https://github.com/squidfunk/mkdocs-material/issues/6307. Also note that we are considering moving the search plugin out of Material for MkDocs in the mid-term, and providing a unified search plugin for all themes with a nice UI and UX. This would however be a second step ☺️

waylan:
> I'm not sure I understand what `keywords` is supposed to be for, and how it is used and presented in the search.

The reason for is it all layed out in the second half of the first comment. The keywords are invisible to the user (the visitor to the sight using the search feature). Because the underlying search tool (lunr.js) is not able to properly tokenize code in the search index, the document author can define search keywords which will ensure that the relevant results will be returned for searches.

>  we are considering moving the search plugin out of Material for MkDocs in the mid-term, and providing a unified search plugin for all themes with a nice UI and UX. 

That is great news. I always intended the existing search plugin to be a filler until a better solution came along. If your plugin can be made standalone, then I would be happy to see this one deprecated. That said, we should probably still go forward with some form of the fix here as this addresses a real problem with the existing plugin and it sounds like yours won't be available for some time.

squidfunk:
@waylan thanks for clarifying. I've read your OP and it was not apparent how (if) they are presented to the user.

> Because the underlying search tool (lunr.js) is not able to properly tokenize code in the search index, the document author can define search keywords which will ensure that the relevant results will be returned for searches.

Interestingly, this is the reason why I reworked the entire search plugin in the first place ☺️ Material for MkDocs search can parse HTML and render (inline) code blocks without problems. If you feel that this makes sense to integrate into MkDocs, feel free to steal anything you want from Material for MkDocs. If you need assistance where to find things, let me know.

> That is great news. I always intended the existing search plugin to be a filler until a better solution came along. If your plugin can be made standalone, then I would be happy to see this one deprecated. That said, we should probably still go forward with some form of the fix here as this addresses a real problem with the existing plugin and it sounds like yours won't be available for some time.

Yes, it'll first be part of Material for MkDocs where we will test it, and we'll move it outside when stable. And yes, I currently can't put an ETA on it, but I'm working hard to make that happen.

waylan:
So I just looked at Material's plugin and noticed that you are defining a more sophisticated token [separator](https://squidfunk.github.io/mkdocs-material/plugins/search/#config.separator). I had dismissed that in favor of manually defining keywords. Maybe I was too hasty in that decision. I think maybe I'll try that and see if I can eliminate the need for keywords.

squidfunk:
@waylan correct, but please note that this also requires a [completely different tokenizer implementation](https://github.com/squidfunk/mkdocs-material/blob/cd086110bfd65eb6dfeed8c3a01f500e9cdb7468/src/templates/assets/javascripts/integrations/search/internal/tokenize/index.ts#L50-L136) in JavaScript. I've written this tokenizer from scratch, it is compatible with Lunr.js, and it is actually able to handle multi-character separators (which the default tokenizer does not). The new search has an even more sophisticated tokenizer, which will be open sourced in the coming months.

waylan:
Wow, I had no idea you were writing your own search engine. I'm happy to hear a lunr.js replacement is coming. As your new engine will be available for all themes in the future, I don't think it makes sense to get too crazy with the changes here. 

The keywords feature has already been removed and in my last commit I am cleaning up whitespace in the index. This is still an improvement of the search results, and in my testing, it seems that I can use a more sophisticated separator regex to get better results on my pages. I don't think it makes sense to default to a more complex regex as it is really only useful for API documentation and can be added to the config by users who need it.

I think this is ready to go.

PR #3562, opened 2024-01-31
Enable autoescape

### LINKED ISSUE #3357: Header and navigation sidebars display incorrectly arbitrary markup

oprypin:
Per https://github.com/mkdocs/mkdocs/issues/3357#issuecomment-1918086639

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -9,6 +9,7 @@ from urllib.parse import urljoin, urlsplit
 
 import jinja2
 from jinja2.exceptions import TemplateNotFound
+from markupsafe import Markup
 
 import mkdocs
 from mkdocs import utils
@@ -212,11 +213,15 @@ def _build_page(
         # Run `page_context` plugin events.
         context = config.plugins.on_page_context(context, page=page, config=config, nav=nav)
 
+        page.content = Markup(page.content or '')
         if excluded:
             page.content = (
-                '<div class="mkdocs-draft-marker" title="This page will not be included into the built site.">'
-                'DRAFT'
-                '</div>' + (page.content or '')
+                Markup(
+                    '<div class="mkdocs-draft-marker" title="This page will not be included into the built site.">'
+                    'DRAFT'
+                    '</div>'
+                )
+                + page.content
             )
 
         # Render the template.
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -32,6 +32,7 @@ from urllib.parse import urlsplit, urlunsplit
 import markdown
 import pathspec
 import pathspec.gitignore
+from markupsafe import Markup
 
 from mkdocs import plugins, theme, utils
 from mkdocs.config.base import (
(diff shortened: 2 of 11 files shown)


pawamoy:
I have tested this branch locally. The only thing that seems to break is the `typeset` plugin from @squidfunk's Material for MkDocs:

typeset
typeset2

If I disable the plugin, the `code` tags are removed, leaving the title unstyled in the ToC. The `span` tag is also removed, and in that particular case, following Javascript code cannot insert text into it, therefore the ToC title ends up as `Funding`.

typeset3
typeset4

Code tags in navigation through gen-files + literate-nav continue to work, as well as code tags in table of contents through mkdocstrings' `heading` Jinja filter making use of `data-toc-label`.

oprypin:
@squidfunk the change matches what is **now** being discussed in #3357 :)

Yes I'll consider naming the release 2.0

Regarding plugins in general, I am checking my "regressions" workflow and I'll be further expanding it, to warn authors in advance or avoid changes in the first place.

Regarding the typeset plugin, I have no access to its source code, but assuming it sets `nav_item.title = 'foo'`, it should simply set `nav_item.title = markupsafe.Markup('foo')` (which can be applied already, as it's forwards-compatible or whatever it's called). Or maybe with further edits that I'll soon make, even that might not be necessary.

oprypin:
So, the idea of autoescaping can work and it's nice, but reckless to try to push through considering the long historic context. Leaving the PR in a working state for posterity.

mondeja:
I see that this should be a must. Whatever page trying to add Mkdocs in a untrusted input context is totally vulnerable to XSS. Are you sure that this shouldn't be enabled by default in a breaking change release?

squidfunk:
@mondeja I kindly disagree. MkDocs pages are not subject to XSS attacks because all content is built statically. There is no threat vector that carries untrusted user input.

mondeja:
> There is no threat vector that carries untrusted user input. Adding auto-escape for the sake of mitigating XSS is fixing something that is not broken. Happy to be proven wrong.

Your theme has to built a feature that includes untrusted input that has to be created with third party integrations, probably this limitation has influenced that decision. Additionally, untrusted input could come from a server too or in the form of third party plugins building the static content from third party sources. What is Mkdocs really depends on the usage, the "is a static generator" is not true and some of the features of your theme prove that.

I'm seeing that your project has eaten up the main one. In this toxic state of the ecosystem I don't think it will contribute anything again.

squidfunk:
> Your theme builds a feature which includes untrusted input that has been created with third party integrations, probably these kinds of limitations influenced that decision. 

I think the distinction between user (viewer) and author (writer) is important here. Only authors are able to change the content or add plugins that do. The only thing the typeset plugin effectively does is to allow to render parts of the page's content in the navigation and table of contents, nothing more. Are you maybe able to provide a demo that shows how XSS can be pulled off in a way that the author is unaware, that does only happen when the typeset plugin is active?

squidfunk:
> I'm seeing that your project has eaten up the main one. In this toxic state of the ecosystem I don't think it will contribute anything again.

I'm sorry, what? Yes, Material for MkDocs is one of the biggest dependents on MkDocs, and I'm not sure I should feel sorry that people actually love to use, but I'm not seeing how this is toxic. Could you please elaborate?

PR #3564, opened 2024-02-04
Strip all HTML when getting the title from the first H1 tag

### LINKED ISSUE #3357: Header and navigation sidebars display incorrectly arbitrary markup

oprypin:
Not stripping it was a bug, and also inconsistent with how ToC titles are extracted.

* Fixes #3357
* Closes #3546

per latest status https://github.com/mkdocs/mkdocs/issues/3357#issuecomment-1925438350


pawamoy:
Tested locally, no issue detected (mkdocstrings and Material for MkDocs).

ofek:
Nice!

waylan:
I have been working on Python-Markdown/markdown#1441 in my spare time over the last few days and have not been following any updates on GitHub. After creating my PR, I saw this for the first time. It appears that my changes there would directly affect this here. Sorry if I've thrown a wrench in the works.

For example, I removed the `markdown.extensions.toc.stashedHTML2text` function, which would break the changes here. Instead, I am running all postprocessors (which was removed in this change). I felt like this was necessary to resolve some third-party markup (like the emoji extension). I still need to flesh out some new tests for my changes, but I have verified that the existing tests (for the Python-Markdown lib) are all passing.

Of note here is that I added a new attribute to the `toc_tokens` which contains the rich text `HTML` content of each heading along with the associated TOC data. My thinking was that MkDocs could use this new attribute without needing to parse/step through the entire document. As now Markdown already ensures it is fully rendered, MkDocs doesn't need to be dipping into Markdown's internals to do that itself.

The changes in the linked PR are in a draft, so subject to change at this point. Therefore, any feedback is welcome to ensure we don't break things and/or we provide a useable solution. Of course, before MkDocs can make use of our changes we will need to finalize them and make a release.

oprypin:
OK I can revert the usage of markdown.extensions.toc.stashedHTML2text  and we can find a common solution

PR #3570, opened 2024-02-12
Don't physically copy static files for `mkdocs serve`

### LINKED ISSUE #2662: no copy serve for large sites(10Gb)
Dear mkdocs team,

Thanks for your great work. May I know is there a way to`mkdocs serve` without copy files? My site is about 10Gb with 2000+files(pdfs), most of those are just for downloads and reading in chrome. Every time serve, I notice it will copy those files with endless time cost. I guess if we don't copy, it would be much faster. This servers only for local view, so there are no security issues without using copy.

Thanks

oprypin:
Just read them from their original location in the live server instead

* Closes #2662


squidfunk:
Have you tested compatibility with plugins that augment the serve capabilities of MkDocs? For example, the mkdocs monorepo plugin provides a [consolidated serve mode](https://github.com/backstage/mkdocs-monorepo-plugin/blob/b1965a216af41c7e503753c63551d894fac8399d/mkdocs_monorepo_plugin/plugin.py#L74-L100), same as the [projects plugin](https://squidfunk.github.io/mkdocs-material/plugins/projects/).

PR #3613, opened 2024-03-28
Add Polish translation for built-in themes

s-rakowski:
Hello, I added `.po` files with Polish language strings for the two built-in themes.
Unfortunately, lunr-languages doesn't support Polish but I think this is out of scope for this pull request.

I also noticed that in the "ReadTheDocs" theme, in the footer the word "theme" is not added as a translatable string:

> Built with [MkDocs](https://www.mkdocs.org/) using a [theme](https://github.com/readthedocs/sphinx_rtd_theme) provided by [Read the Docs](https://readthedocs.org/).

Therefore it gets translated to:

> Zbudowano przy użyciu [MkDocs](https://www.mkdocs.org/) z pomocą [theme](https://github.com/readthedocs/sphinx_rtd_theme) dostarczonego przez [Read the Docs](https://readthedocs.org/).

 Would it be OK to modify `footer.html` within this PR so that the word "theme" gets added to `messages.pot`?

PR #3625, opened 2024-04-09
Re-generate localization files, update a new word

oprypin:
- Re-generate localization files
- Translate the word "theme" added to the readthedocs theme  in https://github.com/mkdocs/mkdocs/pull/3613


tomchristie:
Hmmm... not clear to me why we've got a 3.12 build failure here, any thoughts?

oprypin:
It's a build failure that I caused on master branch and then fixed 
https://github.com/mkdocs/mkdocs/commits/1451fb3f2cbc70420982e5b1f3ea7c25ec409f01
So there is no failure from this pull request, it's just an outdated result

PR #3649, opened 2024-04-17
Remove jQuery from mkdocs theme

### LINKED ISSUE #3648: Remove jQuery dependency from mkdocs theme
Just as a reminder.

This is now easy to do, there are just some basic usages in base.js

* Ref https://github.com/mkdocs/mkdocs/pull/3493

> as of Bootstrap version 5, jQuery was dropped as a dependency

squidfunk:
Since the `mkdocs` theme updated to Bootstrap 5, jQuery is not necessary anymore. I've replaced all uses of jQuery with common browser APIs and adapted the use of the modals to Bootstrap 5. My testing shows that `master` behaves exactly as if this PR were merged. If not, please report errors here and I'll fix them.

I don't think jQuery can be removed from the `readthedocs` theme.

Closes #3648


squidfunk:
Side note: merging this PR will reduce the payload that the browser needs to download by 90kb (30kb gzipp'd), which is an improvement of -14% (currently 213kb gzipp'd). Thus, the MkDocs core theme should load faster!

waylan (approved):
Looks good to me. It is great to see this being worked on. I looked through the changes and they look good to me, but I'm weakest with JavaScript so I may have missed something.

squidfunk:
Do we want to release this with MkDocs 1.6.0, or afterwards? I've tested all that was apparent to me from reading the code and have found no problems, but I'm also not a heavy user of the `mkdocs` theme.

waylan:
I would include this in 1.6.0 if we can.

squidfunk:
Great! Since everybody seems to agree that we want this in 1.6.0, let's merge it.

PR #3659, opened 2024-04-20
Deployed 0998fec7e with MkDocs version: 1.6.0

oprypin:


oprypin:
I created this pull request because `mkdocs gh-deploy` was getting stuck for me. But it worked now, so never mind.

Issue #3680, opened 2024-04-24
New default theme...


tomchristie:
<p align="right">Escalated from discussion #3661</p>

<img width="988" alt="image" src="https://github.com/mkdocs/mkdocs/assets/647359/f694f035-62a5-4677-95c4-c45958e82442">

<p align="center">・・・</p>

I've spent a bit of time looking at a re-working from scratch. but perhaps we should start with some simple iterative improvements to the default theme?...

* The current coloring is just too loud. Here's a more muted `darkslateblue`, without the unnecessary gradiant gloss.
* Drop the `previous` and `next` controls from the nav bar.
* Drop the text from the search/repo/color mode.

---

<p align="center"><img width="537" alt="image" src="https://github.com/mkdocs/mkdocs/assets/647359/f1cf1a37-907e-43bc-bde6-9fb13e4fa939"></p>

<p align="center">・・・</p>

* Less verbose search modal.

---

<p align="center"><img width="293" alt="image" src="https://github.com/mkdocs/mkdocs/assets/647359/2b1bdbd8-3e29-481a-8070-a312e9b1f241"></p>

<p align="center">・・・</p>

* Bring transparency to the table of contents.

PR #3681, opened 2024-04-25
Simplify navbar

tomchristie:
<p align="right">Refs #3680</p>

image

* Drop `next`/`prev` controls.
* Drop text from `search` & `repo` controls, except in expanded menu.

<details>
<summary><i>expanded menu</i></summary>

image
</detail>

### DIFF
--- a/mkdocs/themes/mkdocs/base.html
+++ b/mkdocs/themes/mkdocs/base.html
@@ -114,24 +114,9 @@
                         {%- if 'search' in config.plugins %}
                         <li class="nav-item">
                             <a href="#" class="nav-link" data-bs-toggle="modal" data-bs-target="#mkdocs_search_modal">
-                                <i class="fa fa-search"></i> {% trans %}Search{% endtrans %}
+                                <i class="fa fa-search"></i><span class="d-lg-none ms-2">Search</span>
                             </a>
                         </li>
-                        {%- endif %}
-                      {%- endblock %}
-
-                      {%- block next_prev %}
-                        {%- if page and (page.next_page or page.previous_page) %}
-                            <li class="nav-item">
-                                <a rel="prev" {% if page.previous_page %}href="{{ page.previous_page.url|url }}" class="nav-link"{% else %}class="nav-link disabled"{% endif %}>
-                                    <i class="fa fa-arrow-left"></i> {% trans %}Previous{% endtrans %}
-                                </a>
-                            </li>
-                            <li class="nav-item">
-                                <a rel="next" {% if page.next_page %}href="{{ page.next_page.url|url }}" class="nav-link"{% else %}class="nav-link disabled"{% endif %}>
-                                    {% trans %}Next{% endtrans %} <i class="fa fa-arrow-right"></i>
-                                </a>
-                            </li>
                         {%- endif %}
                       {%- endblock %}
 
@@ -140,11 +125,11 @@
                             <li class="nav-item">
                                 <a href="{{ page.edit_url }}" class="nav-link">
                                     {%- if config.repo_name == 'GitHub' -%}
-                                        <i class="fa-brands fa-github"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-github"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name == 'Bitbucket' -%}
-                                        <i class="fa-brands fa-bitbucket"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-bitbucket"></i><span class="d-lg-none ms-2"></span>{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name == 'GitLab' -%}
-                                        <i class="fa-brands fa-gitlab"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-gitlab"></i><span class="d-lg-none ms-2"></span>{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name -%}
                                         {% trans repo_name=config.repo_name%}Edit on {{ repo_name }}{% endtrans %}
                                     {% else %}
@@ -171,8 +156,7 @@
                       {%- if config.theme.user_color_mode_toggle %}
                             <li class="nav-item dropdown">
                               <button id="theme-menu" aria-expanded="false" data-bs-toggle="dropdown" data-bs-display="static" aria-label="Toggle theme" class="nav-link dropdown-toggle">
-                                <i class="fa-solid fa-circle-half-stroke fa-fw"></i>
-                                <span class="d-lg-none ms-2">Toggle theme</span>
+                                <i class="fa-solid fa-circle-half-stroke fa-fw"></i><span class="d-lg-none ms-2">Toggle theme</span>
                               </button>
                               <ul class="dropdown-menu dropdown-menu-end">
                                 <li>


mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
-                                </a>
-                            </li>
-                            <li class="nav-item">
-                                <a rel="next" {% if page.next_page %}href="{{ page.next_page.url|url }}" class="nav-link"{% else %}class="nav-link disabled"{% endif %}>
-                                    {% trans %}Next{% endtrans %} <i class="fa fa-arrow-right"></i>
-                                </a>
-                            </li>
-                        {%- endif %}
-                      {%- endblock %}
+                      {%- block next_prev %}{}{%- endblock %}
```
pawamoy, line 123: Shouldn't the block be empty? (i.e. remove `{}`)

tomchristie:
Good pointer, thanks.

I've re-included it as an empty block here and amended the docs marginally. ☺️

mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
                       {%- block repo %}
                         {%- if page and page.edit_url %}
                             <li class="nav-item">
                                 <a href="{{ page.edit_url }}" class="nav-link">
                                     {%- if config.repo_name == 'GitHub' -%}
-                                        <i class="fa-brands fa-github"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-github"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name == 'Bitbucket' -%}
-                                        <i class="fa-brands fa-bitbucket"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-bitbucket"></i><span class="d-lg-none ms-2"></span>{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
```
pawamoy, line 132: Extra closing span tag, same on the next line.

pawamoy (changes requested):

mkdocs/themes/mkdocs/base.html
tomchristie, line 132:
```suggestion
                                        <i class="fa-brands fa-bitbucket"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
                                     {%- elif config.repo_name == 'GitLab' -%}
-                                        <i class="fa-brands fa-gitlab"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-gitlab"></i><span class="d-lg-none ms-2"></span>{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
```
tomchristie, line 134:
```suggestion
                                        <i class="fa-brands fa-gitlab"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
```

pawamoy (approved):

HumbleDeer:
The removal of the "Edit on GitHub" text might obscure the purpose of the GitHub icon in the navbar. 

If this simplification is still desirable, I suggest amending this by adding an "Edit on GitHub" link at the bottom of a page instead, as is custom with a fair few other static site generator packages and other documentation generators alike.

tomchristie:
Good catch. Tho we should switch this up completely and just have a `repo_url` control here, rather than `page.edit_url`.

If we want to keep an "Edit this page" in the default theme, then that control should be in the page area not the nav bar area.

mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
-                                </a>
-                            </li>
-                        {%- endif %}
-                      {%- endblock %}
+                      {%- block next_prev %}{%- endblock %}
 
                       {%- block repo %}
                         {%- if page and page.edit_url %}
                             <li class="nav-item">
                                 <a href="{{ page.edit_url }}" class="nav-link">
```
tomchristie, line 128:
```suggestion
                                <a href="{{ repo_url }}" class="nav-link">
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
                                     {%- elif config.repo_name == 'Bitbucket' -%}
-                                        <i class="fa-brands fa-bitbucket"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-bitbucket"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name == 'GitLab' -%}
-                                        <i class="fa-brands fa-gitlab"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-gitlab"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name -%}
                                         {% trans repo_name=config.repo_name%}Edit on {{ repo_name }}{% endtrans %}
```
tomchristie, line 136:
```suggestion
                                        {{ repo_name }}
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
-                                        <i class="fa-brands fa-github"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-github"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name == 'Bitbucket' -%}
-                                        <i class="fa-brands fa-bitbucket"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-bitbucket"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name == 'GitLab' -%}
-                                        <i class="fa-brands fa-gitlab"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-gitlab"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
```
tomchristie, line 134:
```suggestion
                                        <i class="fa-brands fa-gitlab"></i><span class="d-lg-none ms-2">{{ config.repo_name }}</span>
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
                       {%- block repo %}
                         {%- if page and page.edit_url %}
                             <li class="nav-item">
                                 <a href="{{ page.edit_url }}" class="nav-link">
                                     {%- if config.repo_name == 'GitHub' -%}
-                                        <i class="fa-brands fa-github"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-github"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
                                     {%- elif config.repo_name == 'Bitbucket' -%}
-                                        <i class="fa-brands fa-bitbucket"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-bitbucket"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
```
tomchristie, line 132:
```suggestion
                                        <i class="fa-brands fa-bitbucket"></i><span class="d-lg-none ms-2">{{ config.repo_name }}</span>
```

mkdocs/themes/mkdocs/base.html
```diff
@@ -114,37 +114,24 @@
...
-                      {%- endblock %}
+                      {%- block next_prev %}{%- endblock %}
 
                       {%- block repo %}
                         {%- if page and page.edit_url %}
                             <li class="nav-item">
                                 <a href="{{ page.edit_url }}" class="nav-link">
                                     {%- if config.repo_name == 'GitHub' -%}
-                                        <i class="fa-brands fa-github"></i> {% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}
+                                        <i class="fa-brands fa-github"></i><span class="d-lg-none ms-2">{% trans repo_name=config.repo_name %}Edit on {{ repo_name }}{% endtrans %}</span>
```
tomchristie, line 130:
```suggestion
                                        <i class="fa-brands fa-github"></i><span class="d-lg-none ms-2">{{ config.repo_name }}</span>
```

mkdocs/themes/mkdocs/base.html
tomchristie, line 136:
```suggestion
                                    {%- elif config.repo_name -%}
                                        {{ config.repo_name }}
```

HumbleDeer:
> Good catch. Tho we should switch this up completely and just have a `repo_url` control here, rather than `page.edit_url`.
> 
> If we want to keep an "Edit this page" in the default theme, then that control should be in the page area not the nav bar area.

Other docs packages often have edit buttons in the page area (indeed), and have it optional or to be enabled manually. I think that's a solid option, and agree it should be a page area element.

In that sense (and I don't mean this snarkily), a navigation bar is for navigation. Anything not navigation or locating of pages like a navigation is possibly better off in the footer or page content.

dagnelies:
- 👍  Drop next/prev controls.
- 👍  Move "edit on Github" from the navbar to the page
- 😢  Drop text from search controls

I think these are good choice, as it would make the navbar lighter ...except for the "search". Searching is so crucial that it deserves to be clearly visible IMHO.

HumbleDeer:
> * 😢  Drop text from search controls
> 
> I think these are good choice, as it would make the navbar lighter ...except for the "search". Searching is so crucial that it deserves to be clearly visible IMHO, including the "Search" label, not only the small icon.

You're absolutely right to remark this.

The "Search" text besides and in line with the magnifying glass icon is paramount to the function and form of a search feature and the search box. 

Why?

- The "Search..." text in a small dedicated space with the magnifying glass signifies this is a text field you can enter your own text into
- The "Search..." with the magnifying glass functions as a "call to action", which means it by itself intuitively shows a user that it can be interacted with.
- The magnifying glass is commonly and typically used as the CTA for submitting your entered search query and executing the actual search action; It is as such not usually used to represent the opening of a search string entering field. 
  It is a good idea to keep it as a CTA only, and retain its classic "submit" CTA.

mkdocs/themes/mkdocs/base.html
janbrasna: Actually I think that both the `edit_url` and `next_prev` can go the page/content footer. That's how other generators mostly place them by default (and some inhouse boutique docs stacks too) and consumers are likely to be looking for them down there (or, basically acting as a CTA after finishing reading the current page — prev, next, or edit if needs updating).

Good call to have the repo icon in header only link to the `repo_url` to avoid any confusion (edit links render as 404 for logged-out users etc.), again pretty much a pattern nowadays.

janbrasna: PS: watch out for the `{%- if page and page.edit_url %}` condition right above if switching away from page.edit_url here…

PR #3682, opened 2024-04-25
Resolve CI

tomchristie:
Resolve CI test case.


pawamoy (approved):
Could use a tuple instead of a list but otherwise looking good :+1:

tomchristie:
✅ ⏱️ 👀 😅

*that's a bit of [a turnaround time](https://github.com/mkdocs/mkdocs/actions/runs/8836092856/job/24261851413) we've got ourselves here.*

PR #3784, opened 2024-07-26
Fix strict mode in Docker by removing 0.0.0.0 warning

### LINKED ISSUE #3724: Please allow running MkDocs with Docker and strict mode
Currently it seems impossible to run MkDocs from Docker.

We don't want to impose the burden of configuring a Python environment to contributors to our documentation (our community consists mainly of JavaScript and/or C++ developers), so we use use a Docker container that packages MkDocs (for Material).

Unfortunately, after enabling strict mode we are now greeted with this error

```
docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material

WARNING -  Config value 'dev_addr': The use of the IP address '0.0.0.0' suggests a production environment or the use of a proxy to connect to the MkDocs server. However, the MkDocs' server is intended for local development purposes only.
           Please use a third party production-ready server instead.

Aborted with 1 configuration warnings in 'strict' mode!
```

And the container fails to start. See also https://github.com/squidfunk/mkdocs-material/issues/7197 and https://github.com/mkdocs/mkdocs/issues/2108

When I use

```
docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material serve
```

The server running inside the Docker container cannot be accessed from outside Docker.

squidfunk:
Fixes #3724 and #2108. Consensus for this change was reached in https://github.com/mkdocs/mkdocs/issues/3724#issuecomment-2178996420

> I'd suggest we simply don't enforce the check against 0.0.0.0. (https://github.com/mkdocs/mkdocs/pull/2022)
Less config is better. Less code is better.

### DIFF
--- a/mkdocs/config/config_options.py
+++ b/mkdocs/config/config_options.py
@@ -488,16 +488,6 @@ class IpAddress(OptionallyRequired[_IpAddressValue]):
 
         return _IpAddressValue(host, port)
 
-    def post_validation(self, config: Config, key_name: str):
-        host = config[key_name].host
-        if key_name == 'dev_addr' and host in ['0.0.0.0', '::']:
-            self.warnings.append(
-                f"The use of the IP address '{host}' suggests a production environment "
-                "or the use of a proxy to connect to the MkDocs server. However, "
-                "the MkDocs' server is intended for local development purposes only. "
-                "Please use a third party production-ready server instead."
-            )
-
 
 class URL(OptionallyRequired[str]):
     """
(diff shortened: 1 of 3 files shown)


tomchristie (approved):
Trusting the user here seems sensible, yep.

pawamoy (approved):
Looks good :) No mention of this in the docs?

squidfunk:
I searched for `0.0.0.0` but didn't find anything. I think since it's only a warning, the warning itself was not documented, and the general advise to not use the dev server for production would be, if mentioned in the docs, unaffected as it still holds.

PR #3786, opened 2024-07-29
add on_page_template event

carlocastoldi:
By adding this new event, i'd like to allow plugins to change template based on the page being built.

An example of usage can be seen in this in-development plugin: [mkdocs_multi_theme_plugin](https://github.com/carlocastoldi/mkdocs-multi-theme-plugin/blob/master/mkdocs_multi_theme_plugin/plugin.py)

Before resorting to this upstream change, i have considered using other events (e.g. `on_env`), but I am afraid none allow per-page editing of the templates from the ground up.

### DIFF
--- a/mkdocs/commands/build.py
+++ b/mkdocs/commands/build.py
@@ -207,10 +207,14 @@ def _build_page(
         context = get_context(nav, doc_files, config, page)
 
         # Allow 'template:' override in md source files.
-        template = env.get_template(page.meta.get('template', 'main.html'))
+        template_file = page.meta.get('template', 'main.html')
+        template = env.get_template(template_file)
 
         # Run `page_context` plugin events.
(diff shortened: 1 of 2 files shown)


carlocastoldi:
no, there is no associated issue to my knowledge.

Due to my requirements, i needed to have some pages following a theme, while others using another theme. I tried to only build [a plugin](https://github.com/carlocastoldi/mkdocs-multi-theme-plugin), however I soon understood that the best solution required me to modify mkdocs itself.
While, according to my needs, I think I could live just with my small fork of mkdocs, I think this new hook could also be useful to others. Hence why i opened this pull request!

carlocastoldi:
fyi, there was a brief discussion associated to something related to this topic: https://github.com/mkdocs/mkdocs/discussions/3645

PR #3807, opened 2024-08-17
Update getting-started markdown curl

janbrasna:
The generated lorem-markdownum has random links to random domains, often working, and often NSFW/unsafe — so this constrains the example links to only local anchors for safety.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -81,7 +81,7 @@ option is the only required option in your configuration file.
 Now add a second page to your documentation:
 
 ```bash
-curl 'https://jaspervdj.be/lorem-markdownum/markdown.txt' > docs/about.md
+curl 'https://jaspervdj.be/lorem-markdownum/markdown.txt?no-external-links=on' > docs/about.md
 ```
 
 As our documentation site will include some navigation headers, you may want to


pawamoy (approved):

docs/getting-started.md
```diff
@@ -81,7 +81,7 @@ option is the only required option in your configuration file.
 Now add a second page to your documentation:
 
 ```bash
-curl 'https://jaspervdj.be/lorem-markdownum/markdown.txt' > docs/about.md
+curl 'https://jaspervdj.be/lorem-markdownum/markdown.txt?no-external-links=on' > docs/about.md
```
janbrasna, line 84: I have created https://tinyurl.com/lorem-markdownum-txt and https://dub.sh/lorem-markdownum shortlinks including the `no-external-links` param to avoid the long API URL with query param, but that'd need a -L flag to follow redirects and also kinda obscures what's the target URL so I ended up with just this original endpoint.

Nevertheless I've opened https://github.com/jaspervdj/lorem-markdownum/issues/14 to not generate random links, so if that one ever gets resolved I'll get back to this and change it back to default once safe;)

tomchristie:
We had a neat link to an alternative Lorum Ipsum generator in a different thread, that I thought at the time looked like an improvement... https://brettterpstra.com/md-lipsum/

Would this be an improvement?

janbrasna:
Yup that was me — I just wasn't able to generate several versions and check we're rendering them correctly (wrt features; definitions, footnotes etc.) before proposing that.

Let me quickly check a few outputs (again, their defaults, as we probably don't want a three-line-long query URL just fine-tuned to what we need;D) and get back to you.

janbrasna:
@tomchristie So that one has markdown features making the output look somewhat subpar:

1. it generates OL and UL in succession that doesn't get interpreted well:

> Screen Shot 2024-08-20 at 20 25 08

2. Includes definition lists:

> Screen Shot 2024-08-20 at 20 27 36

3: I don't even know what the hell `==` is supposed to do ;D

> Screen Shot 2024-08-20 at 20 28 19

tomchristie:
Okay. I'd suggest we take a pass on this PR, let things be. `?no-external-links=on` is graceless and I dislike it.

docs/getting-started.md
janbrasna, line 84: @tomchristie If this looks better:

```suggestion
curl -L 'https://tinyurl.com/lorem-markdownum' > docs/about.md
```

? (I kinda dislike relying on some redirector changing its behavior in the future breaking this etc.)

tomchristie:
> The generated lorem-markdownum has random links to `[the internet]`.

PR #3851, opened 2024-10-19
add `headers` config options for `mkdocs serve`

stefanhoelzl:
this changes allows to specifiy custom HTTP headers to be sent when using `mkdocs serve`

This new options is needed, when you need extra HTTP headers like [CORS](https://de.wikipedia.org/wiki/Cross-Origin_Resource_Sharing)

With this change these headers can be specified in `mkdocs.yaml` like this:
```yaml
headers:
  Access-Control-Allow-Origin: "*"
```

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -67,7 +67,12 @@ def serve(
         build(config, serve_url=None if is_clean else serve_url, dirty=is_dirty)
 
     server = LiveReloadServer(
-        builder=builder, host=host, port=port, root=site_dir, mount_path=mount_path
+        builder=builder,
+        host=host,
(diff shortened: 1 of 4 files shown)


pawamoy:
@stefanhoelzl thanks for the PR, but it is better to open an issue first, to discuss your use-case and whether it's already supported or not, and whether we want to support it, and how.

stefanhoelzl:
> @stefanhoelzl thanks for the PR, but it is better to open an issue first, to discuss your use-case and whether it's already supported or not, and whether we want to support it, and how.

Thanks, created a [issue](https://github.com/mkdocs/mkdocs/issues/3852) where it can be discussed

PR #3859, opened 2024-11-04
fix and test description of draft_docs

### LINKED ISSUE #3858: draft_docs does not support comments, which are shown in the docs
The [documentation for `draft_docs`](https://www.mkdocs.org/user-guide/configuration/#draft_docs) shows inline comments, like this:

```yaml
draft_docs: |
  drafts/               # A "drafts" directory anywhere.
  _unpublished.md       # A md file ending in _unpublished.md
  !/foo_unpublished.md  # But keep this particular file.
```

However, it seems that adding comments like that causes the pattern matching to break (presumably the comments get included in the pattern).

I copied this section verbatim as a starting point, but it didn't cause the matching pages to be excluded from the build, and it took a surprising amount of trial and error to work out what the problem was. As soon as I removed the inline comments, everything started working as expected. Sorry if it's just me being a bit dim, but I doubt I'll be the only one.

I'm not sure if it this is a problem with the docs (i.e. not expected to work, and the explanations would be better as prose outside the example) or with the code (it's a bug and the comments should get automatically stripped).

The same issue probably applies to `exclude_docs`.

pxc:
Update the User Guide descriptions that use inline comments in gitignore syntax (which doesn't work), and add a test that the mult-iline `draft_docs` definition replacement does work correctly.

Fixes  #3858

### DIFF
--- a/mkdocs/tests/__init__.py
+++ b/mkdocs/tests/__init__.py
@@ -1,7 +1,7 @@
 import logging
 import unittest.util
 
-unittest.util._MAX_LENGTH = 100000
+unittest.util._MAX_LENGTH = 100000  # type: ignore
 
 
 class DisallowLogsHandler(logging.Handler):
--- a/mkdocs/tests/__init__.py
+++ b/mkdocs/tests/__init__.py
@@ -1,7 +1,7 @@
 import logging
 import unittest.util
 
-unittest.util._MAX_LENGTH = 100000  # type: ignore
+unittest.util._MAX_LENGTH = 100000  # type: ignore[misc]
 
 
 class DisallowLogsHandler(logging.Handler):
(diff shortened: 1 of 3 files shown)


mkdocs/tests/__init__.py
```diff
@@ -1,7 +1,7 @@
 import logging
 import unittest.util
 
-unittest.util._MAX_LENGTH = 100000
+unittest.util._MAX_LENGTH = 100000  # type: ignore
```
pxc, line 4: I was getting this mypy error here:

```
mkdocs/tests/__init__.py:4: error: Cannot assign to final name "_MAX_LENGTH"  [misc]
```

pawamoy: :+1: 

```suggestion
unittest.util._MAX_LENGTH = 100000  # type: ignore[misc]
```

pawamoy (approved):
Amazing, thanks a lot! Just a tiny suggestion :slightly_smiling_face:

rachel-rigdon-sp:
Was this change deployed? I see that the public docs still have the incorrect doc example.

Issue #3882, opened 2024-12-09
Document the is_index property of page objects


IreneKnapp:
Hi all! This is a feature request. Briefly: The page objects exposed through the template language have a property is_index, returning a boolean based on the page's input filename. I suspect this being visible to template code is an accident, but I think it's useful and would like to argue for documenting it and treating it as a stable interface.

So, thanks for hearing me out. I'm writing a theme, and I found a scenario where I wanted to determine whether a given page was an index page or not. Specifically, I wanted to treat it differently in a breadcrumb bar, collapsing redundant information when the name of an index page would otherwise appear right after the name of the section it's part of.

I'm fully aware that this is somewhat quirky behavior which would not be appropriate for every document. In fact, it's definitely inappropriate for documents that are encyclopedia-like, where sections don't necessarily have introductions, and the first page in a section might be just one of many similar entries. However, it is a thing some authors, including myself, might reasonably want for documents that have some sort of narrative structure whereby there's front matter at the beginning of a section that contextualizes the later pages in the section. Also, a single document might have both kinds of sections, so a project-level setting isn't an appropriate solution.

I feel like this is super abstract, so I'm going to offer an example.

* Core concepts
   * Introduction
   * Thing One
   * Thing Two
* Alphabetical list of dragons
  * Blue dragon
  * Green dragon
  * Red dragon

... If we imagine that "core concepts > introduction" is an index page, it makes sense to want to treat it specially navigation-wise. However, "blue dragon" should not receive the same treatment; it's not an index page, just a page that is coincidentally the first in its section.

I hope that makes sense.

My impression is that MkDocs is trying to be a general-purpose tool, so I think a feature like this is appropriate. That way it can be up to theme and document authors to decide when to use it, and perhaps most won't, but I know I'm planning to. If this is out of scope for MkDocs, that's fine and I'm not hurt. The code for is_index is already written, so nominally it seems easy enough to treat it as an official feature... anyway, I thought I'd ask!

lazee:
Hi @IreneKnapp . Why do you think it is visible by accident?. The code

```
@property
def is_index(self) -> bool:
   return self.file.name == 'index'
```

has been in the codebase for many years (6 years). So my guess is that there are no plans to remove it. But you are right about it being undocumented. I suggest filing a tiny PR with a doc string. Or, if you are concerned about using it, you can always check on the filename, like this function does.

IreneKnapp:
[...]

PR #3886, opened 2024-12-24
Document the `is_index` property of page objects

IreneKnapp:
I was told on https://github.com/mkdocs/mkdocs/issues/3882 that this is likely to be a stable interface and that there's no reason it shouldn't be documented as such, so I'm submitting this PR to do that.

Thanks in advance!


IreneKnapp:
apologies for the force-pushes, I inadvertently included my NixOS packaging of the various mkdocs plugins as part of the original commit

mkdocs/structure/pages.py
```diff
@@ -124,6 +124,7 @@ def active(self, value: bool):
 
     @property
     def is_index(self) -> bool:
+        """When `True`, indicates that this page is an index page, in the sense that it gets sorted to the beginning of the list of its siblings because its filename was `index.md` or something similar."""
```
pawamoy, line 127: Feels a bit long for a summary.

```suggestion
        """When `True`, indicates that this page is an index page.
        
        Index pages get sorted to the beginning of the list of their siblings
        because their filename is `index.md` or `README.md`.
        """
```

Actually not entirely sure about `README.md`. I think @tomchristie would be able to confirm (and maybe rephrase to something better).

pawamoy (changes requested):
No worries, and thanks! Just one comment.

mkdocs/structure/pages.py
pawamoy: Maybe just drop the second line in the body.

reteps:
Unrelated, but `is_index` currently could have improved handling. See https://github.com/squidfunk/mkdocs-material/issues/7561

Issue #3905, opened 2025-01-17
"mkdocs serve" does not cleanup after a SIGTERM


toddke:
It looks like mkdocs does not handle SIGTERM and leaves its temporary site_dir in the /tmp directory. These directories are never cleaned up. mkdocs should handle SIGTERM and/or cleanup leftover /tmp directories on startup.

#### My setup
I run mkdocs inside a docker container [see the [mkdocs-material project](https://github.com/squidfunk/mkdocs-material)]. By default, the docker starts "mkdocs serve".

The issue occurs when the container is stopped. To stop a container, docker [sends SIGTERM](https://docs.docker.com/reference/cli/docker/container/stop/#description) to the main process and waits for the signal to be handled. In the case of mkdocs-material, the main process is tini, but tini already [forwards signals](https://github.com/krallin/tini/blob/master/src/tini.c#L526) to the child mkdocs process. To confirm the behaviour, I've also sent SIGTERM directly to mkdocs [eg. `kill -s TERM <mkdocs-pid>`] and mkdocs still fails to clean up /tmp directories.

pawamoy:
Thanks, I was able to replicate the described issue. I added the bug label.

alvaro-osvaldo-tm:
I will fix it in this weekend.

alvaro-osvaldo-tm:
Done, I made the fix at pull request  at #3913 .

All tests was passed and I attached in the pull request comment a 'end-to-end' test for these specific case.

Any doubt, I'm open to answer.

toddke:
Thank you for the quick turnaround @alvaro-osvaldo-tm

orias:
I'm using mkdocs in VSCode Debian WSL(w11) & my /tmp/ folder is giant.

<img width="677" height="600" alt="Image" src="https://github.com/user-attachments/assets/58e7c364-5747-4981-90fb-d6f78df52765" />

Thanks for reporting this!

alvaro-osvaldo-tm:
Hi @orias , thanks for following up on this issue. 
I have developed a patch for this problem, but it has not been merged yet. 

I will speak with the maintainers to facilitate the merge; hopefully, it will be available soon.

orias:
Thank you!

PR #3913, opened 2025-01-27
fix: Fixed directory cleanup after terminated by signal (#3905)

alvaro-osvaldo-tm:
- Now the directory is cleanup when receive any 'terminate' class signal
- The 'Server' class have an attribute 'is_active' to know the server state

### DIFF
--- a/mkdocs/commands/serve.py
+++ b/mkdocs/commands/serve.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import logging
 import shutil
+import sys
 import tempfile
 from os.path import isdir, isfile, join
+from signal import SIGINT, SIGTERM, signal
 from typing import TYPE_CHECKING
 from urllib.parse import urlsplit
 
@@ -78,7 +80,47 @@ def serve(
                     return f.read()
         return None
 
+    def handle_signal(signum, frame) -> None:
+
+        from signal import strsignal
+
+        signal_name = strsignal(signum)
+        log.info(f"Received signal '{signal_name}'")
+
+        shutdown()
+
+    def shutdown() -> None:
+        if not server.is_active:
+            return
+
+        log.info("Shutting down...")
+
+        try:
+            server.shutdown()
+        finally:
+            config.plugins.on_shutdown()
+
+        if isdir(site_dir):
+            shutil.rmtree(site_dir)
+
     server.error_handler = error_handler
+
+    signal(SIGTERM, handle_signal)
+    signal(SIGINT, handle_signal)
+
+    if sys.platform == "linux":
+        from signal import SIGHUP, SIGQUIT, SIGUSR1, SIGUSR2
+
+        signal(SIGHUP, handle_signal)
+        signal(SIGUSR1, handle_signal)
+        signal(SIGUSR2, handle_signal)
+        signal(SIGQUIT, handle_signal)
+    elif sys.platform == "win32":
+        from signal import CTRL_BREAK_EVENT, CTRL_C_EVENT, SIGBREAK
+
+        signal(SIGBREAK, handle_signal)
+        signal(CTRL_C_EVENT, handle_signal)
+        signal(CTRL_BREAK_EVENT, handle_signal)
 
     try:
         # Perform the initial build
@@ -100,13 +142,7 @@ def serve(
             for item in config.watch:
                 server.watch(item)
 
-        try:
-            server.serve(open_in_browser=open_in_browser)
-        except KeyboardInterrupt:
-            log.info("Shutting down...")
-        finally:
-            server.shutdown()
+        server.serve(open_in_browser=open_in_browser)
+
     finally:
-        config.plugins.on_shutdown()
-        if isdir(site_dir):
-            shutil.rmtree(site_dir)
+        shutdown()
--- a/mkdocs/livereload/__init__.py
+++ b/mkdocs/livereload/__init__.py
@@ -21,6 +21,7 @@ import urllib.parse
 import webbrowser
 import wsgiref.simple_server
 import wsgiref.util
+from os import getpid
 from typing import Any, BinaryIO, Callable, Iterable
 
 import watchdog.events
@@ -111,6 +112,8 @@ class LiveReloadServer(socketserver.ThreadingMixIn, wsgiref.simple_server.WSGISe
                 self.address_family = socket.AF_INET6
         except Exception:
             pass
+
+        self._is_active = False
         self.root = os.path.abspath(root)
         self.mount_path = _normalize_mount_path(mount_path)
         self.url = _serve_url(host, port, mount_path)
@@ -170,6 +173,7 @@ class LiveReloadServer(socketserver.ThreadingMixIn, wsgiref.simple_server.WSGISe
             self.observer.unschedule(self._watch_refs.pop(path))
 
     def serve(self, *, open_in_browser=False):
+        self._is_active = True
         self.server_bind()
         self.server_activate()
 
@@ -178,6 +182,7 @@ class LiveReloadServer(socketserver.ThreadingMixIn, wsgiref.simple_server.WSGISe
 
             paths_str = ", ".join(f"'{_try_relativize_path(path)}'" for path in self._watched_paths)
             log.info(f"Watching paths for changes: {paths_str}")
+
 
         if open_in_browser:
             log.info(f"Serving on {self.url} and opening it in a browser")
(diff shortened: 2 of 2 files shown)


alvaro-osvaldo-tm:
# Problem

When received a signal SIGTERM the directory cleanup is ignored due the execution not reach the 'clean up code'.

# Solution

I wasn't able to move the 'clean up code' to  make it available to both SIGTERM and SIGINT cases without side effects, so I created the function 'handle_signal' that handle both SIGTERM, SIGINT  and other edge case for both Windows and Linux and realize the shutdown propertly.

To avoid race conditions a property was created to 'Server' class to know if was yet stopped.


# Test

A shell script for a end-to-end test is available to validate the behaviour. 
[test-3905.txt](https://github.com/user-attachments/files/18552534/test-3905.txt)

alvaro-osvaldo-tm:
- Perhaps the function 'handle_signal'  should put outside the 'serve.py' file to avoid 'SOLID'  principles violation
- The function 'shutdown' code perhaps should be put into the 'Server' class , method 'shutdown'.

mkdocs/livereload/__init__.py
```diff
@@ -261,6 +265,10 @@ def serve_request(self, environ, start_response) -> Iterable[bytes]:
         start_response(msg, [("Content-Type", "text/html")])
         return [error_content]
 
+    @property
+    def is_active(self) -> bool:
+        return self._is_active
```
pawamoy, line 270: You used a property to make it read-only, right?

mkdocs/commands/serve.py
```diff
@@ -78,8 +80,47 @@ def error_handler(code) -> bytes | None:
...
+        if isdir(site_dir):
+            shutil.rmtree(site_dir)
+
     server.error_handler = error_handler
 
+    signal(SIGTERM, handle_signal)
+    signal(SIGINT, handle_signal)
+
+    if sys.platform == "linux":
+        from signal import SIGHUP, SIGQUIT, SIGUSR1, SIGUSR2
```
pawamoy, line 111: @I don't think SIGQUIT should be handled. If I read [GNU docs](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html) correctly, as well as this [SO answer](https://stackoverflow.com/a/4047975/3451029), programs should rather not remove temporary files on SIGQUIT, to allow users to inspect them.

I'm not sure about SIGUSR1 and SIGUSR2. Their default action is to terminate a process, but that doesn't mean we should expect users to send them to MkDocs processes.

pawamoy (commented):
Thanks for the PR!

mkdocs/livereload/__init__.py
alvaro-osvaldo-tm: Exactly, I don't want the value to be violated externally.

mkdocs/commands/serve.py
alvaro-osvaldo-tm: Exactly,

- Utilising SIGQUIT to clean up the directory was a mistake
- The signals SIGUSR is edge cases, no problem for me if want to remove.
- SIGHUP can occur when a terminal is closed , as when you open a new terminal window with _`xterm -e 'mkdocs serve'`_ and then closes the terminal window. The default behaviour is to end the processes.

pawamoy: Yeah SIGHUP is probably a good idea. Let's remove SIGQUIT and SIGUSR1/SIGUSR1 then 🙂

alvaro-osvaldo-tm:
It's ready, removed SIGQUIT and SIGUSR signals. Let me know if you need anything else.

- Sorry for take so long , It wasn't clear to me if was me that should remove the SIGQUIT and SIGUSR signals. 
- Also,  I accidentally merged no related code in the same pull request, so I had to made a 'force push'

mkdocs/commands/serve.py
```diff
@@ -78,8 +80,45 @@ def error_handler(code) -> bytes | None:
                     return f.read()
         return None
 
+    def handle_signal(signum, frame) -> None:
+        from signal import strsignal
```
pawamoy, line 84: nit: I think there's no reason to import lazily, could you move it to the top of the module?

pawamoy (changes requested):
So IMO it looks good. I'd just like to see some tests: they would open a subprocess of `mkdocs serve` with `subprocess.Popen`, and then send a signal to the subprocess with `os.kill` to assert that the process correctly terminates and cleaned up (or not) temporary directories.

@tomchristie with such tests would you be comfortable merging?

mkdocs/commands/serve.py
alvaro-osvaldo-tm: Sure, moving now

pawamoy (commented):
As mentioned previously, could you add some tests @alvaro-osvaldo-tm?

alvaro-osvaldo-tm:
Hi, the test was made 

- I was careful for Windows cases , but  still there is an issue I need to check in a windows machine, I don't have one at moment so I will do it this week.
- The test was in a new class 'Shutdown_by_signal_tests' because required an infrastructure to locate the 'mkdocs serve'  directory. Because it's not possible to define the temporary directory as parameter in 'mkdocs server' shell command, so I need to check every '/tmp' directory to found.
- Exists many 'sleeps' to prevent race conditions when running testing

I tried to make the test code readable  , if need , I'm available for revision.

pawamoy:
Thank you @alvaro-osvaldo-tm :rocket: Let me know when you've been able to handle the Windows issue you mentioned.

alvaro-osvaldo-tm:
Hi, it take me a while.

For my surprise windows don't handle POSIX events smooth as Linux.

In windows it handle CTRL_C_EVENT using specific 'Windows API' instead to handle SIGTERM using 'Python's events' library.

I made this conclusion testing into an windows environment, reading some '[Python test cases](https://github.com/python/cpython/blob/main/Lib/test/win_console_handler.py#L39)' and more research.

I have a proposal to enable the cleanup feature for both POSIX and Windows , while keeping  the implementation simple and maintainable. 

It's use the factory pattern, the 'code' and the 'test' will request an implementation in the factories for the signal handling and receive one according the 'operation system'. 

Behind the factory there will be two implementations, one for 'POSIX' and other for 'Windows' with the methods to
configure the 'shutdown callback'  and other to dispatch a 'shutdown' signal for testing

Something like in the 'serve.py' 
```
    def shutdown() -> None:
        if not server.is_active:
            return

        log.info("Shutting down...")

        try:
            server.shutdown()
        finally:
            config.plugins.on_shutdown()

        if isdir(site_dir):
            shutil.rmtree(site_dir)

event_handler = Event_Handler_Factory.new()
event_handler.registry_shutdown( lambda: shutdown )

``` 

And something like in 'shutdown_by_signal_tests.py' testing code:

```
event_handler = Event_Dispatch_Factory.new()
event_handler.shutdown( mkdocs_pid )

for signal in event_handler.get_supported_signals():
   event_handler.dispatch( mkdocs_pid , signal )
   ...
   self.assertTrue( was_cleaned )

```

And the factories make the selection for the correct implementation

```
class Event_Handler_Factory:

  @staticmethod
  def new() -> IEvent_Handler:

    if sys.platform == 'win32':
     return Event_Handler_For_Win32()

     return Event_Handler_For_POSIX()


class Event_Dispatch_Factory:

  @staticmethod
  def new() -> IEvent_Dispatch:

    if sys.platform == 'win32':
     return Event_Dispatch_For_Win32()

     return Event_Dispatch_For_POSIX()

```

Did you like I implement it ? 
No problem if the decisions change after the implementation.

I ask because is slightly different from the current mkdocs code patterns.

orias:
Is this still a known issue? I'm running into giant /tmp/ folders on mkdocs 1.6.1.

https://github.com/mkdocs/mkdocs/issues/3905

Image

alvaro-osvaldo-tm:
Hi @pawamoy, @tomchristie.

I propose refactoring the current implementation to decouple operating system (OS) logic from the core command. By applying the Strategy Pattern, we can isolate OS-specific interrupt handling into dedicated classes.

Currently, the Linux implementation is ready. This architecture allows us to merge the feature now with a 'No-Op' strategy for Windows and macOS, preventing regressions while keeping the codebase open for future contributions.

The modularity of this design makes implementing additional event handlers straightforward. Once the Linux version is merged, I can provide the Windows implementation as a follow-up PR, and possibly MacOS support later.

**Key benefits:**

- Separation of Concerns: mkdocs/commands/serve.py remains nearly unchanged, merely invoking a generic listener interface.
- Traceability: OS-specific bugs will be isolated within their respective strategy classes.
- Cognitive Load: Maintainers only need to review the interface contract rather than complex conditional logic per OS.

**Below is a diagram for context:**

```mermaid
flowchart 

subgraph InterruptStrategy["Interrupt Strategy"]
    Observer
    Handler
end


subgraph Serve["Serve.py"]
 defServe["def Serve()"]
 defShutdown["def shutdown()"]
end

subgraph Infrastructure
    Linux
    Windows
    MacOS
end

defServe -- invoke --> Observer

Observer -- notify --> Handler

Infrastructure -.->|realizes| Observer

Handler  -- callback --> defShutdown


```

PR #3929, opened 2025-02-14
Handle gracefully if raw html parser fails

evan-flynn-apexai:


pawamoy:
What issue are you trying to fix?

evan-flynn-apexai:
@pawamoy apologies for not making an issue first but I'll give an overview of what we observed:

1. After bumping to `mkdocs==1.6.1` our API documentation was failing to build with an error `'unknown status keyword <our-api> in marked section`
    - turned out our doxygen-to-markdown pipeline was spitting out markdown that looked like this:
      ```
     template <typename U  =T\> <br>std::enable_if_t<![namespace::is_some_class](/api-reference/classes/my_cool_class.md)< U >::value, void >
      ```
    - seems that the parser cant handle all the C++ syntax-that-looks-like-html-and-markdown :sweat_smile: 
    - so instead of erroring out, since technically I think this is still valid markdown...this parser should just continue on

The additional `error()` method I added you can remove, I think its being [deprecated anyways](https://github.com/python/cpython/commit/e34bbfd61f405eef89e8aa50672b0b25022de320)

pawamoy:
I see, thank you :slightly_smiling_face:

When you say "was failing to build with an error", what do you mean by error? A logging error message, or a traceback from the raised exception? What exception is this by the way?

evan-flynn-apexai:
Well, that'll depend on the version of cypthon you have.

For the older version we have, its a `NameError` exception on [this line](https://github.com/python/cpython/blob/main/Lib/_markupbase.py#L157) because the `if` statement just above doesn't set the `match` variable in all cases. This has been fixed now in [this commit](https://github.com/python/cpython/commit/e34bbfd61f405eef89e8aa50672b0b25022de320#diff-7304624ffcc8be1268e5f86ebf1723034477694ce33456dc330f9c03f2d1e84f) that I linked to above, that switches it to properly thrown an exception when it reaches the last `else` of that if statement on line 154 in `Lib/_markupbase.py`.

In either case - the above markdown results in an exception. I know wrapping it like this in a `try/except` is gross and not great code, but at least the `mkdocs build` succeeds and we can continue on maybe digging deeper into a proper fix in the `HtmlParser` class.

pawamoy:
Thanks, so NameError and AssertionError.

I'll note that the snippet above converts to the following HTML:

```html
<p>template &lt;typename U  =T&gt; <br>std::enable_if_t&lt;<img alt="namespace::is_some_class" src="/api-reference/classes/my_cool_class.md" />&lt; U &gt;::value, void &gt;</p>
```

Notice the `img` tag instead of `a` tag, because of the exclamation mark. I don't think that's what you want :thinking: 

It's then displayed like this by browsers:

```
template <typename U =T>
std::enable_if_t<namespace::is_some_class< U >::value, void >
```

IMO your Doxygen-to-Markdown pipeline should output `<code>` tags with `<a>` within them for cross-references when needed, or it should escape more aggressively (`<` to `&lt;`, etc.). Currently it looks like the generated HTML *is* invalid. 

I'd be tempted to say that I'm +0 on this, that it just needs a bare try/except and a "why" comment, but seeing how the try/except wraps the whole feed block, I believe it will result in the post-processor collecting only a subset of anchors. Then users with invalid HTML (that will now error silently) will get confused that they either get false-positives or that they don't get true-positives (anchor validation), and this will be very cumbersome to investigate and debug :thinking: I'd prefer seeing a solution that continues parsing the rest of the HTML without bubbling up exceptions.

You could try patching the `_HTMLHandler`'s `parse_marked_section` method to always return -1:

```python
from mkdocs.structure.pages import _HTMLHandler
parser = _HTMLHandler()
parser.parse_marked_section = lambda *args, **kwargs: -1
parser.feed('template <typename U  =T\> <br>std::enable_if_t<![namespace::is_some_class](/api-reference/classes/my_cool_class.md)< U >::value, void >')
```

evan-flynn-apexai:
@pawamoy wow, thank you so much for spending the time here and digging in - I really appreciate it! I'll take another look at the `doxygen-to-markdown` pipeline we've got and see if I can implement your suggestion there instead.

closing this then as probably not worth while to merge into upstream here :+1: thank you though for your help! :pray:

pawamoy:
Thank you for your understanding :bow:

evan-flynn-apexai:
@pawamoy fyi I found a much better solution than this one here thanks to your tips - so again, thank you so much for your help :pray:

pawamoy:
Awesome! Curious to see it, if you may share it :smile:

evan-flynn-apexai:
@pawamoy we use a slightly modified version of `doxybook2` to generate markdown from doxygen XML files. `doxybook2` has a [config flag](https://github.com/matusnovak/doxybook2/blob/5fa2d091efd1e1e9a6bc84317772ab22a7a5b9b1/README.md?plain=1#L318) to enable the fix you exactly gave above about wrapping the link in `<code>` blocks to avoid this error :sweat_smile:

PR #3945, opened 2025-03-21
Fix: Decrase anchor validation treeprocessor prioriy

### LINKED ISSUE #3690: Anchor validation false positives after upgrading to 1.6
First of all, thanks for the very useful [new anchor validation feature](https://www.mkdocs.org/about/release-notes/#anchors)! Upgrading to MkDocs 1.6, I was able to [fix many stale anchors in our documentation](https://github.com/squidfunk/mkdocs-material/issues/7076#issuecomment-2076300125). However, in the process, I came across a few false positives – the anchors are correct, but MkDocs reports them nonetheless. This is particularly the case when using the [Tabbed](https://facelessuser.github.io/pymdown-extensions/extensions/tabbed/) extension with [`combine_header_slug`](https://facelessuser.github.io/pymdown-extensions/extensions/tabbed/#tab-ids), which allows to prepend the slug of the nearest header to the tab.

I believe this needs to be fixed in MkDocs, since it might be some kind of ordering issue when running anchor validation, but I'm tagging @facelessuser here for visibility. By using the [Tabbed](https://facelessuser.github.io/pymdown-extensions/extensions/tabbed/) extension, anchor links to tabs are much more readable, something that many users asked for on the Material for MkDocs issue tracker.

## Example

The following example is used for illustration. You can also download it here: [anchor-validation-bug.zip](https://github.com/mkdocs/mkdocs/files/15144456/anchor-validation-bug.zip)

`index.md`

``` markdown
# Page with tabs

## Headline

=== "Tab A"

    Anchor for this tab is `#headline-tab-a`

=== "Tab B"

    Anchor for this tab is `#headline-tab-b`
```

`foo.md`

``` markdown
# Links to tabs

- [Tab A](index.md#headline-tab-a)
- [Tab B](index.md#headline-tab-b)
```

MkDocs will not detect those links and report them as non-existent:

```
INFO    -  Doc file 'foo.md' contains a link 'index.md#headline-tab-a', but the doc 'index.md' does not contain an anchor '#headline-tab-a'.
INFO    -  Doc file 'foo.md' contains a link 'index.md#headline-tab-b', but the doc 'index.md' does not contain an anchor '#headline-tab-b'.
INFO    -  Documentation built in 0.12 seconds
```

When running the example, it's easy to see that anchor are correct and work as expected.

joapuiib:
Resolves #3690 

> (...) the suggested solution is to decrease the treeprocessor processor priority to 3.9:

https://github.com/mkdocs/mkdocs/blob/fb1d106747930b094bc464189d4b1c349fc541c6/mkdocs/structure/pages.py#L343

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -340,9 +340,9 @@ class _ExtractAnchorsTreeprocessor(markdown.treeprocessors.Treeprocessor):
                     add(anchor)
 
     def _register(self, md: markdown.Markdown) -> None:
-        md.treeprocessors.register(
-            self, "mkdocs_extract_anchors", priority=3.9
-        )  # decreased from 5 (see: #3690)
+        # Priority 3.9 to run *after* PyMDown Extensions' `tab` (4), which runs after `toc` (5),
+        # see https://github.com/mkdocs/mkdocs/issues/3690.
+        md.treeprocessors.register(self, "mkdocs_extract_anchors", priority=3.9)
 
 
 class _RelativePathTreeprocessor(markdown.treeprocessors.Treeprocessor):


mkdocs/structure/pages.py
```diff
@@ -340,7 +340,9 @@ def run(self, root: etree.Element) -> None:
                     add(anchor)
 
     def _register(self, md: markdown.Markdown) -> None:
-        md.treeprocessors.register(self, "mkdocs_extract_anchors", priority=5)  # Same as 'toc'.
+        md.treeprocessors.register(
+            self, "mkdocs_extract_anchors", priority=3.9
+        )  # decreased from 5 (see: #3690)
```
pawamoy, line 345:
```suggestion
        # Priority 3.9 to run *after* PyMDown Extensions' `tab` (4), which runs after `toc` (5),
        # see https://github.com/mkdocs/mkdocs/issues/3690.
        md.treeprocessors.register(self, "mkdocs_extract_anchors", priority=3.9)
```

pawamoy (commented):
Thanks! More than just tests passing, we should also look out for regressions by building a few projects with this change, asserting there's no issue.

joapuiib:
I've tested it with one of my sites that was giving the anchor validation false positive, and it seems that it works with no issues.

However, it might be better if you test it as well, just in case 😅

byrnereese:
Nudge. I would love to see this fix get merged. :)

PR #3956, opened 2025-04-15
Break only when h1 tag is finding when extracting a page title

mschoettle:
#3191 introduced an improvement to finding the first `h1` tag in a page. There is a small bug in that it breaks after any first element instead of only when the `h1` tag is found.

This causes issues with pages that contain a comment, or frontmatter.


oprypin:
I can assure you that this `break` was not accidentally mis-indented. Accepting a heading only at the very beginning folows the same idea as the [original behavior since the year 2017](https://github.com/mschoettle/mkdocs/blob/b9af8b1fbaa63a7aafbd0ddc1987424f7df021d5/mkdocs/utils/__init__.py#L432). I had to comment because this is written as if prior to https://github.com/mkdocs/mkdocs/pull/3191 this title in your case would have been found, but that's not the case.

Allowing a comment, if there's some problem with it, should probably be a much more targeted change.

mkdocs/tests/structure/page_tests.py
```diff
@@ -342,6 +342,35 @@ def test_page_title_from_setext_markdown(self):
...
+        self._test_extract_title(self._COMMENT_CONTENT, expected='Hello')
+
+    _FRONTMATTER_CONTENT = dedent(
+        '''
+        <!--
+        Comment
+        -->
+        ---
+        title: test
+        ---
```
oprypin: It should also be double-checked whether this is actually treated as a frontmatter or just a bunch of visible Markdown. Because I'd guess the latter.

mschoettle:
> I can assure you that this `break` was not accidentally mis-indented. Accepting a heading only at the very beginning folows the same idea as the [original behavior since the year 2017](https://github.com/mschoettle/mkdocs/blob/b9af8b1fbaa63a7aafbd0ddc1987424f7df021d5/mkdocs/utils/__init__.py#L432). I had to comment because this is written as if prior to #3191 this title in your case would have been found, but that's not the case.
> 
> Allowing a comment, if there's some problem with it, should probably be a much more targeted change.

It was not my intention to suggest that #3191 broke previous behaviour. Sorry if that came across. I was trying to figure out where the title is determined and found the most recent change, so I did not look at original behaviour at all. I updated the description.

The way I understood "report the first H1 tag" in the PR description was that it would look until it found it. I agree that the heading should come first. I'll try to figure something out that still expects the heading to come first but allows for comments.

mkdocs/tests/structure/page_tests.py
mschoettle: You are right, that doesn't work. I'll remove that test case.

mschoettle:
I reverted my change and added a more targeted change to skip the first element if it is an HTML placeholder.

oprypin:
Nice :)

oprypin (approved):

PR #3980, opened 2025-05-18
adding a Switch theme button 

DrTeresia:
adding a switch theme button for Software Engineering lab practice, is it possible to change theme dynamically in mkdocs?


pawamoy:
It's not possible, no 🙁 The HTML files are built once, with a given theme's templates. There's no mechanism to build for several themes and switch between them.

tomchristie:
Tho you could create a theme that is actually several themes in a trenchcoat.

PR #3982, opened 2025-05-22
solving an example from a course

Aniket28ks:


pawamoy:
@Aniket28ks please inform the authors of your course that their trainees often mistakenly open PRs against the wrong repository, the actual MkDocs project, and not a test repo, and that they should update their course to prevent that 🙂

Issue #4001, opened 2025-06-24
AssertionError: we should not get here! from html.parser with Python 3.13.5


Lucas-C:
Hi!

Here is a minimal reproduction case:

* `mkdocs.yml`:
```yaml
site_name: fpdf2

docs_dir: docs/
site_dir: public/

nav:
- 'Project Home': 'index.md'
```

* `docs/index.md`:
```markdown
The PDF object as an `obj<</>>endobj` text block.
```

When calling: `python -m mkdocs serve --open`

I get this error:
```
$ python -m mkdocs serve --open
INFO    -  Building documentation...
INFO    -  Cleaning site directory
ERROR   -  Error reading page 'index.md': we should not get here!
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/user/.pyenv/versions/3.13.5/lib/python3.13/site-packages/mkdocs/__main__.py", line 370, in <module>
    cli()
    ~~~^^
  File "/home/user/.pyenv/versions/3.13.5/lib/python3.13/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/user/.pyenv/versions/3.13.5/lib/python3.13/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
  File "/home/user/.pyenv/versions/3.13.5/lib/python3.13/site-packages/click/core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/user/.pyenv/versions/3.13.5/lib/python3.13/site-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.pyenv/versions/3.13.5/lib/python3.13/site-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
  File "/home/user/.pyenv/versions/3.13.5/lib/python3.13/site-packages/mkdocs/__main__.py", line 272, in serve_command
    serve.serve(**kwargs)
    ~~~~~~~~~~~^^^^^^^^^^
[...]

PR #4007, opened 2025-07-02
Fix issue with handling `</>`

facelessuser:
Markdown recently made a fix for an issue handling `</>` which would get stripped out of content. Reference:
https://github.com/Python-Markdown/markdown/commit/64a3c0fbc00327fbfee1fd6b44da0e5453287fe4

MkDocs uses Markdown's monkey-patched HTMLParser, but does not use the HTML extractor that has the logic to handle the things that have been monkey-patched in. Because of this, MkDocs now fails when parsing `</>`. This fixes the issue by adding the same logic to handle the issue that Markdown uses.

### DIFF
--- a/mkdocs/structure/pages.py
+++ b/mkdocs/structure/pages.py
@@ -542,6 +542,13 @@ class _HTMLHandler(markdown.htmlparser.htmlparser.HTMLParser):  # type: ignore[n
         super().__init__()
         self.present_anchor_ids: set[str] = set()
 
+    def parse_starttag(self, i: int) -> int:
+        # Treat `</>` as normal data as it is not a real tag.
+        if self.rawdata[i:i + 3] == '</>':
+            self.handle_data(self.rawdata[i:i + 3])
+            return i + 3
+        return super().parse_starttag(i)
+
     def handle_starttag(self, tag: str, attrs: Sequence[tuple[str, str]]) -> None:
         for k, v in attrs:
             if k == 'id' or (k == 'name' and tag == 'a'):


facelessuser:
This fixes #4001

facelessuser:
@tomchristie is this project still being maintained? I'd love to see this bug fix get merged and released.

facelessuser:
It seems I am unable to push this forward. I am willing to provide changes and updates if the Mkdocs team is willing to engage. If not, I guess we are not getting this fix.

Guts:
I guess we should ping @lovelydinosaur

thomasborgen:
We also need this fix 🙏🏼

SOV710:
This bug is breaking our documentation. MkDocs team, would it be possible to merge this PR and release 6.1.2? Much appreciated!

facelessuser:
I had to fix some other HTML issues in Python Markdown. Because it broke MkDocs for Python Markdown, I had to resort to hackery. The good news is that the next Python Markdown, will fix this issue for MkDocs. PR is here: https://github.com/Python-Markdown/markdown/pull/1593.

mschoettle:
@facelessuser Thanks a lot! I just updated markdown to `3.10.2` and `</>` now renders as is 👍

facelessuser:
Yep, it was one of those things; it hadn't occurred to me that it could be fixed in Python Markdown. I'd argue it shouldn't have had to be fixed there, but necessity is the mother of invention. When Python Markdown, which uses MkDocs as its documentation engine, was now breaking due to this, a solution became necessary.

facelessuser:
This isn't needed anymore, as a fix that works here is available in the latest Python Markdown directly.

Issue #4014, opened 2025-08-02
MkDocs `use_directory_urls: true` bug with `click==8.2.2`


glenn-jocher:
### Context

Latest version of `click` dependency (released August 2, 2025) is breaking `use_directory_urls: true` functionality in MkDocs and MkDocs Material.

### Bug description

This is causing 404's when clicking on links to previously published pages.

Example Good with click==8.2.1
https://docs-paasvkc6c-ultralytics.vercel.app/models/yolo11/

Example Bad with click==8.2.2
https://docs-paasvkc6c-ultralytics.vercel.app/models/yolo11/index.html

### Related links

https://github.com/ultralytics/ultralytics/issues/21568

### Reproduction

Reproduce with steps above.

### Steps to reproduce

Set `use_directory_urls: true` in mkdocs.yml with `click==8.2.2` installed in environment, then run `mkdocs serve`.

### Browser

Chrome

squidfunk:
This is a quite fatal bug, as all URLs are different when 8.2.2 is silently pulled in in CI. It affects all MkDocs projects, and is not dependent on the theme. To illustrate, here's a minimal reproduction that showcases the problem: [click-bug.zip](https://github.com/user-attachments/files/21558889/click-bug.zip)

Running this project, here's the `sitemap.xml` with `click==8.2.1`:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
    <url>
         <loc>http://example.com/</loc>
         <lastmod>2025-08-02</lastmod>
    </url>
    <url>
         <loc>http://example.com/foo/</loc>
         <lastmod>2025-08-02</lastmod>
    </url>
</urlset>
```

... and here with `click==8.2.2`:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
    <url>
         <loc>http://example.com/index.html</loc>
         <lastmod>2025-08-02</lastmod>
    </url>
    <url>
         <loc>http://example.com/foo.html</loc>
         <lastmod>2025-08-02</lastmod>
    </url>
</urlset>
```

[...]

PR #4015, opened 2025-08-02
fix: 🐞 MkDocs compatibility with Click 8.2.2+ boolean flag handling

onuralpszr:
# Fix Attempt MkDocs compatibility with Click 8.2.2+ boolean flag handling

## Problem

MkDocs stopped respecting the `use_directory_urls: true` configuration setting when used with Click 8.2.2+, causing sitemap.xml to generate `.html` URLs instead of directory-style URLs.

**Root Cause**: Click 8.2.2 introduced behavioral changes in [PR #2956](https://github.com/pallets/click/pull/2956) where boolean flags with `default=None` now return `False` instead of `None` when not specified by the user. This broke MkDocs' config loading logic which relies on `None` values to distinguish between "parameter not specified" vs "explicitly set to False".

## Affected Behavior (https://github.com/mkdocs/mkdocs/issues/4014) 

- ❌ **Before**: `use_directory_urls: true` in mkdocs.yml → sitemap contains `/foo/` (correct)
- ❌ **With Click 8.2.2**: `use_directory_urls: true` in mkdocs.yml → sitemap contains `/foo.html` (wrong)
- ✅ **After this fix**: `use_directory_urls: true` in mkdocs.yml → sitemap contains `/foo/` (correct)

## Solution

This fix uses Click's `ctx.get_parameter_source()` API to detect when CLI parameters are user-specified vs defaults. When `ParameterSource.DEFAULT` is detected, we return `None` to preserve MkDocs' original config precedence logic.

### Parameters Fixed

1. **`--use-directory-urls/--no-directory-urls`** - Primary fix for sitemap URL generation
2. **`--strict/--no-strict`** - Ensures config-driven strict mode works correctly  
3. **Build type flags** (`--dirty`, `--clean`, `--dirtyreload`) - Fixes serve command parameter handling

### Implementation

Each affected parameter now uses a custom callback that:
```python
def callback(ctx, param, value):
    if ctx.get_parameter_source(param.name) == click.core.ParameterSource.DEFAULT:
        return None  # Preserve original "not specified" behavior
    return value     # User explicitly set the flag
```
 
 This will also give times to mkdocs-material and mkdocs itself to check the behavioral change maybe thoroughly so it can also helps MkDocs users to easy transaction while they are using this fix and using latest click version hopefully 
 
 Thank you
 
 cc @glenn-jocher @squidfunk
 
 Fix: https://github.com/mkdocs/mkdocs/issues/4014

### DIFF
--- a/mkdocs/__main__.py
+++ b/mkdocs/__main__.py
@@ -215,21 +215,87 @@ def color_option(f):
     )(f)
 
 
+def use_directory_urls_option(f):
+    def callback(ctx, param, value):
+        # Only set the value if the parameter was actually specified by the user.
+        # This works around a Click 8.2.2 behavioral change where boolean flags with default=None
+        # return False instead of None when not specified.
+        if hasattr(ctx, 'get_parameter_source'):
+            if ctx.get_parameter_source(param.name) == click.core.ParameterSource.DEFAULT:
+                return None
+        return value
+
+    return click.option(
+        '--use-directory-urls/--no-directory-urls',
+        is_flag=True,
+        default=None,
+        help=use_directory_urls_help,
+        callback=callback,
+    )(f)
+
+
+def strict_option(f):
+    def callback(ctx, param, value):
+        # Only set the value if the parameter was actually specified by the user.
+        # This works around a Click 8.2.2 behavioral change where boolean flags with default=None
+        # return False instead of None when not specified.
+        if hasattr(ctx, 'get_parameter_source'):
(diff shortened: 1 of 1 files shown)


tomchristie:
I think we can safely close this. Given that `flask` was also impacted by the same yanked `click` regression we're not likely to be impacted by this a second time. Thanks for the neat catch @onuralpszr.

And the closed PR will still be here if we actually need it.

PR #4058, opened 2025-11-06
Restrict click dependency version to `<=8.2.1`

### LINKED ISSUE #4032: `mkdocs` does not watch for file changes when using `click>8.2.1`
I noticed when I did a dependency upgrade to some of the latest and greatest, that `mkdocs` was no longer watching for file changes. Eventually I narrows this down. With `click==8.2.1` everything works and file changes trigger everything to reload. However, when this gets bumped to `click=8.3.0` no files are being watched and no dynamic doc reloading when doing a `mkdocs serve`. 

It is likely related to this issue: https://github.com/mkdocs/mkdocs/issues/4014

Can we pin to 8.2.1 as a hot-fix until this is resolved?

johnthagen:
Closes #4032

As requested by a maintainer:

- https://github.com/mkdocs/mkdocs/issues/4032#issuecomment-3490948321

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 dynamic = ["version"]
 requires-python = ">=3.8"
 dependencies = [
-    "click >=7.0, <8.3.0",
+    "click >=7.0, <=8.2.1",
     "Jinja2 >=2.11.1",
     "markupsafe >=2.0.1",
     "Markdown >=3.3.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 dynamic = ["version"]
 requires-python = ">=3.8"
 dependencies = [
-    "click >=7.0",
+    "click >=7.0, <8.3.0",
     "Jinja2 >=2.11.1",
     "markupsafe >=2.0.1",
     "Markdown >=3.3.6",


lovelydinosaur:
Seems like the lightest footprint change to resolve this in the short term.
Presumably we merge this and then roll a release?

johnthagen:
> Presumably we merge this and then roll a release?

@lovelydinosaur Yes, I think that would be best.

johnthagen:
@lovelydinosaur Actually, it looks like Click might have fixed this upstream

- https://github.com/pallets/click/pull/3152
- https://github.com/pallets/click/issues/3084

So perhaps waiting for Click `8.3.2` to release could be better? @davidism might know when Click plans to release this merged fix.

davidism:
Soon, but as I and most volunteer maintainers point out any time this is asked, we don't promise any particular timeline.

davidism:
Rather than using a max pin, you can also exclude the specific versions that don't work so you still get the next release.

dotlambda:
> Actually, it looks like Click might have fixed this upstream

The mkdocs tests still fail with that patch applied to click.
c.f. https://github.com/NixOS/nixpkgs/pull/454434#issuecomment-3564024069

johnthagen:
According to a `click` maintainer, there will be no more bug fix releases for `click` before v9

- https://github.com/pallets/click/issues/3107#issuecomment-3830687046

johnthagen:
Latest upstream Click issue

- https://github.com/pallets/click/issues/3403

johnthagen:
This issue was fixed upstream in Click 8.4.0.

- https://github.com/pallets/click/pull/3404

PR #4108, opened 2026-05-06
Fix indented path spec config handling

YangLeeFangfei:
## Summary

Normalize indented multiline `PathSpec` config values before parsing them with pathspec.

This avoids treating common indentation in multiline strings as part of the gitignore-style patterns.

## Verification

- hatch run test.py3.9-default:test


YangLeeFangfei:
Note: the current CI failures appear unrelated to this change. The same 8 test matrix jobs are failing on the latest scheduled CI run for upstream `master`: https://github.com/mkdocs/mkdocs/actions/runs/25246011788

One example failure is `tests.cli_tests.CLITests.test_serve_watch_theme`, where `livereload` is passed as `False` instead of the expected `True`. This PR only changes `PathSpec` config parsing.

PR #4138, opened 2026-06-06
Fix draft docs user guide test with pathspec 1.1

marko1olo:
The `draft_docs` user-guide regression test copies the YAML block scalar example into a direct Python string. YAML strips the block indentation before MkDocs sees the value, but the direct test string kept that indentation.

With current `pathspec`, those leading spaces stop the comment and glob lines from matching, so the test no longer exercises the documented `draft_docs` behavior and CI fails with the draft pages included in `mkdocs build`.

This dedents the direct test string so it matches the actual YAML block scalar value from the user guide.

Tests run:
- `python -m unittest mkdocs.tests.build_tests.BuildTests.test_draft_docs_with_comments_from_user_guide`
- `python -m unittest mkdocs.tests.build_tests.BuildTests.test_draft_pages_with_invalid_links`
- `python -m unittest mkdocs.tests.build_tests.BuildTests`
- `ruff check mkdocs\tests\build_tests.py`
- `python -X utf8 -m isort --check-only --diff mkdocs\tests\build_tests.py`
- `python -X utf8 -m black --workers 1 -q --check --diff mkdocs\tests\build_tests.py`
- `git diff --check`


marko1olo:
CI note after rechecking the two failing PyPy jobs:

- `test (pypy-3.9-v7.x, ubuntu-latest)` fails during dependency installation before MkDocs tests run. `cryptography` is built from source and its `pyo3-ffi` build fails with `the configured PyPy interpreter version (3.9) is lower than PyO3's minimum supported version (3.11)`.
- `test (pypy-3.9-v7.x, windows-latest)` also fails during dependency installation before tests run. `zstandard` fails while getting wheel build requirements because PyPy's CFFI/distutils path raises `ModuleNotFoundError: No module named 'distutils.msvc9compiler'`.

I reran the relevant local suite with temp files redirected away from the locked user temp directory:

```text
TEMP=.tmp-tests TMP=.tmp-tests python -m unittest mkdocs.tests.build_tests.BuildTests
Ran 51 tests in 11.990s
OK
```

So the remaining red jobs look like PyPy/dependency-build infrastructure failures, not failures in this patch.

PR #4146, opened 2026-06-07
Use PyPy 3.11 in CI

marko1olo:
The current PyPy 3.9 CI jobs can fail before MkDocs tests start while installing Hatch.

Two separate bootstrap issues show up with the old PyPy target:

- latest `cryptography` no longer publishes PyPy 3.9 wheels, so pip falls back to a source build that now requires PyPy 3.11 through PyO3
- on Windows, Hatch's `zstandard` dependency has no PyPy wheel for the old interpreter and the source build fails while looking for `distutils.msvc9compiler`

MkDocs already tests CPython 3.8 and 3.9 directly. Updating the PyPy leg to PyPy 3.11 keeps PyPy implementation coverage while using currently supported wheels for the CI bootstrap dependencies.

Local checks:
- `git diff --check`
- `python -m pip install --dry-run --upgrade hatch`
- verified `cryptography` and `backports-zstd` resolve PyPy 3.11 wheels for `manylinux_2_28_x86_64`, `win_amd64`, and `macosx_11_0_arm64`


marko1olo:
Follow-up from the first run on the updated branch: the PyPy 3.11 bootstrap now gets past dependency installation successfully.

The remaining test failures I checked are the existing `BuildTests.test_draft_docs_with_comments_from_user_guide` baseline failure, not the Hatch/cryptography/zstandard install failure. For example, the PyPy 3.11 Ubuntu job installed `hatch`, `cryptography-48.0.0`, and `backports-zstd-1.5.0`, then failed later on `other_unpublished.html` in the draft docs test.

Issue #3650, opened 2024-04-17
Docs enhancement request: document mkdocs-serve errors


ljsinclair:
I haven't found any docs on mkdocs-serve errors I've come across and have solutions below for anyone who's interested:

| Error | Cause | Example |
|---|---|---|
| PermissionError: [Errno 13] Permission denied | Leading `/` on URL | `/path/to/filename.md` |
| IsADirectoryError: [Errno 21] | No trailing `.md` on filename in URL | `path/to/filename/` |
| OSError [Errno 98] | Port 8080 assigned twice | * Run `lsof -i :8000` to obtain the PID number `nnnn` for the port<br/>* Use the PID number in command: `sudo kill 9 nnnn` |

glaurne:
I’d like to work on this issue if it’s still available.

PR #4151, opened 2026-06-25
docs: clarify nav page paths

Uomocapra:
﻿## Summary

Clarifies a point that is easy to misread in `nav`: documentation pages should be written as Markdown paths relative to `docs_dir`, while values starting with `/` are URL paths from the server root.

Refs #3650.

## Testing

I checked the change with `git diff --check`, `npx --yes markdownlint-cli docs/user-guide/configuration.md --ignore docs/CNAME`, and `python -m mkdocs build --strict`.


Uomocapra:
CI note: the failing test jobs here are not caused by this docs change. I reproduced the `draft_docs` failure on upstream `master` with `pathspec 1.1.1` and opened #4152 for that fixture issue.

PR #4152, opened 2026-06-25
tests: normalize draft_docs fixture indentation

Uomocapra:
﻿## Summary

The `draft_docs` fixture had extra indentation from the Python string, which newer `pathspec` versions treat as part of the pattern. This dedents the fixture so the test matches the YAML block shown in the user guide.

I reproduced the same failure on upstream `master` before applying the change.

## Testing

I checked the change with `git diff --check`, `hatch run +py=3.12 test:python -m unittest mkdocs.tests.build_tests.BuildTests.test_draft_docs_with_comments_from_user_guide`, `hatch run +py=3.12 test:python -m unittest mkdocs.tests.build_tests.BuildTests`, and `hatch run style:check`.


Uomocapra:
CI note: the CPython jobs, lint, package, docs build, and style checks pass for this fixture change. The remaining PyPy failures happen while installing dependencies before the test suite starts, so they appear unrelated to this PR.

PR #4207, opened 2026-08-09
fix: give a clean error when the dev server fails to bind

### LINKED ISSUE #4131: `WinError 10013` - command of `mkdocs serve`
I'm getting this error. Is there a solution?

Manushprajwal7:
## Summary

Fixes #4131.

`mkdocs serve` crashes with a raw, unhandled traceback whenever the dev server's socket cannot be bound. This is most commonly caused by the address already being in use, but it also happens for OS-specific access errors — including `WinError 10013` on Windows, which the linked issue reports and which can occur even for a "free" port if it falls in a range Windows has reserved for its own use (e.g. via Hyper-V/WSL).

The root cause: `server.serve()` → `server_bind()` raises a plain `OSError` in that case, and nothing catches it, so it propagates straight out of the `serve` command and crashes with a full traceback instead of being handled like every other CLI startup failure (see `commands/gh_deploy.py` for the established `log.error()` + `raise Abort()` pattern used elsewhere in this codebase).

## Changes

- `mkdocs/commands/serve.py`: catch `OSError` around `server.serve()` and raise `Abort` with a clear, actionable message pointing at the `--dev-addr` option — matching the existing error-handling convention used by `gh_deploy.py` and `build.py`.
- Added a regression test in `mkdocs/tests/cli_tests.py`.

## Verification

I reproduced the exact bug on Windows by holding a port open with a second process and running `mkdocs serve -a 127.0.0.1:<port>` against it — this gave the *exact* traceback from the issue, ending in `PermissionError: [WinError 10013] An attempt was made to access a socket in a way forbidden by its access permissions`. With this fix applied, the same reproduction now prints (exit code 1, no traceback):

```
ERROR   -  Cannot bind to address 127.0.0.1:<port> - [WinError 10013] An attempt was made to access a socket in a way forbidden by its access permissions. This usually means the address is already in use, or your OS is denying access to it (for example, Windows may reserve some ports for its own use). Try serving on a different address with the '--dev-addr' option.

Aborted!
```

## Test plan

- [x] Manually reproduced the original crash and confirmed the fix resolves it (see above)
- [x] Added `test_serve_bind_error_aborts_cleanly` in `cli_tests.py`
- [x] `python -m unittest mkdocs.tests.cli_tests` — 38 passed
- [x] Full suite: `python -m unittest discover -s mkdocs/tests` — same pass/fail counts as the unmodified baseline (4 pre-existing Windows-symlink errors + 2 pre-existing unrelated failures, neither related to this change)
- [x] `ruff check` on changed files — clean


Manushprajwal7:
Style check now passes after fixing a formatting nit in the new test.

The remaining `test` job failures (`test_draft_docs_with_comments_from_user_guide` on Python 3.10+, plus an unrelated `zstandard` wheel-build failure on pypy-3.9/Windows) are pre-existing and unrelated to this change — the same `test_draft_docs_with_comments_from_user_guide` failure has been occurring on every scheduled CI run against `master` for the past month (e.g. [this run](https://github.com/mkdocs/mkdocs/actions/runs/31243474392) against `2862536`, the commit this PR is based on). My added test (`test_serve_bind_error_aborts_cleanly`) and the `serve.py` change pass cleanly across all platforms/versions.

Manushprajwal7:
Root-caused the persistent \`test\` job failures (\`test_draft_docs_with_comments_from_user_guide\` on Python 3.10+) — it's a real bug independent of this PR, caused by a newer \`pathspec\` release no longer implicitly stripping leading whitespace from gitignore-style pattern lines (mkdocs' own \`PathSpec\` config option never dedented its input). Filed the fix separately, since it's unrelated to what this PR does: #4208.

This PR (bind-error handling) stays focused on its original scope.

Manushprajwal7:
This branch now also includes the fix from #4208 (dedent PathSpec config values), cherry-picked in temporarily — it was causing most of the \`test\` job failures here too (same root cause, confirmed independent of this PR's actual change). Once #4208 merges to master, that commit will naturally disappear from this diff on rebase; until then this keeps CI green so the actual bind-error fix is easy to review.

PR #4208, opened 2026-08-09
fix: dedent PathSpec config values before compiling patterns

Manushprajwal7:
## Summary

While working on #4207 I noticed CI's `test` job failing consistently on `test_draft_docs_with_comments_from_user_guide` (Python 3.10+) — and confirmed it also fails the same way on current `master`, unrelated to that PR. This fixes the actual root cause.

**Root cause:** `draft_docs`, `exclude_docs`, and `not_in_nav` are all backed by `config_options.PathSpec`, which compiles the value via `pathspec.gitignore.GitIgnoreSpec.from_lines(lines=value.splitlines())`. `pathspec`'s newer major release (1.x) correctly stops implicitly stripping leading whitespace from each pattern line — per the real gitignore format, leading whitespace is significant unless escaped. mkdocs never dedented the value first, so a value written with common leading indentation on every line (natural for a multi-line string in `mkdocs.yml`, and exactly what the affected test does) silently matches nothing once pathspec stopped compensating for it. mkdocs only declares `pathspec >=0.11.1` with no upper bound, so this breaks silently as soon as a newer pathspec gets resolved.

Verified directly against `pathspec.gitignore.GitIgnoreSpec`:
- `pathspec==0.11.1` (mkdocs' declared minimum): matches correctly with or without this fix, since it stripped leading whitespace itself.
- `pathspec==1.1.1` (current release): matching silently breaks without this fix, works correctly with it.

## Changes

- `mkdocs/config/config_options.py`: dedent the value in `PathSpec.run_validation()` before splitting it into lines, so behavior is consistent regardless of which pathspec version is installed — matching how a YAML `|` block scalar's common indentation is normally handled.
- Added `PathSpecTest` in `mkdocs/tests/config/config_options_tests.py` covering both the indented and non-indented cases directly against the config option.

## Test plan

- [x] `test_draft_docs_with_comments_from_user_guide` (the originally-failing test) now passes under `pathspec==1.1.1`
- [x] Full suite passes identically under both `pathspec==0.11.1` and `pathspec==1.1.1` (726-727 passed; only the 4 pre-existing Windows-symlink-privilege errors remain, unrelated and present on unmodified `master` too)
- [x] New `PathSpecTest` covers both the indented (regression) and plain (no change in behavior) cases
- [x] `black`/`isort`/`ruff check` on changed files — clean


Manushprajwal7:
The 2 remaining \`pypy-3.9\` failures (Ubuntu and Windows) are unrelated to this change — they fail at the dependency-install step, before any test code runs: \`cryptography\`'s Rust extension no longer builds on PyPy 3.9 (PyO3 dropped support for it), and \`zstandard\` fails to build on Windows because \`distutils\` was removed. Confirmed the same two jobs fail identically on current \`master\` ([run 31243474392](https://github.com/mkdocs/mkdocs/actions/runs/31243474392)), so this is a pre-existing CI/toolchain issue independent of this PR.

PR #4218, opened 2026-09-16
fix(search): remove lunr stop-word filter so programming keywords are indexed

### LINKED ISSUE #4167: Searching for words that match Python keywords won't return anything
This was first noticed here: https://github.com/astral-sh/ruff/issues/21869

It seems keywords are excluded. Searching for `while`, `if`, `for`, `from`, etc., using the search bar will only result in words or blobs that contains the text, but not the text itself.

Any of those keywords give no results on MKDocs own website as well. https://www.mkdocs.org/ 

Image

Image

Image

Image

Image

okxint:
Fixes #4167.

## Problem

`lunr.js` includes a built-in `stopWordFilter` that drops common English
words — "for", "if", "in", "is", "from", "while", "not", "will", etc. — from
both the **index pipeline** (at build time) and the **search pipeline** (at
query time).

Those words overlap heavily with Python keywords (and keywords in many other
languages), so searching for them on any MkDocs site returns no results.  This
was first noticed in the Ruff documentation (astral-sh/ruff#21869) but affects
every MkDocs site built with the default search plugin.

## Fix

Inside the `lunr()` builder callback in both JS files, explicitly remove the
stop-word filter from both pipelines **before** adding fields:

```js
if (lunr.stopWordFilter) {
  this.pipeline.remove(lunr.stopWordFilter);
  this.searchPipeline.remove(lunr.stopWordFilter);
}
```

Removing it from `searchPipeline` is equally important: even if the index is
built without stop-word filtering, a query term that is itself a stop word
would be stripped before the index is consulted, still producing zero hits.

The `if (lunr.stopWordFilter)` guard makes the change safe when non-English
language plugins replace the pipeline (some of them unregister `stopWordFilter`
themselves).

## Files changed

| File | Change |
|------|--------|
| `mkdocs/contrib/search/templates/search/worker.js` | Remove stop-word filter in the runtime Lunr builder |
| `mkdocs/contrib/search/prebuild-index.js` | Remove stop-word filter in the Node pre-build script |


okxint:
Friendly ping. Happy to add a test or adjust the approach if there's feedback.

